PR #130, opened 2024-10-28
Add list stargazers tool

EricGustin:
# PR Description
As a celebration for arcade-ai becoming open sourced, this PR adds a tool to list the stargazers for a particular repository

### Example `arcade chat -h localhost` usage:
image


toolkits/github/arcade_github/tools/activity.py
```diff
@@ -38,3 +38,56 @@ async def set_starred(
...
+    limit: Annotated[
+        Optional[int],
+        "The maximum number of stargazers to return. If not provided, all stargazers will be returned.",
+    ] = None,
+) -> Annotated[dict, "A dictionary containing the stargazers for the specified repository"]:
+    """List the stargazers for a GitHub repository."""
+    url = get_url("repo_stargazers", owner=owner, repo=repo)
+    headers = get_github_json_headers(context.authorization.token)
+
+    per_page = 30
```
nbarbettini, line 59: Can this be bigger? (more efficient)
If so, then I'd set `per_page` to be e.g. `min(limit, 100)`

toolkits/github/arcade_github/tools/activity.py
```diff
@@ -38,3 +38,56 @@ async def set_starred(
...
+                    "login": stargazer.get("login"),
+                    "id": stargazer.get("id"),
+                    "node_id": stargazer.get("node_id"),
+                    "html_url": stargazer.get("html_url"),
+                }
+                for stargazer in data
+            ])
+
+            if len(data) < per_page:
+                break
```
nbarbettini, line 88: If I'm reading this right, this will sometimes result in `len(data) > limit` because it adds a full page of results every time, and _then_ checks that it's over the limit.

toolkits/github/arcade_github/tools/activity.py
EricGustin: Thanks, addressed

toolkits/github/arcade_github/tools/activity.py
EricGustin: yeah, if
```
if len(data) < per_page:
    break
```
is satisfied, then this means that there are less stargazers than the requested limit.

Otherwise, if there are equal or more stargazers than the requested limit, then the `while` condition will no longer be satisfied.

nbarbettini: I gotcha, I missed this crucial bit below
```
stargazers = stargazers[:limit]
```

😅

nbarbettini (approved):
LGTM! 🚀

PR #137, opened 2024-11-01
fix: circular import in arcade.core

nbarbettini:
Fixes a circular import issue where `arcade.sdk -> arcade.core` but also `arcade.core -> arcade.sdk`. My mistake!

Moved some of the shared classes down into `core`, and re-exported them to `sdk` to keep the expected interface for devs.


examples/fastapi/arcade_example_fastapi/main.py
```diff
@@ -19,7 +19,7 @@
 
 class ChatRequest(BaseModel):
     message: str
-    user_id: str
+    user_id: str | None = None
```
nbarbettini, line 22: Meant to be an optional parameter, fixed

examples/fastapi/arcade_example_fastapi/main.py
```diff
@@ -12,14 +12,14 @@
 
 app = FastAPI()
 
-actor_secret = os.environ.get("ARCADE_ACTOR_SECRET")
+actor_secret = os.environ["ARCADE_ACTOR_SECRET"]
```
nbarbettini, line 15: Standardized on the syntax used a few lines above.

Accessing via key will force an error if the key does not exist (rather than fail silently), which is desired in this case because it provides immediate feedback.

examples/fastapi/arcade_example_fastapi/main.py
```diff
@@ -33,14 +33,13 @@ async def postChat(request: ChatRequest, tool_choice: str = "execute"):
...
-                # "GitHub.SearchIssues",
-                # "Slack.SendDmToUser",
-                # "Slack.SendMessageToChannel",
+                "Math.Add",
+                "Math.Subtract",
+                "Math.Multiply",
+                "Math.Divide",
+                "Math.Sqrt",
+                # Other tools can be added as needed:
+                # "Math.SumList"
```
nbarbettini, line 42: This example shows adding the math toolkit, so `tools` should reference math tools. Made this clearer!

EricGustin (approved):
Looks good!

PR #138, opened 2024-11-01
fix: runtime error when OTEL is not enabled

nbarbettini:
If an `otel_meter` is not passed when constructing the actor (which is allowed, it's an optional param), the actor would later crash at runtime.


EricGustin (approved):
Nice catch

PR #142, opened 2024-11-01
fix: Wait for healthy actor in arcade dev

nbarbettini:
Context: Currently, `arcade dev` starts the actor process and then waits a hardcoded amount of time (2sec) for the actor to start up. This isn't enough time on some slower machines, which leads to the engine trying to start but failing.

Fix: Wait until the actor is healthy according to its own `/actor/health` endpoint.

### DIFF
--- a/arcade/arcade/cli/launcher.py
+++ b/arcade/arcade/cli/launcher.py
@@ -277,7 +277,7 @@ def _manage_processes(
             console.print("Starting actor server...", style="bold green")
             actor_process = _start_process("Actor", actor_cmd, debug=debug)
 
-            _wait_for_healthy_actor(actor_host, actor_port)
+            _wait_for_healthy_actor(actor_process, actor_host, actor_port)
 
             # Start the engine
             console.print("Starting engine...", style="bold green")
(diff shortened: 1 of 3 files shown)


arcade/arcade/cli/launcher.py
```diff
@@ -28,8 +30,8 @@
 
 
 def start_servers(
-    host: str,
-    port: int,
+    actor_host: str,
+    actor_port: int,
```
nbarbettini, line 34: Renamed for clarity

arcade/arcade/cli/launcher.py
```diff
@@ -55,13 +57,13 @@ def start_servers(
...
     # Prepare command-line arguments for the actor server and engine
-    actor_cmd = _build_actor_command(host, port, debug)
+    actor_cmd = _build_actor_command(actor_host, actor_port, debug)
 
     # even if the user didn't pass an env file we may have found it in the default locations
     engine_cmd = _build_engine_command(engine_config, engine_env=env_file if env_file else None)
 
     # Start and manage the processes
-    _manage_processes(actor_cmd, engine_cmd, debug=debug)
+    _manage_processes(actor_cmd, actor_host, actor_port, engine_cmd, debug=debug)
```
nbarbettini, line 66: Now passing host and port into here

arcade/arcade/core/config_model.py
```diff
@@ -118,6 +118,14 @@ def load_from_file(cls) -> "Config":
 
         config_data = yaml.safe_load(config_file_path.read_text())
 
+        if config_data is None:
+            raise ValueError(
+                "Invalid credentials.yaml file. Please ensure it is a valid YAML file."
+            )
+
+        if "cloud" not in config_data:
+            raise ValueError("Invalid credentials.yaml file. Expected a 'cloud' key.")
```
nbarbettini, line 127: Unrelated -- added more graceful/informative error messages if the `credentials.yaml` file is borked for some reason

arcade/arcade/cli/launcher.py
Spartee: Should we allow host and port to also be passed for a short time? Get into that practice now?

Spartee (approved):
Approved with one comment. Really looks fine. makes sense.

PR #143, opened 2024-11-01
fix: No output from arcade login when already logged in

nbarbettini:
I ran into this and was scratching my head why `arcade login` was exiting with no output.

Fix: Get config data from the new top-level object we introduced.

### DIFF
--- a/arcade/arcade/cli/authn.py
+++ b/arcade/arcade/cli/authn.py
@@ -133,14 +133,12 @@ def check_existing_login() -> bool:
         try:
             with open(config_file_path) as f:
                 config: dict[str, Any] = yaml.safe_load(f)
-            api_key = config.get("api", {}).get("key")
-            email = config.get("user", {}).get("email")
+            cloud_config = config.get("cloud", {})
+            api_key = cloud_config.get("api", {}).get("key")
+            email = cloud_config.get("user", {}).get("email")
 
             if api_key and email:
-                console.print(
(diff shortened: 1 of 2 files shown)


arcade/arcade/cli/authn.py
```diff
@@ -133,14 +133,12 @@ def check_existing_login() -> bool:
         try:
             with open(config_file_path) as f:
                 config: dict[str, Any] = yaml.safe_load(f)
-            api_key = config.get("api", {}).get("key")
-            email = config.get("user", {}).get("email")
+            cloud_config = config.get("cloud", {})
+            api_key = cloud_config.get("api", {}).get("key")
+            email = cloud_config.get("user", {}).get("email")
```
nbarbettini, line 138: These were always coming back null because they are nested under `cloud` now.

arcade/arcade/cli/authn.py
```diff
@@ -133,14 +133,12 @@ def check_existing_login() -> bool:
...
 
             if api_key and email:
-                console.print(
-                    f"You're already logged in as {email}. "
-                    f"Delete {config_file_path} to log in as a different user."
-                )
+                console.print(f"You're already logged in as {email}. ", style="bold green")
```
nbarbettini, line 141: Kept the same message, but split it up between here and the caller (main.py)

EricGustin (approved):
Thanks! Was somewhere on my list of TODOs but had forgotten about this.

PR #144, opened 2024-11-02
Fix log that stated the number of tools found in toolkit

EricGustin:
The original log was counting the number of files in the toolkit module. This PR fixes this such that it counts the number of tools in the toolkit.


nbarbettini (approved):
Good catch

PR #145, opened 2024-11-05
Return success message if playback is altered

EricGustin:
# PR Description
Previously, if a tool adjusted the playback state, then the tool would return the current playback state after the modification had occurred. The problem with this approach was that Spotify would not update the playback state in time (sometimes), so the tools were returning stale data!


nbarbettini (approved):
LGTM

PR #147, opened 2024-11-05
fix: arcade dev waits forever if no toolkits are installed

nbarbettini:
Fixes an edge case where the actor doesn't start because no toolkits are installed in the local environment, but `arcade dev` keeps waiting for a healthy actor.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -40,6 +40,17 @@
       "cwd": "${workspaceFolder}"
     },
     {
+      "name": "Debug `arcade dev`",
+      "type": "python",
+      "request": "launch",
+      "program": "${workspaceFolder}/arcade/run_cli.py",
+      "args": ["dev", "-c", "~/Team/Engine/config.dev.yaml", "-e", "~/Team/Engine/.env"],
+      "console": "integratedTerminal",
+      "jinja": true,
+      "justMyCode": true,
+      "cwd": "${workspaceFolder}"
+    },
+    {
       "name": "Debug `arcade evals -d` on current file",
       "type": "python",
       "request": "launch",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -44,7 +44,7 @@
       "type": "python",
       "request": "launch",
       "program": "${workspaceFolder}/arcade/run_cli.py",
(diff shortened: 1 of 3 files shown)


arcade/arcade/cli/launcher.py
```diff
@@ -364,7 +364,7 @@ def _wait_for_healthy_actor(
 ) -> None:
     """Wait until an HTTP request to `host:port/actor/health` returns 200"""
 
-    while not actor_process.poll():  # Stop waiting if the actor process has exited
+    while actor_process.poll() is None:  # Continue waiting UNLESS the actor process has exited
```
nbarbettini, line 367: This technically fixes it. `poll()` returns the exit code, which was `0` and misinterpreted here as falsey.

`poll() is None` is a better condition that handles any reason the process will exit (zero or non-zero exit codes).

arcade/arcade/cli/serve.py
```diff
@@ -98,13 +98,12 @@ def serve_default_actor(
...
-        logger.info("Serving the following toolkits:")
-        for toolkit in toolkits:
-            num_tools = sum(len(tools) for tools in toolkit.tools.values())
-            logger.info(f"  - {toolkit.name} ({toolkit.package_name}): {num_tools} tools")
+        raise RuntimeError("No toolkits found in Python environment.")
+
+    logger.info("Serving the following toolkits:")
+    for toolkit in toolkits:
+        num_tools = sum(len(tools) for tools in toolkit.tools.values())
+        logger.info(f"  - {toolkit.name} ({toolkit.package_name}): {num_tools} tools")
```
nbarbettini, line 106: A better fix. This ensures a non-zero exit code is returned by the default actor if no toolkits are found (consistent with the fact that this is an error condition that prevents the actor from starting).

I un-indented the following lines because the `else` condition is unnecessary.

.vscode/launch.json
```diff
@@ -39,6 +39,17 @@
       "justMyCode": true,
       "cwd": "${workspaceFolder}"
     },
+    {
```
EricGustin, line 42: this config and environment path seems to be specific to your machine

EricGustin (approved):

.vscode/launch.json
nbarbettini: Fixed, thanks

PR #151, opened 2024-11-06
CLI Improvements

EricGustin:
1. Fixes bug where arcade login doesn't work for localhost
    - `arcade login -h localhost` will open login page at `http://localhost:8000/...`
    - Optionally specify the port: `arcade login -h localhost -p 8000`

3. Adds `local` flag to `arcade show`
    - `-h localhost`, `-h 127.0.0.1`, and `-h 0.0.0.0` shows the tools that are in the local engine's catalog
    - `--local` show the tools that are in the local environment.

### DIFF
--- a/arcade/arcade/cli/main.py
+++ b/arcade/arcade/cli/main.py
@@ -4,7 +4,6 @@ import threading
 import uuid
 import webbrowser
 from typing import Any, Optional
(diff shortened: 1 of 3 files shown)


nbarbettini (approved):
Thanks for adding new tests 👍

PR #152, opened 2024-11-12
Add multi-line chat and help menu

EricGustin:
# PR Description
Given that `arcade chat` is the entry point for all users, this PR makes the developer's first experience with Arcade a better experience.

1. Add multi-line & pasting support to `arcade chat`. To start a multi-line chat message, start your message with `"""`. To end the multi-line chat message, end your line with `"""`.
image

2. Added chat commands
  - `/?` displays all available chat commands that are available
  - `/show` shows all of the tools available. This uses `arcade show`. Currently no option to specify the toolkit, or tool.
  - `/clear` clears the chat conversation history of the current session
  - `/exit` exits the chat. (ctrl+c still works of course!)
  
image

### DIFF
--- a/arcade/arcade/cli/utils.py
+++ b/arcade/arcade/cli/utils.py
@@ -3,7 +3,9 @@ import ipaddress
 import os
 import webbrowser
 from dataclasses import dataclass
+from enum import Enum
 from pathlib import Path
+from textwrap import dedent
 from typing import Any, Callable, Union, cast
 from urllib.parse import urlencode, urlparse
 
@@ -34,6 +36,13 @@ class OrderCommands(TyperGroup):
     def list_commands(self, ctx: Context) -> list[str]:  # type: ignore[override]
         """Return list of commands in the order appear."""
         return list(self.commands)  # get commands using self.commands
+
+
+class ChatCommand(str, Enum):
+    HELP = "/?"
+    CLEAR = "/clear"
+    SHOW = "/show"
+    EXIT = "/exit"
 
 
 def create_cli_catalog(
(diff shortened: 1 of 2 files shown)


arcade/arcade/cli/utils.py
```diff
@@ -587,3 +596,73 @@ def delete_deprecated_config_file() -> None:
...
+    """Display the help message for arcade chat."""
+    help_message = dedent(f"""\
+        [default]
+        Available Commands:
+          {ChatCommand.SHOW:<10} Show all available tools
+          {ChatCommand.CLEAR:<10} Clear the chat history
+          {ChatCommand.HELP:<10} Help for a command
+          {ChatCommand.EXIT:<10} Exit the chat
+
+        Use \"\"\" to begin & end a multi-line message[/default]
```
nbarbettini, line 636: Curious whether this is more standard, or shift-enter. I feel like I'm more used to doing shift-enter in apps where enter sends a message.

Either way, this PR is a great improvement. 👍

nbarbettini (approved):

arcade/arcade/cli/utils.py
EricGustin: I took inspiration from Ollama for the triple quotes.

I agree Shift+Enter is nicer, but I think that would require something more involved than using the builtin `input()`

nbarbettini: Makes sense 👍

arcade/arcade/cli/utils.py
```diff
@@ -587,3 +596,73 @@ def delete_deprecated_config_file() -> None:
...
+
+        show(
+            toolkit=None,
+            tool=None,
+            host=host,
+            port=port,
+            force_tls=force_tls,
+            force_no_tls=force_no_tls,
+            debug=False,
+        )
```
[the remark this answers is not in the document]
EricGustin: Good point. Updating now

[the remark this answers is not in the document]
EricGustin: Updated to pass the `show` function itself so that an import is not needed

PR #158, opened 2024-11-19
Update starlette dep via bumping FastAPI

EricGustin:
Patches vulnerability that is due to starlette version < 0.40.0 by bumping the fastAPI version to 0.115.3

FastAPI 0.115.3 release notes: https://github.com/fastapi/fastapi/releases/tag/0.115.3

### DIFF
--- a/examples/fastapi/poetry.lock
+++ b/examples/fastapi/poetry.lock
@@ -35,23 +35,31 @@ trio = ["trio (>=0.26.1)"]
 
 [[package]]
 name = "arcade-ai"
-version = "0.1.0"
-description = ""
+version = "0.1.1"
+description = "Arcade AI Python SDK and CLI"
 optional = false
 python-versions = ">=3.10,<4.0"
 files = []
 develop = true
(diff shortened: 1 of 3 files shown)


examples/fastapi/poetry.lock
```diff
@@ -35,23 +35,31 @@ trio = ["trio (>=0.26.1)"]
 
 [[package]]
```
EricGustin, line 36: should we just .gitignore the lock file?

nbarbettini: I talked with Sam about this before. We ignore the lockfile in the SDK project(s), but not in these example apps. The thinking is, the examples are intended to be ready-to-run and in that case you do want specific versions of everything "pinned".

nbarbettini (approved):
LGTM

examples/fastapi/poetry.lock
EricGustin: Makes sense!

PR #162, opened 2024-11-24
Slack Tools

EricGustin:
implements additional tools for Slack related to retrieving conversations metadata, list of members, history of messages, as well as sending messages to private/public channels and DMs / multi-person DMs.


toolkits/slack/arcade_slack/utils.py
```diff
@@ -0,0 +1,201 @@
...
+    return message
+
+
+def convert_datetime_to_unix_timestamp(datetime_str: str) -> int:
+    """Convert a datetime string to a unix timestamp."""
+    try:
+        dt = datetime.strptime(datetime_str, "%Y-%m-%d %H:%M:%S")
+        return int(dt.timestamp())
+    except ValueError:
+        raise RetryableToolError(
```
Spartee, line 183: design question... should a ``RetryableToolError`` ever be raised outside a tool? maybe? thoughts?

toolkits/slack/evals/eval_slack_chat.py
```diff
@@ -0,0 +1,625 @@
...
+    BinaryCritic,
+    EvalRubric,
+    EvalSuite,
+    ExpectedToolCall,
+    SimilarityCritic,
+    tool_eval,
+)
+
+import arcade_slack
+from arcade_slack.tools.chat import (
```
Spartee, line 14: might be good to split this up.

toolkits/slack/evals/eval_slack_chat.py
```diff
@@ -0,0 +1,625 @@
...
+    )
+
+    suite.add_case(
+        name="Get conversation metadata by id",
+        user_message="Get the metadata of the conversation with id '1234567890'",
+        expected_tool_calls=[
+            ExpectedToolCall(
+                func=get_conversation_metadata_by_id,
+                args={
+                    "conversation_id": "1234567890",
```
Spartee, line 461: make this harder. these numbers being sequential makes this too easy and unrealistic.

toolkits/slack/evals/eval_slack_chat.py
```diff
@@ -0,0 +1,625 @@
...
+        "Get the members of the general channel",
+        "Get a list of people in the #general channel",
+        "Get a list of people in the general channel",
+        "Show me who's in the #general channel",
+        "Show me who's in the general channel",
+        "Who is in the #general channel?",
+        "Who is in the general channel?",
+    ]
+
+    for user_message in user_messages:
```
Spartee, line 494: @EricGustin use case for parameterized. I tend to think ``bulk_add_cases`` or ``add_case_batch`` would be good. arguments would accept input containers expected to be of the same length or create repeats to match passed container length if input is of 1 element

toolkits/slack/tests/test_users.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+@pytest.fixture
```
Spartee, line 4: You have this a couple times. Let's put this into a conftest.py so you can use it for all the tests.

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
+        "channels",
+        limit=limit,
+        next_cursor=next_cursor,
+        types=conversation_types_filter,
+        exclude_archived=True,
+    )
+
+    return {
+        "conversations": [
+            extract_conversation_metadata(conversation)
```
Spartee, line 184: I see why you're writing this like this, but this is hard to read. Typically I prefer making a function longer in exchange for readability and long term ability to change the pieces within this list comprehension. leaving this up to you though.

Spartee: This is totally a personal preference as what you wrote is technically more "pythonic" but for example
 
```python
return {
    "conversations": [
        extract_conversation_metadata(conversation)
        for conversation in results
        if conversation.get("is_im") or conversation.get("is_member")
    ]
}
```

vs

```python
conversations = []
for chat in results:
    if chat.get("is_im") or chat.get("is_member"):
        conversations.append(extract_conversation_metadata(chat))

return {"conversations": conversations}
```

I've found the latter is easier to edit/maintain over time. like if I need to introduce error handling for cases I didn't consider or extra methods. The comprehension only has so much space until it gets unwieldy.

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
+                "Conversation not found",
+                developer_message=f"Conversation with ID '{conversation_id}' not found.",
+                additional_prompt_content=f"Available conversations: {available_conversations}",
+                retry_after_ms=500,
+            )
+
+    # Get the members' info
+    # TODO: This will probably hit rate limits. We should probably call list_users() and
+    # then filter the results instead.
+    members = await asyncio.gather(*[
```
Spartee, line 316: what happens if members is empty?

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
 
-    # Step 1: Retrieve the list of channels
-    channels_response = slackClient.conversations_list()
-    channel_id = None
-    for channel in channels_response["channels"]:
-        if channel["name"].lower() == channel_name.lower():
-            channel_id = channel["id"]
+        conversation_next_cursor = response["next_cursor"]
+
+        for conversation in response["conversations"]:
```
Spartee, line 345: KEY ERRORRRRRRR error handling :)

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
+) -> Annotated[
+    dict,
+    (
+        "The conversation history / messages and next cursor for paginating results (when "
+        "there are additional messages to retrieve)."
+    ),
+]:
+    """Get the history of a conversation in Slack."""
+    error_message = None
+    if oldest_datetime and oldest_relative:
```
Spartee, line 422: is this datetime handling useful anywhere else? in the future? 

if so, let's extract into a function

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
+# can avoid exposing this arg to the LLM.
+@tool(
+    requires_auth=Slack(
+        scopes=["channels:history", "groups:history", "im:history", "mpim:history"],
+    )
+)
+async def get_conversation_history_by_name(
+    context: ToolContext,
+    conversation_name: Annotated[str, "The name of the conversation to get history for"],
+    oldest_relative: Annotated[
```
Spartee, line 486: llm will do better if named ``oldest_relative_message_time``. always try to be as literal as possible. same goes for the rest of these.

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
+        Optional[str],
+        (
+            "The oldest message to include in the results, specified as a datetime object in the "
+            "format 'YYYY-MM-DD HH:MM:SS'"
+        ),
+    ] = None,
+    latest_datetime: Annotated[
+        Optional[str],
+        (
+            "The latest message to include in the results, specified as a datetime object in the "
```
Spartee, line 510: is it a datetime object? "datetime string"?

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
+            include_num_members=True,
         )
 
-    # Step 2: Send the message to the channel
-    response = slackClient.chat_postMessage(channel=channel_id, text=message)
-    response.validate()
+        return extract_conversation_metadata(response["channel"])
+
+    except SlackApiError as e:
+        if e.response["error"] == "channel_not_found":
```
Spartee, line 570: does the response always have an error in this case?

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
+        "The conversation metadata. Metadata does not include: "
+        "- members, users, or people on the conversation / channel; "
+        "- history / messages of the conversation / channel."
+    ),
+]:
+    """Get the metadata of a conversation in Slack."""
+    next_cursor = None
+    conversation_names = []
+
+    while True:
```
Spartee, line 605: no. antipattern. relying on breaks scares me. change to a variable and set. ideally with some kind of timeout or secondary fallback to escape the scope.

toolkits/slack/arcade_slack/tools/users.py
```diff
@@ -0,0 +1,60 @@
...
+    next_page_token = None
+
+    while limit == -1 or len(users) < limit:
+        iteration_limit = (
+            200 if limit == -1 else min(limit - len(users), 200)
+        )  # Slack recommends max 200 results at a time
+        response = await slackClient.users_list(cursor=next_page_token, limit=iteration_limit)
+        if response.get("ok"):
+            for user in response.get("members", []):
+                if not is_user_deleted(user) and (not exclude_bots or not is_user_a_bot(user)):
```
Spartee, line 52: need more comments on functions like this. 10 months from now you won't remember what this does.

toolkits/slack/arcade_slack/utils.py
```diff
@@ -0,0 +1,201 @@
...
+    **kwargs,
+) -> tuple[list, Optional[str]]:
+    """Paginate a Slack AsyncWebClient's function results.
+
+    The purpose is to abstract the pagination work and make it easier for the LLM to retrieve the
+    amount of items requested by the user, regardless of limits imposed by the Slack API. We still
+    return the next cursor, if needed to paginate further.
+    """
+    results = []
+    while len(results) < limit:
```
Spartee, line 139: what if results is longer than the max pagination limit?

Spartee (commented):
Lots of comments mostly to share how I think about development. Ask me Q that come up as you read this.

Spartee:
@EricGustin @byrro let's make this ready for review and update the description

byrro:
> @EricGustin @byrro let's make this ready for review and update the description
> 

It's almost all done. Should submit the PR today.

toolkits/slack/arcade_slack/tools/users.py
byrro: True, I actually forgot to apply a refactor to this one. Will do a clean-up before submitting it for review.

toolkits/slack/arcade_slack/utils.py
byrro: It's not going to be.

The `slack_limit` variable makes the Slack API return no more than `min(limit - len(results), MAX_PAGINATION_LAIMIT)`.

Then, it extend results with the items returned and continue looping if it hasn't reached the limit, and slack indicates there's more to retrieve.

toolkits/slack/arcade_slack/utils.py
byrro: That got me thinking, too.

The reason I ended up doing it is because we'd nee to repeat this except block twice in a tool, making it unnecessarily longer and a bit harder to read. This makes the tool cleaner. 

I think it's a fair trade-off, and since those are tool util, not general-purpose, they're expected to be called within a tool execution runtime anyways...

toolkits/slack/arcade_slack/utils.py
Spartee: Good answer. Write this as comments in the code

toolkits/slack/arcade_slack/utils.py
```diff
@@ -0,0 +1,201 @@
...
+
+        next_cursor = response.get("response_metadata", {}).get("next_cursor")
+
+        if not next_cursor:
+            break
+
+    return results, next_cursor
+
+
+def enrich_message_metadata(message: dict) -> dict:
```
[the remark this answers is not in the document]
byrro: no, removed it

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +91,532 @@ def send_dm_to_user(
...
+        "there are additional messages to retrieve)."
+    ),
+]:
+    """Get the history of a conversation in Slack."""
+    conversation_metadata = await get_conversation_metadata_by_name(
+        context=context, conversation_name=conversation_name
+    )
+    return await get_conversation_history_by_id(
+        context=context,
+        conversation_id=conversation_metadata["id"],
```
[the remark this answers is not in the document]
byrro: the `"id"` (and `team_id`, when applicable) is one field that slack will always return.. most other fields will depend on which endpoint we're requesting

the `conversation_metadata` in the code you highlighted comes from the `conversations.list` endpoint (https://api.slack.com/methods/conversations.list), which returns a `conversation` type (https://api.slack.com/types/conversation). the most basic parameter it has is the `id`.

toolkits/slack/arcade_slack/tools/chat.py
byrro: In the context of a `SlackApiError` raised, yes. But I'm changing it to `.get("error")` just to be extra sure our error handling won't break in case they change their API and someone is still running an old version of our toolkit.

toolkits/slack/arcade_slack/tools/chat.py
byrro: fair. should we use timeouts in all iterations that have the potential to run endlessly?

I can't really enforce a timeout without using threads, though (which is probably undesirable here in terms of perf), but I think a naive timeout implementation would suffice

byrro: it's done. replaced breaks with a variable check. used async timeout to make sure loops doesn't run past a certain amount of time. since all api calls are asynchronous, we don't actually need a thread to enforce the timeout. thank the async gods 🙏

toolkits/slack/tests/test_users.py
byrro: actually, we need to mock the class in two different modules: chat and users. to have a single fixture, we'd need to parameterize the location on each test case. not good. better to have two fixtures, I moved it from conftest to each of the testing modules, with the respective class location.

byrro:
> > @EricGustin @byrro let's make this ready for review and update the description
> 
> It's almost all done. Should submit the PR today.

@EricGustin @Spartee I finished implementing / refactoring all tools with pytests and evals. also made the changes requested by @Spartee  today.

need top discuss with you about evals, how can I improve them, write more adversarial evaluations. let's discuss tomorrow at the office.

the current interface for date/time filtering in get_conversation_history (by id or name) is rendering flaky behavior by the LLM. most of the time it nails, but I'd say it makes some mistakes in 5-10% of runs.

might be out of the scope of this PR, but we need to improve our interface for datetime and timedelta arguments. other toolkits would benefit too.

toolkits/slack/arcade_slack/utils.py
```diff
@@ -0,0 +1,502 @@
...
+Because of that, our TypedDicts ended up having to be mostly total=False and most of the fields'
+type hints are Optional. Use Slack dictionary fields with caution. It's advisable to validate the
+value before using it, so that we can raise errors that are clear to understand.
+
+See, for example, the 'Common Fields' section of the user type definition at:
+https://api.slack.com/types/user#fields (https://archive.is/RUZdL)
+"""
+
+
+class SlackUserFieldData(TypedDict, total=False):
```
byrro, line 32: @Spartee please take a look at the dict typing to see if it's in line with what you have in mind

toolkits/slack/arcade_slack/utils.py
```diff
@@ -0,0 +1,502 @@
...
+                    )
+                except KeyError:
+                    raise ValueError(f"Response key {response_key} not found in Slack response")
+
+                next_cursor = response.get("response_metadata", {}).get("next_cursor")
+
+                if len(results) >= limit or not next_cursor:
+                    should_continue = False
+    except asyncio.TimeoutError:
+        raise PaginationTimeoutError(max_pagination_timeout_seconds)
```
byrro, line 429: @Spartee I'm unsure about the best approach to handle timeouts here... ideas?

toolkits/slack/arcade_slack/utils.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def enrich_message_datetime(message: dict) -> dict:
+    """Enrich message metadata with formatted datetime.
+
+    It helps LLMs when they need to display the date/time in human-readable format. Slack
+    will only return a unix-formatted timestamp (it's not actually unix ts, but the unix ts
+    in the user's timezone - I know, odd, but it is what it is).
+    """
+    ts = message.get("ts")
```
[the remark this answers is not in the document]
byrro: Slack guarantees nothing about the presence or the type of fields returned in their dicts. it varies a lot depending on which endpoint we're calling as well.

since these utilities will be called by several tools, each querying different endpoints, I think we should lean on being more relaxed.

need to check for types, though. in the `ts` case, I check whether it's a valid float-like string before generating the datetime str.

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +92,508 @@ def send_dm_to_user(
...
+                for conversation in conversations["conversations"]
+            )
 
-    if not channel_id:
-        raise RetryableToolError(
-            "Channel not found",
-            developer_message=f"Channel with name '{channel_name}' not found.",
-            additional_prompt_content=format_channels(channels_response),
-            retry_after_ms=500,  # Play nice with Slack API rate limits
+            raise RetryableToolError(
```
EricGustin, line 170: I'm curious how beneficial it is to use a RetryableToolError for this tool. For names, it can be helpful to use because it's obvious that "Johhn" should be retried as "John", but with IDs, I'm not sure if this applies.

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +92,508 @@ def send_dm_to_user(
...
+
+    return {"members": members, "next_cursor": next_cursor}
+
+
+@tool(
+    requires_auth=Slack(
+        scopes=["channels:read", "groups:read", "im:read", "mpim:read"],
+    )
+)
+async def get_members_from_conversation_by_name(
```
EricGustin, line 192: Tested with non-existing conversation name: "get the members in the slack conversation 'ggeneral'". Expected retryable tool error to correct it, but got the following instead.

Error: `Tool Slack.GetMembersFromConversationByName version 0.1.8 failed. Error in get_conversation_metadata_by_name: 'NoneType' object has no attribute 'lower'`

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +92,508 @@ def send_dm_to_user(
...
+
+
+# Note: Bots are included in the results.
+# Note: Direct messages with no conversation history are included in the results.
+@tool(
+    requires_auth=Slack(
+        scopes=["im:read"],
+    )
+)
+async def list_direct_message_channels_metadata(
```
EricGustin, line 587: Example response: `{"conversations":[{"conversation_type":"im","id":"D088LB2QP7T","is_archived":false,"is_member":true,"is_private":true,"is_user_deleted":false,"name":null,"num_members":2,"purpose":"","user":"U087T4BC415"}}`

Question: would it be reasonable to get the username and other info for each person I have a dm with? Would this be too expensive?

EricGustin (commented):
Just finished up trying out your tools. I've left some comments on my experience using this toolkit

toolkits/slack/evals/eval_users.py
```diff
@@ -0,0 +1,161 @@
...
+        ],
+    )
+
+    suite.add_case(
+        name="test list users with pagination",
+        user_message="get the next 5 users",
+        expected_tool_calls=[
+            ExpectedToolCall(
+                func=list_users,
+                args={"limit": 5, "next_cursor": "cursor_xyz"},
```
EricGustin, line 106: naming a cursor token "cursor_xyz" seems too easy. Use a legitimate looking cursor such as "dXNlcjpVsDjzOTZGVDlQRA=="

toolkits/slack/evals/eval_chat.py
```diff
@@ -0,0 +1,1020 @@
...
+    list_direct_message_channels_metadata,
+    list_group_direct_message_channels_metadata,
+    list_private_channels_metadata,
+    list_public_channels_metadata,
+    send_dm_to_user,
+    send_message_to_channel,
+)
+
+
+class RelativeTimeBinaryCritic(BinaryCritic):
```
EricGustin, line 34: cool! We should think about the best file location for custom critics

toolkits/slack/evals/eval_chat.py
```diff
@@ -0,0 +1,1020 @@
...
+        (
+            "List public channels and direct message conversations",
+            "List public channels and direct message conversations I am a member of",
+            list_public_channels_metadata,
+            list_direct_message_channels_metadata,
+        ),
+    ]
+
+    for name, user_message, *expect_called_tool_functions in cases:
+        suite.add_case(
```
EricGustin, line 449: there should be a critic to make sure the model isn't predicting something like "limit": 5 since these cases are trying to get all conversations.

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +92,508 @@ def send_dm_to_user(
...
+    requires_auth=Slack(
+        scopes=["channels:read", "groups:read", "im:read", "mpim:read"],
     )
+)
+async def get_members_from_conversation_by_id(
+    context: ToolContext,
+    conversation_id: Annotated[str, "The ID of the conversation to get members for"],
+    limit: Annotated[
+        Optional[int], "The maximum number of members to return."
+    ] = MAX_PAGINATION_SIZE_LIMIT,
```
EricGustin, line 148: If I want to "list all members in the conversation that has ID 12345", then wouldn't defaulting to a limited number MAX_PAGINATION_SIZE_LIMIT prevent the tool from getting all of the members if there were more than 200 in the conversation?

I wonder if defaulting to `-1` and treating that as infinite (don't stop until next_cursor is None) would be better.

toolkits/slack/evals/eval_chat.py
```diff
@@ -0,0 +1,1020 @@
...
+        catalog=catalog,
+        rubric=rubric,
+    )
+
+    no_arguments_user_messages_by_conversation_name = [
+        "Get the history of the #general channel",
+        "Get the history of the general channel",
+        "list the messages in the #general channel",
+        "list the messages in the general channel",
+    ]
```
EricGustin, line 581: This is missing getting multi-person DMs and DMs by name. How does it perform on conversations that are not channels?

toolkits/slack/evals/eval_chat.py
```diff
@@ -0,0 +1,1020 @@
...
+            expected_tool_calls=[
+                ExpectedToolCall(
+                    func=get_conversation_history_by_name,
+                    args={
+                        "conversation_name": "general",
+                    },
+                ),
+            ],
+            critics=[
+                BinaryCritic(critic_field="conversation_name", weight=1.0),
```
EricGustin, line 596: Would be good to add some critics for the different timestamp parameters. For example, this eval would pass if it predicted something as ridiculous as `oldest_datetime="1738-07-04 12:12:12"` and `oldest_datetime="1738-08-04 12:12:12"`.

EricGustin (commented):
Just went through the evals. Leaving comments below

toolkits/slack/arcade_slack/constants.py
```diff
@@ -0,0 +1,29 @@
+import os
+
+
+class PositiveInt(int):
```
EricGustin, line 4: constants.py might not be the best place for this. Maybe models.py or types.py ?

toolkits/slack/arcade_slack/tools/chat.py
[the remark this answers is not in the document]
byrro: I think so too, applying that

toolkits/slack/evals/eval_chat.py
byrro: moved to a dedicated module at ./arcade_slack/critics.py

toolkits/slack/arcade_slack/tools/chat.py
byrro: modified all tools to set limit to `None` by default, and refactored `async_paginate` accordingly

toolkits/slack/evals/eval_chat.py
byrro: we actually cannot get message history from DMs or multi-person DMs yet. will need to implement separate tools for that. renamed the tool to `get_messages_in_channel_by_name` to better reflect what it does.

toolkits/slack/arcade_slack/constants.py
byrro: yea, moved to models already

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -72,43 +92,508 @@ def send_dm_to_user(
...
+
+# TODO: make the function accept a current unix timestamp argument to allow testing without
+# mocking. Have to wait until arcade.core.annotations.Inferrable is implemented, so that we
+# can avoid exposing this arg to the LLM.
+@tool(
+    requires_auth=Slack(
+        scopes=["channels:history", "groups:history", "im:history", "mpim:history"],
+    )
+)
+async def get_conversation_history_by_name(
```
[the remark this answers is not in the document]
byrro: will need to implement separate tools for that. for the sake of documentation of our talk: DM and multi-person DM don't actually have a name in Slack. we need a tool that:

1. receives the names of the user(s) which is(are) member(s) of the conversation
2. get the user IDs
3. retrieve all DM or multi-person DM conversations (depending if it's one or more names) and check which is the one we're looking for
4. retrieve the messages from that conversation

toolkits/slack/arcade_slack/tools/chat.py
byrro: fixed

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -18,49 +34,53 @@
...
-    return {}
+        # Step 3: Send the message as if it's from you (because we're using a user token)
+        response = await slackClient.chat_postMessage(channel=dm_channel_id, text=message)
+
+    except SlackApiError as e:
+        error_message = e.response["error"] if "error" in e.response else str(e)
+        raise ToolExecutionError(
+            "Error sending message",
+            developer_message=f"Slack API Error: {error_message}",
+        )
```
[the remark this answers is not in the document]
byrro: I think the benefit here is passing over the value in `e.response["error"]`, which comes from the Slack API. I can see it being useful in debugging certain cases. that message may not be helpful, like in the case you highlighted, it might be in other cases, perhaps? I think it's better to keep it (no harm in doing so) than risk losing valuable debugging data. makes sense?

byrro (approved):
all good! nice job, @EricGustin!

EricGustin:
> all good! nice job, @EricGustin!

haha!

PR #167, opened 2024-11-27
Add Atlassian as a supported provider.

wdawson:


EricGustin (approved):
Nice - these are coming in fast!

PR #168, opened 2024-12-03
Bump arcadepy dependency

nbarbettini:
`arcadepy` is now at [0.2.1 on pypi](https://pypi.org/project/arcadepy/0.2.1/).
The SDK references `arcadepy` but it is locked to the `0.1.x` minor release.


nbarbettini:
@sdreyer and @EricGustin for FYI. This is just a dependency version bump

PR #169, opened 2024-12-07
Add next_token to X Search tools

EricGustin:
# PR Description
Adds an optional `next_token` input parameter to the `X.SearchRecentTweetsByUsername` and `X.SearchRecentTweetsByKeywords` tools. 

This allows users to paginate through tweets. A `next_token` is provided in the tools's response. 

For example, to access the `next_token` when using the `tools.execute`, you can do `next_token = response.output.value["meta"].get("next_token", None)` and then pass it to the tool on your next call through the tools' `next_token` input parameter.

### DIFF
--- a/toolkits/x/arcade_x/tools/tweets.py
+++ b/toolkits/x/arcade_x/tools/tweets.py
@@ -1,4 +1,4 @@
-from typing import Annotated, Any
+from typing import Annotated, Any, Optional
 
 import httpx
 from arcade.sdk import ToolContext, tool
@@ -10,6 +10,7 @@ from arcade_x.tools.utils import (
     get_headers_with_token,
     get_tweet_url,
     parse_search_recent_tweets_response,
+    remove_none_values,
 )
 
 TWEETS_URL = "https://api.x.com/2/tweets"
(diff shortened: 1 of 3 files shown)


toolkits/x/arcade_x/tools/tweets.py
```diff
@@ -63,17 +64,25 @@ async def search_recent_tweets_by_username(
     context: ToolContext,
     username: Annotated[str, "The username of the X (Twitter) user to look up"],
     max_results: Annotated[
-        int, "The maximum number of results to return. Cannot be less than 10"
+        int, "The maximum number of results to return. Must be in range [1, 100] inclusive"
     ] = 10,
+    next_token: Annotated[
+        Optional[str], "The pagination token starting from which to return results"
+    ] = None,
```
nbarbettini, line 71: Do we need to explain anything here to reduce the likelihood of the LM hallucinating a value for this?

nbarbettini (approved):
Approved with question.
For follow-up -- new evals?

toolkits/x/arcade_x/tools/tweets.py
EricGustin: From my experience, even for older models, it errs on the side of caution and will either predict the `next_token` param correctly, or it doesn't provide the token when it should.

EricGustin:
Added eval @nbarbettini

PR #171, opened 2024-12-13
X Toolkit: Support tweets longer than 280 characters

EricGustin:
# PR Description
* Update `search_recent_tweets_by_username`, `search_recent_tweets_by_keywords`, and `lookup_tweet_by_id` to support long tweets. Previously, only the first 280 characters of the tweet's text were returned by the tool.

### DIFF
--- a/toolkits/x/arcade_x/tools/utils.py
+++ b/toolkits/x/arcade_x/tools/utils.py
@@ -55,6 +55,18 @@ def sanity_check_tweets_data(tweets_data: dict[str, Any]) -> bool:
     return True
 
 
+def expand_long_tweet(tweet_data: dict[str, Any]) -> dict[str, Any]:
+    """Expand a long tweet.
+
+    For tweets exceeding 280 characters,
+    replace the truncated tweet text with the full tweet text.
+    """
+    if tweet_data.get("note_tweet"):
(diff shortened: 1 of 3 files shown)


toolkits/x/arcade_x/tools/utils.py
```diff
@@ -55,6 +55,18 @@ def sanity_check_tweets_data(tweets_data: dict[str, Any]) -> bool:
...
+def expand_long_tweet(tweet_data: dict[str, Any]) -> dict[str, Any]:
+    """Expand a long tweet.
+
+    For tweets exceeding 280 characters,
+    replace the truncated tweet text with the full tweet text.
+    """
+    if tweet_data.get("note_tweet"):
+        tweet_data["text"] = tweet_data["note_tweet"]["text"]
+        del tweet_data["note_tweet"]
+    return tweet_data
```
nbarbettini, line 67: If I'm reading it right, it modifies the dict that was passed in, and then returns that modified dict. It would be a little clearer to either copy and modify, or just don't return anything (which implies to the caller that the input is modified)

EricGustin: updated & opted for in place replacement only

nbarbettini (approved):
LGTM

PR #172, opened 2024-12-13
`arcade evals` Default to localhost

EricGustin:
# PR Description
* `arcade evals` now run evaluations against Arcade Engine at `http://localhost:9099` by default.
* Added optional flag `--cloud` to run evaluations against Arcade's Cloud Engine at `https://api.arcade-ai.com`. Overrides `-h` flag.
* Always print the Engine that the evaluations are using. Previously this was reserved for `-d` flag.

### DIFF
--- a/arcade/arcade/cli/main.py
+++ b/arcade/arcade/cli/main.py
@@ -345,10 +345,15 @@ def evals(
         help="The models to use for evaluation (default: gpt-4o)",
     ),
     host: str = typer.Option(
-        DEFAULT_ENGINE_HOST,
+        "localhost",
         "-h",
         "--host",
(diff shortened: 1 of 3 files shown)


arcade/arcade/cli/main.py
```diff
@@ -380,13 +387,12 @@ def evals(
     if not eval_files:
         return
 
-    if show_details:
-        console.print(
-            Text.assemble(
-                ("\nRunning evaluations against Arcade Engine at ", "bold"),
-                (base_url, "bold blue"),
-            )
+    console.print(
```
EricGustin, line 390: seems important enough to always show this

PR #176, opened 2024-12-18
Release Toolkits Individually

sdreyer:


.github/workflows/publish-toolkit.yml
```diff
@@ -0,0 +1,73 @@
...
+
+    - name: Install Python Dependencies
+      run: pip install ./arcade
+
+    - name: Test Toolkit
+      id: Test_Toolkit
+      working-directory: toolkits/${{ steps.set-toolkit.outputs.toolkit }}
+      run: |
+        make install
+        make test
```
EricGustin, line 50: should we add `make check` as well?

sdreyer: yeah probably

EricGustin: or maybe that should be done before merging to main?

sdreyer: we should probably have another workflow that tests and checks them all on commit

EricGustin (approved):

PR #177, opened 2024-12-18
Fix `/show` for Cloud Engine

EricGustin:
# PR Description
### The following bug was observed: 
* When connected to the cloud engine for `arcade chat`, and the user types `/show`, then the local environment tools are displayed. Instead, the cloud engine's tools should be displayed.

### Why was this bug happening?:
* When a user entered the `/show` command, the CLI Command `show` was being called directly. Since the function was a CLI command, the `local` parameter was not being processed and resolved to its intended value because the Typer CLI interface was being bypassed. So, the conditional `if local:` would always evaluate to `True`.

### How this was fixed:
* I created a wrapper function for the `show` CLI Command. Now, when the user types `/show`, then the wrapper function is called instead of the `show` CLI command. This ensures that all input parameters are resolved to their intended values.

### DIFF
--- a/arcade/arcade/cli/main.py
+++ b/arcade/arcade/cli/main.py
@@ -18,19 +18,16 @@ from arcade.cli.constants import DEFAULT_CLOUD_HOST, DEFAULT_ENGINE_HOST, LOCALH
 from arcade.cli.display import (
     display_arcade_chat_header,
     display_eval_results,
-    display_tool_details,
     display_tool_messages,
-    display_tools_table,
 )
 from arcade.cli.launcher import start_servers
+from arcade.cli.show import show_logic
 from arcade.cli.utils import (
     OrderCommands,
     compute_engine_base_url,
     compute_login_url,
-    create_cli_catalog,
     delete_deprecated_config_file,
     get_eval_files,
-    get_tools_from_engine,
     get_user_input,
     handle_chat_interaction,
     handle_tool_authorization,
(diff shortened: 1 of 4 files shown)


arcade/tests/cli/test_show.py
```diff
@@ -0,0 +1,40 @@
+from unittest.mock import patch
+
+from arcade.cli.show import show_logic
+
+
+def test_show_logic_local_false():
```
EricGustin, line 6: seems like a basic test, but this would have been failing before this PR!

arcade/arcade/cli/main.py
```diff
@@ -177,38 +174,7 @@ def show(
...
-        else:
-            # Display the list of tools as a table
-            display_tools_table(tools)
-
-    except Exception as e:
-        if debug:
-            raise
-        error_message = f"❌ Failed to list tools: {escape(str(e))}"
-        console.print(error_message, style="bold red")
+    show_logic(toolkit, tool, host, local, port, force_tls, force_no_tls, debug)
```
EricGustin, line 177: Moved logic to a wrapper function so that the CLI command can be called directly without any unintended consequences such as unresolved parameters.

sdserranog (approved):
It works great! 🚀

arcade/arcade/cli/utils.py
```diff
@@ -671,6 +671,7 @@ def handle_user_command(
             toolkit=None,
             tool=None,
             host=host,
+            local=False,
```
nbarbettini, line 674: What if I am actually chatting with localhost? e.g. `arcade chat -h`

EricGustin: In this case, the `host=localhost`.

`local`: display tools in the local environment (never used for arcade chat)
`host=localhost`: display tools that the local actor is serving

nbarbettini (approved):

PR #181, opened 2024-12-19
Remove lock flag from check

sdreyer:


toolkits/math/Makefile
```diff
@@ -46,7 +46,7 @@ bump-version: ## Bump the version in the pyproject.toml file
 .PHONY: check
 check: ## Run code quality tools.
 	@echo "🚀 Checking Poetry lock file consistency with 'pyproject.toml': Running poetry check --lock"
```
EricGustin, line 48: omit --lock in the echo

EricGustin (approved):

PR #182, opened 2024-12-19
Test toolkits

sdreyer:


.github/workflows/test-toolkits.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Install Poetry
+      uses: snok/install-poetry@v1
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+        cache: 'pip'
+
+    - name: Install Python Dependencies
```
EricGustin, line 48: Since arcade-ai and toolkits will no longer be released at the same time, I think we should install the latest release of arcade-ai.

Consider the case where I create a new toolkit that uses a new function in the SDK that hasn't been released yet. The toolkit would pass the tests, but then once released, it would fail whenever someone tries to use the tools.

sdreyer: true, you wanna change that?

PR #191, opened 2025-01-07
Return media attachments metadata when retrieving X tweets

byrro:
Modifies X tweet tools to return metadata about media attachments (photo, GIF or video) when retrieving a tweet by ID, username or keywords.

The tool will always return media attachments by default. Since it's only metadata, it shouldn't add significant network overhead to existing implementations of the tool.

My guess is more often than not people will want this info included. When not needed, it doesn't hurt to include by default. It'd be annoying to have to ask the LLM to include it every time they need.

### DIFF
--- a/toolkits/x/arcade_x/tools/tweets.py
+++ b/toolkits/x/arcade_x/tools/tweets.py
@@ -6,6 +6,7 @@ from arcade.sdk.auth import X
 from arcade.sdk.errors import RetryableToolError
 
 from arcade_x.tools.utils import (
+    expand_attached_media,
     expand_long_tweet,
     expand_urls_in_tweets,
     get_headers_with_token,
@@ -64,6 +65,9 @@ async def delete_tweet_by_id(
 async def search_recent_tweets_by_username(
     context: ToolContext,
     username: Annotated[str, "The username of the X (Twitter) user to look up"],
+    return_attachments_metadata: Annotated[
+        bool, "Whether to return metadata about the media attached to the tweet"
+    ] = True,
     max_results: Annotated[
         int, "The maximum number of results to return. Must be in range [1, 100] inclusive"
     ] = 10,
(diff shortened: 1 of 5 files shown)


toolkits/x/arcade_x/tools/utils.py
```diff
@@ -137,3 +137,25 @@ def remove_none_values(params: dict) -> dict:
...
+    params["tweet.fields"] += ",attachments"
+    params["media.fields"] = ",".join([
+        # media_key, url and type are returned by default, added here for clarity
+        "media_key",
+        "url",
+        "type",
+        "duration_ms",
+        "height",
+        "width",
+        "url",
```
EricGustin, line 156: `url` is defined twice in this list which is causing a 400 response from X for me

toolkits/x/evals/eval_x_tools.py
```diff
@@ -78,9 +77,9 @@ def x_eval_suite() -> EvalSuite:
             "at Arcade AI!'"
         ),
         expected_tool_calls=[
-            ExpectedToolCall(
-                func=post_tweet,
-                args={
+            (
```
EricGustin, line 80: We are in the process of moving to the `ExpectedToolCall(...)` syntax. This can be reverted back

toolkits/x/arcade_x/tools/tweets.py
```diff
@@ -151,13 +161,16 @@ async def search_recent_tweets_by_keywords(
...
     params = remove_none_values(params)
 
-    url = (
-        "https://api.x.com/2/tweets/search/recent?"
-        "expansions=author_id&user.fields=id,name,username,entities&tweet.fields=entities,note_tweet"
-    )
+    if return_attachments_metadata:
+        expand_attached_media(params)
+
+    url = f"{TWEETS_URL}/search/recent"
```
EricGustin, line 173: thanks for the cleanup

toolkits/x/arcade_x/tools/tweets.py
```diff
@@ -64,6 +65,9 @@ async def delete_tweet_by_id(
 async def search_recent_tweets_by_username(
     context: ToolContext,
     username: Annotated[str, "The username of the X (Twitter) user to look up"],
+    return_attachments_metadata: Annotated[
```
EricGustin, line 68: nit: since the payload of this is minimal we could forego this being a parameter and just always return it. I'll leave it up to you though!

EricGustin (changes requested):
Looking great so far! A couple comments, but otherwise looks great!

EricGustin (approved):
LGTM

EricGustin (changes requested):
bump version to 0.1.9

EricGustin (approved):
looks good!

PR #193, opened 2025-01-07
Pin poetry to 1.8.5

EricGustin:
# PR Description
Poetry released v2 with many breaking changes a couple days ago. The `install-poetry` action that our workflows use default to that v2 version, so many of our workflows are failing. This PR forces that action to use poetry version 1.8.5 and also uses 1.8.5 for toolkits

A ticket to migrate to 2.0.0 has been filed for future work


nbarbettini (approved):
LGTM

PR #195, opened 2025-01-08
Least Privileged Scope for Update Calendar

EricGustin:


Spartee:
Thanks for doing this. QQ. why are you updating the version here?

nbarbettini (approved):

PR #196, opened 2025-01-08
Remove tools relying on deprecated Spotify endpoints

byrro:
Spotify deprecated several endpoints in Nov, 2024. Two of them were being used in Tracks tools. We're removing those from the Toolkit.

Spotify announcement: https://developer.spotify.com/blog/2024-11-27-changes-to-the-web-api
Archive: https://archive.is/LMBe5


EricGustin:
make sure related evals/tests and removed too

EricGustin (approved):
Sad to see these go, but good catch on Spotify's deprecation

PR #197, opened 2025-01-09
Fix 'Check Toolkits' Workflow

EricGustin:
# PR Description
The 'Check Toolkits' workflow was failing if no toolkits were changed. This PR gracefully exits the workflow for this case.

### DIFF
--- a/.github/workflows/check-toolkits.yml
+++ b/.github/workflows/check-toolkits.yml
@@ -34,7 +34,7 @@ jobs:
         dirs=$(echo "${CHANGED_FILES}" | tr ' ' '\n' | grep "toolkits/" | cut -d'/' -f2 | sort -u)
         if [ -n "$dirs" ]; then
           echo "$dirs" | while read -r dir; do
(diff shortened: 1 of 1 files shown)


.github/workflows/check-toolkits.yml
```diff
@@ -32,7 +32,11 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.PAT }}
       run: |
         dirs=$(echo "${CHANGED_FILES}" | tr ' ' '\n' | grep "toolkits/" | cut -d'/' -f2 | sort -u)
-        echo "$dirs" | while read -r dir; do
-          echo "$dir"
-          gh workflow -R ArcadeAI/arcade-ai run "Publish Toolkit" -f toolkit=${dir}
-        done
+        if [ -n "$dirs" ]; then
```
EricGustin, line 35: added a check to see if `dirs`, which contains all of the changed toolkits, is empty. If it is empty, then gracefully exit

nbarbettini (approved):
LGTM

PR #198, opened 2025-01-09
Check if toolkit version changed before attempting publish

EricGustin:
# PR Description
Changes to a toolkit without changes to the toolkit's version fail the 'Publish Toolkit' workflow with `HTTP Error 400: File already exists ('arcade_zoom-0.1.7.tar.gz', with blake2_256 hash '02183cda607f06616e7edb17e3d22bc11d1d83b074b3e44066b78ec72602fb37'). See https://pypi.org/help/#file-name-reuse for more information.`, for example.

This PR adds the `--skip-existing` flag to `poetry publish` to avoid attempting to publish an existing version. Skips slack notification if publish is skipped.

The `grep`'d string comes from https://github.com/python-poetry/poetry/blob/main/src/poetry/publishing/uploader.py#L246-L249

### DIFF
--- a/.github/workflows/publish-toolkit.yml
+++ b/.github/workflows/publish-toolkit.yml
@@ -49,19 +49,33 @@ jobs:
         make check
         make test
 
+    - name: Check Version
+      id: Check_Version
+      working-directory: toolkits/${{ steps.set-toolkit.outputs.toolkit }}
+      run: |
+        VERSION=$(poetry version -s)
+        echo "version=$VERSION" >> $GITHUB_OUTPUT
+
+        PACKAGE=$(poetry version | cut -d' ' -f1)
+
+        if pip index versions $PACKAGE | grep -q "^$VERSION$"; then
+          echo "Version $PACKAGE == $VERSION already exists on PyPI. Skipping publish."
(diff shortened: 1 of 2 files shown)


toolkits/math/evals/eval_math_tools.py
```diff
@@ -25,7 +25,7 @@
 def math_eval_suite():
     suite = EvalSuite(
         name="Math Tools Evaluation",
-        system_message="You are an AI assistant with access to math tools. Use them to help the user with their math-related tasks.",
+        system_message="You're an AI assistant with access to math tools. Use them to help the user with their math-related tasks.",
```
EricGustin, line 28: adding this minor change to ensure the 'Check Version' step that I added is working as intended.

.github/workflows/publish-toolkit.yml
```diff
@@ -49,19 +49,33 @@ jobs:
         make check
         make test
 
+    - name: Check Version
```
EricGustin, line 52: if toolkit version is already on PyPi, then do not publish to PyPi

nbarbettini (approved):

PR #201, opened 2025-01-10
Update `arcade new` template to pass linting

EricGustin:
# PR Description
A new toolkit generated via `arcade new` would not initially pass the pre commit checks. This PR resolves this issue.


Spartee (approved):
thank you

PR #203, opened 2025-01-13
Update fastapi example

EricGustin:


examples/fastapi/poetry.lock
EricGustin, line 1: would it be too much to ask the user to run `poetry install` so that we don't need to maintain this lock file?

PR #204, opened 2025-01-14
Pytests for Spotify tools

byrro:


toolkits/spotify/arcade_spotify/tools/search.py
```diff
@@ -38,8 +40,11 @@ async def search(
...
-    )
-    response.raise_for_status()
-    return dict(response.json())
+    try:
+        response = await send_spotify_request(
+            context, "GET", url, params={"q": q, "type": ",".join(types), "limit": limit}
+        )
+        response.raise_for_status()
+        return dict(response.json())
+    except httpx.HTTPStatusError as e:
```
EricGustin, line 49: The [`@tool` decorator](https://github.com/ArcadeAI/arcade-ai/blob/main/arcade/arcade/sdk/tool.py#L28-L41) already catches the error and raises a `ToolExecutionError`. The current practice for our tools is to use `raise_for_status()`, and then let the tool decorator catch the exception instead of having to explicitly try/catch inside each tool.

toolkits/spotify/tests/test_player.py
```diff
@@ -0,0 +1,652 @@
...
+    mock_response.status_code = 204
+    mock_httpx_client.request.return_value = mock_response
+
+    response = await get_currently_playing(context=tool_context)
+
+    assert response["is_playing"] is False
+
+
+@pytest.mark.asyncio
+async def test_get_currently_playing_too_many_requests_error(tool_context, mock_httpx_client):
```
EricGustin, line 509: might be worth exploring if we could parameterize some of these tests. For example, parameterize by tool name for all of the tests that return 429.

Spartee (commented):
LGTM

toolkits/spotify/arcade_spotify/tools/search.py
EricGustin: This is up for debate though

toolkits/spotify/tests/test_player.py
byrro: There's some of repetition indeed. Pushed parameterization for a few tests

EricGustin (approved):
🚀

PR #207, opened 2025-01-16
Removed unused param

EricGustin:
# PR Description
For the `Google.GetThread` tool, we had a parameter named `metadata_headers`. This parameter only makes a difference if the format is "metadata", but the tool will never have the format "metadata". So, the input parameter is useless. This parameter should have never been added to the tool and we should remove it before public beta.


nbarbettini (approved):
Version bump needed?

PR #213, opened 2025-01-21
Update langchain integration to 0.2.0

Spartee:
**PR Description**

This update bumps the integration’s version to `0.2.0` and brings several important changes to how `langchain-arcade` interfaces with Arcade tools:

1. **Updated Tool Definition Imports**  
   • Replaces `arcadepy.types.shared.ToolDefinition` with `arcadepy.types.ToolGetResponse as ToolDefinition`.  
   • The parameter extraction is now done via `tool_def.input.parameters` instead of the previous `tool_def.inputs.parameters`.

2. **Authorization Flow Adjustments**  
   • Uses `auth_response.url` instead of `auth_response.authorization_url`.  
   • The `authorize` and `is_authorized` methods now rely on the Arcade client’s updated arguments (`client.auth.status(id=authorization_id)`).

3. **Tool Execution Parameter Renaming**  
   • The `execute` method now expects `input=kwargs` instead of `inputs=kwargs`, aligning with Arcade’s new API spec.

4. **Tool Retrieval Enhancements**  
   • `_retrieve_tool_definitions` is revised to better handle pagination and tool listing (including when no tools/toolkits are explicitly provided).

5. **Version & Dependency Updates**  
   • Increases `langchain-arcade` to `0.2.0`.  
   • Switches `arcadepy` dependency to `~1.0.0rc1`.  
   • Updates example requirements to consume `langchain-arcade[langgraph]>=0.2.0`.

These changes may affect existing code that relies on older parameter names (`inputs.parameters` → `input.parameters`) and the renamed execute argument. Please ensure any integrations or custom usage of Arcade tools is updated accordingly.

### DIFF
--- a/contrib/langchain/langchain_arcade/_utilities.py
+++ b/contrib/langchain/langchain_arcade/_utilities.py
@@ -1,7 +1,7 @@
 from typing import Any, Callable
 
 from arcadepy import NOT_GIVEN, Arcade
-from arcadepy.types.shared import ToolDefinition
+from arcadepy.types import ToolGetResponse as ToolDefinition
 from langchain_core.runnables import RunnableConfig
 from langchain_core.tools import StructuredTool
 from pydantic import BaseModel, Field, create_model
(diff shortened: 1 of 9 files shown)


nbarbettini:
Can you add a `wait_for_authorization` method or otherwise expose `client.auth.wait_for_completion`? That method includes a highly efficient long polling implementation.

That would make the loop/sleep in this example unnecessary: https://github.com/ArcadeAI/arcade-ai/blob/0e4cd81ab889d3871f92ed45e09ba040795b6b24/examples/langchain/custom_graph_with_auth.py#L60-L61

nbarbettini (approved):
LGTM

PR #215, opened 2025-01-21
Add Ability to Disable Tools 

EricGustin:
# PR Description
This PR introduces a new environment variable `ARCADE_DISABLED_TOOLS`. Tools that are added to this env var are not added to the worker's `ToolCatalog`. In effect, they are disabled for the worker.

## How to use the `ARCADE_DISABLED_TOOLS` environment variable
* Each tool is separated by a comma.
* For each tool, specify the toolkit name in camel case and the tool name in camel case.
* Do not include versions. (This is a simple implementation. We can add disabling specific versions in the future if needed)
* Separate the toolkit name and the tool name with your environment's tool name separator. By default, the tool name separator is `.`, but you can override this with the `ARCADE_TOOL_NAME_SEPARATOR` environment variable.

Correct: `export ARCADE_DISABLED_TOOLS="Math.Add,Spotify.GetAvailableDevices,Math.Sqrt"`
Incorrect: `export ARCADE_DISABLED_TOOLS="Math.Add@0.1.0,Spotify.get_available_devices,Sqrt`

### DIFF
--- a/arcade/arcade/cli/serve.py
+++ b/arcade/arcade/cli/serve.py
@@ -100,11 +100,6 @@ def serve_default_worker(
     if not toolkits:
         raise RuntimeError("No toolkits found in Python environment.")
 
-    logger.info("Serving the following toolkits:")
-    for toolkit in toolkits:
-        num_tools = sum(len(tools) for tools in toolkit.tools.values())
-        logger.info(f"  - {toolkit.name} ({toolkit.package_name}): {num_tools} tools")
-
     worker_secret = os.environ.get("ARCADE_WORKER_SECRET")
     if not disable_auth and not worker_secret:
         logger.warning(
@@ -124,8 +119,19 @@ def serve_default_worker(
     worker = FastAPIWorker(
         app, secret=worker_secret, disable_auth=disable_auth, otel_meter=otel_handler.get_meter()
     )
+
+    toolkit_tool_counts = {}
     for toolkit in toolkits:
+        prev_tool_count = worker.catalog.get_tool_count()
         worker.register_toolkit(toolkit)
+        new_tool_count = worker.catalog.get_tool_count()
+        toolkit_tool_counts[f"{toolkit.name} ({toolkit.package_name})"] = (
+            new_tool_count - prev_tool_count
+        )
+
+    logger.info("Serving the following toolkits:")
+    for name, tool_count in toolkit_tool_counts.items():
+        logger.info(f"  - {name}: {tool_count} tools")
 
     logger.info("Starting FastAPI server...")
 
--- a/arcade/arcade/core/catalog.py
+++ b/arcade/arcade/core/catalog.py
@@ -116,11 +116,11 @@ class ToolCatalog(BaseModel):
 
     _disabled_tools: set[str] = set()
 
-    def __init__(self, **data):
+    def __init__(self, **data) -> None:
         super().__init__(**data)
         self._load_disabled_tools()
 
-    def _load_disabled_tools(self):
+    def _load_disabled_tools(self) -> None:
         """Load disabled tools from the environment variable.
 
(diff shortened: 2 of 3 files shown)


arcade/tests/core/test_catalog.py
```diff
@@ -136,3 +136,33 @@ def test_get_tool_by_name_with_invalid_version():
 
     with pytest.raises(ValueError):
         catalog.get_tool_by_name("SampleToolkit.SampleTool", version="2.0.0")
+
```
nbarbettini, line 139: Does it make sense to add a test to make sure this change does not introduce a regression? in other words, if `ARCADE_DISABLED_TOOLS` is _not_ set, things work normally

EricGustin: I was thinking that the existing unit tests that add tools to catalog would verify there's no regression as long as they pass.

Or do you have a more involved regression test in mind that could give us more confidence?

nbarbettini: No I think you're right. The existing tests cover the regression angle.

Not sure if Python considers "unset" and "empty string" differently, in which case it would be useful to have a test that ensured `ARCADE_DISABLED_TOOLS=""` did not cause a crash.

arcade/arcade/core/catalog.py
```diff
@@ -146,6 +176,9 @@ def add_tool(
         if fully_qualified_name in self._tools:
             raise KeyError(f"Tool '{definition.name}' already exists in the catalog.")
 
+        if str(fully_qualified_name).lower() in self._disabled_tools:
+            return
```
sdreyer, line 180: Should we add a log here that says something like "Tool Toolkit.Name is disabled" to avoid confusion?

EricGustin: Added a log, thanks

```
INFO  [15:29:38.332] Tool 'Math.Add' is disabled and will not be cataloged.
INFO  [15:29:38.332] Tool 'Math.Subtract' is disabled and will not be cataloged.
INFO  [15:29:38.333] Tool 'Math.Divide' is disabled and will not be cataloged.
INFO  [15:29:38.334] Tool 'Math.Sqrt' is disabled and will not be cataloged.
INFO  [15:29:38.535] Tool 'Spotify.GetAvailableDevices' is disabled and will not be cataloged.
INFO  [15:29:38.536] Serving the following toolkits:
INFO  [15:29:38.536]   - slack (arcade_slack): 2 tools
INFO  [15:29:38.536]   - math (arcade_math): 5 tools
INFO  [15:29:38.536]   - spotify (arcade_spotify): 12 tools
```

arcade/tests/core/test_catalog.py
EricGustin: I like it. Added a couple tests
1. where `ARCADE_DISABLED_TOOLS=""`
2. where `ARCADE_DISABLED_TOOLS="      "`

arcade/arcade/cli/serve.py
```diff
@@ -124,8 +119,19 @@ def serve_default_worker(
     worker = FastAPIWorker(
         app, secret=worker_secret, disable_auth=disable_auth, otel_meter=otel_handler.get_meter()
     )
+
+    toolkit_tool_counts = {}
     for toolkit in toolkits:
+        prev_tool_count = worker.catalog.get_tool_count()
         worker.register_toolkit(toolkit)
+        new_tool_count = worker.catalog.get_tool_count()
+        toolkit_tool_counts[f"{toolkit.name} ({toolkit.package_name})"] = (
```
EricGustin, line 128: keeps track of the number of tools added to the catalog (avoids counting disabled tools)

sdreyer (approved):

nbarbettini:
Thank you Eric et al!

PR #220, opened 2025-01-23
Update README and LICENSE

EricGustin:
Updates README to point to updated URLs

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2024, Arcade AI
+Copyright (c) 2025, Arcade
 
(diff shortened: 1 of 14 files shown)


nbarbettini:
@EricGustin Can you also rename all `Arcade AI` to `Arcade`? (excluding the github org `ArcadeAI` which we can't change)

EricGustin:
> @EricGustin Can you also rename all `Arcade AI` to `Arcade`? (excluding the github org `ArcadeAI` which we can't change)

updated (excluding repo/org naming and container naming)

nbarbettini (approved):

PR #229, opened 2025-01-25
Fix dashboard redirect in docker compose

nbarbettini:
When the Engine is running via `docker compose`, the login redirect for the Dashboard was ending up at `0.0.0.0` instead of `localhost`.

### DIFF
--- a/docker/docker.engine.yaml
+++ b/docker/docker.engine.yaml
@@ -171,3 +171,6 @@ auth:
       provider_id: zoom
       client_id: ${env:ZOOM_CLIENT_ID}
       client_secret: ${env:ZOOM_CLIENT_SECRET}
+
+dashboard:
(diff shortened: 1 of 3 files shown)


docker/docker.engine.yaml
```diff
@@ -171,3 +171,6 @@ auth:
       provider_id: zoom
       client_id: ${env:ZOOM_CLIENT_ID}
       client_secret: ${env:ZOOM_CLIENT_SECRET}
+
+dashboard:
+  redirect_uri: "http://localhost:9099/dashboard"
```
sdreyer, line 176: Is there anything underlying that needs to be changed?

nbarbettini: Nope, this already works!

sdreyer (approved):

PR #232, opened 2025-01-27
Fix bug in slack tool pagination

byrro:
The `get_conversation_metadata_by_name` tool retrieves conversation metadata from another tool, `list_conversations_metadata`, but was accessing the `next_cursor` using the Slack API response dict structure, instead of the tool response structure. As a result, in that tool, the tool would never actually paginate to the second page. This PR fixes it and also adjust tests to capture the issue appropriately.


EricGustin (approved):
The changes look good to me

PR #241, opened 2025-02-05
Tool to retrieve Slack messages from a DM conversation with a given username

byrro:
Currently, retrieving DMs with a given username requires several actions: first get the current user's ID; list all users and find the ID of the username; then scan all DM conversations and find the one with the current user's ID and the username's ID, to finally retrieve the messages using that conversation ID.

This tool abstracts all that in a single call.

PS: we'll implement a similar tool for multi-person DM conversations in a subsequent PR.


toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -433,9 +442,89 @@ async def get_messages_in_channel_by_name(
...
+    Leave all arguments with the default None to get messages without date/time filtering"""
+    conversation_metadata = await get_direct_message_conversation_metadata_by_username(
+        context=context, username=username
     )
+
+    if not conversation_metadata:
+        raise ToolExecutionError(
+            "Conversation not found",
+            developer_message=f"Direct Message conversation with username '{username}' not found.",
+            retry_after_ms=500,
```
EricGustin, line 525: ToolExecutionError doesn't have `retry_after_ms` argument

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -490,67 +579,132 @@ async def get_conversation_metadata_by_id(
...
 
 
-@tool(
-    requires_auth=Slack(
-        scopes=["channels:read", "groups:read", "im:read", "mpim:read"],
-    )
-)
-async def get_conversation_metadata_by_name(
+@tool(requires_auth=Slack(scopes=["channels:read"]))
+async def get_channel_metadata_by_name(
```
EricGustin, line 582: Renaming from `get_conversation_metadata_by_name` to `get_channel_metadata_by_name` is in effect removing a tool (breaking change) without any deprecation warning. 

It would probably be better if we deprecated the tool with a warning & provide a transition period.

byrro: true, we can actually keep both for some time and have the old tool call the new one.. it's just that the name was not correct, since it couldn't really retrieve metadata from any conversation, only from channels, so the name change was meant to better express that

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -433,9 +444,88 @@ async def get_messages_in_channel_by_name(
     'latest_relative'.
 
     Leave all arguments with the default None to get messages without date/time filtering"""
-    conversation_metadata = await get_conversation_metadata_by_name(
-        context=context, conversation_name=channel_name
+    conversation_metadata = await get_channel_metadata_by_name(
```
EricGustin, line 447: nit: rename variable too?

toolkits/slack/arcade_slack/tools/chat.py
byrro: Just to document: we ended up concluding that the warnings need to be raised in the runtime where the tool execution is requested, not where it actually runs. Which means, we need a way for tools to pipe warning messages to our SDKs through the engine & worker, so that the SDKs can raise the warnings to the developers using them.

We'll move forward with this PR without warnings, but are starting to work on an implementation for this warning infrastructure.

byrro:
@EricGustin just pushed the changes we discussed. mypy seems to be happy. will find some time to work on the warnings infra hackweekend

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -433,12 +446,84 @@ async def get_messages_in_channel_by_name(
...
+        latest_relative=latest_relative,
+        oldest_datetime=oldest_datetime,
+        latest_datetime=latest_datetime,
+        limit=limit,
+        next_cursor=next_cursor,
     )
+
+
+@tool(requires_auth=Slack(scopes=["im:history", "im:read"]))
+async def get_messages_in_direct_conversation_by_username(
```
EricGustin, line 464: This is going to be extremely pedantic, but......

We have `get_direct_message_conversation_metadata_by_username` and then this new tool named `get_messages_in_direct_conversation_by_username`. Consistency tells me to have this tool named `get_messages_in_direct_message_conversation_by_username`.

I.e., one of the tools is referring to DMs as "direct_message_conversation" and the other is referring to DMs as "direct_conversation".

I'm assuming the LLM would prefer consistency over brevity?

toolkits/slack/arcade_slack/utils.py
```diff
@@ -174,6 +191,110 @@ def extract_basic_user_info(user_info: SlackUser) -> BasicUserInfo:
     )
 
 
+async def retrieve_conversations_by_user_ids(
```
EricGustin, line 194: this one is a doozy. Are there any opportunities to flatten this?  👉👈 

Even if it's just defining a _associate_members function outside of this

EricGustin (approved):

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -218,20 +227,24 @@ async def get_members_in_conversation_by_id(
...
-) -> Annotated[dict, "The conversation members' IDs and Names"]:
+) -> Annotated[dict, "The channel members' IDs and Names"]:
     """Get the members of a conversation in Slack by the conversation's name."""
-    conversation_metadata = await get_conversation_metadata_by_name(
-        context=context, conversation_name=conversation_name, next_cursor=next_cursor
-    )
+    channel = await get_channel_metadata_by_name(context=context, channel_name=channel_name)
+
+    if not channel:
+        raise ToolExecutionError(
```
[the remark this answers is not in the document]
byrro: actually need to remove this if block, the `get_channel_metadata_by_name` actually raises a retryable error with a list of potential channel names, it never returns None

toolkits/slack/arcade_slack/tools/chat.py
byrro: good point, agree we should lean towards clarity, will adjust it 😉

toolkits/slack/arcade_slack/utils.py
byrro: I'll think more tomorrow. Already went through several iterations to get here 😄

Like @Spartee says, gotta stop painting at some point...

byrro: @EricGustin just pushed a new implementation. it looks less doozy to me now 😄

let me know your thoughts on it

toolkits/slack/arcade_slack/tools/chat.py
byrro: @EricGustin if you ever have a comment that feels pedantic, please post it! 😄 this was very useful for me, because I saw that inconsistency after I was done and still neglected the small effort to adjust. It reinforces which side of my mind I should listen to.

PR #242, opened 2025-02-06
Decorator to warn the deprecation of an Arcade tool

byrro:
Implements a `deprecated` decorator to simplify raising a warning when a tool is set for deprecation.


byrro:
closing this PR for now. simply raising warnings inside the tools will not do the job. the developers who need to see the warnings won't actually get them. we need a more elaborate infrastructure to pipe warnings from tools to our customer's runtime where the Arcade SDKs are running.

PR #244, opened 2025-02-08
Improve Arcade CLI

EricGustin:
## PR Description
This PR improves upon the CLI in many ways.

1. Fixes bug where CLI commands that require the user to be logged in do not check for whether the user is logged in. This resulted in cryptic errors.
1. `arcade login` no longer generates an engine env file. When users download the Engine, the distribution already provides an `engine.env` file, so generating one with `arcade login` was redundant and causing confusion around which environment file to use.
1. `arcade login` used to look for `~/.arcade/credentials.yaml`, but this would cause issues if the `ARCADE_WORK_DIR` was set. Now it resolves to `ARCADE_WORK_DIR/.arcade/credentials.yaml` if the env var is set, else `~/.arcade/credentials.yaml`.
1. Refactored & improved error/warning messages in `_get_config_file`  by displaying all config file path locations that were searched. Previously, it only displayed the cwd and `~/.arcade/<config_file>`. `_get_config_file` remains functionially the same.


arcade/arcade/cli/authn.py
```diff
@@ -56,24 +60,21 @@ def _handle_login_response(self) -> bool:
             )
             return False
 
-        # ensure the ~/.arcade directory exists
-        # TODO: this should use WORK_DIR from env if set
-        arcade_dir = os.path.join(os.path.expanduser("~"), ".arcade")
-        if not os.path.exists(arcade_dir):
-            os.makedirs(arcade_dir, exist_ok=True)
+        # ensure the ARCADE_CONFIG_PATH directory exists
+        if not os.path.exists(ARCADE_CONFIG_PATH):
```
EricGustin, line 64: `ARCADE_CONFIG_PATH` uses the ARCADE_WORK_DIR env var if set.

arcade/arcade/cli/launcher.py
```diff
@@ -54,7 +55,7 @@ def start_servers(
     engine_config = _get_config_file(engine_config, default_filename="engine.yaml")
 
     # Ensure engine_env is provided or found and either way, validated
-    env_file = _get_config_file(engine_env, default_filename="arcade.env", optional=True)
+    env_file = _get_config_file(engine_env, default_filename="engine.env", optional=True)
```
EricGustin, line 58: Engine distribution uese `engine.env`, not `arcade.env`

arcade/arcade/cli/launcher.py
```diff
@@ -127,52 +134,47 @@ def _get_config_file(
...
+            console.print(f"❌ Config file not found at {candidate}", style="bold red")
+            raise RuntimeError(f"Config file not found at {candidate}")
+        return str(candidate)
+
+    # List of all config file path locations to check.
+    candidates = [
+        Path(os.getcwd()) / default_filename,
+        Path.home() / ".arcade" / default_filename,
+    ]
+    candidates.extend(Path(path) / default_filename for path in known_engine_config_locations)
```
sdreyer, line 148: Do we have a similar pattern for the env file? Its installed in the same dir as the engine config

EricGustin: yes, but since it is technically optional it will display as a warning if it can't be found.

Output:
image

arcade/arcade/cli/constants.py
```diff
@@ -1,7 +1,15 @@
-DEFAULT_CLOUD_HOST = "cloud.arcade.dev"
-DEFAULT_ENGINE_HOST = "api.arcade.dev"
+import os
+
+PROD_CLOUD_HOST = "cloud.arcade.dev"
+PROD_ENGINE_HOST = "api.arcade.dev"
 LOCALHOST = "localhost"
 
+# The path to the directory containing the Arcade configuration files. Typically ~/.arcade
+ARCADE_CONFIG_PATH = os.path.join(os.path.expanduser(os.getenv("ARCADE_WORK_DIR", "~")), ".arcade")
```
nbarbettini, line 8: Thanks for using Windows-safe methods 🙏

arcade/arcade/cli/main.py
```diff
@@ -70,7 +75,9 @@ def login(
     """
 
     if check_existing_login():
-        console.print("Delete ~/.arcade/credentials.yaml to log in as a different user.\n")
+        console.print("\nTo log out as the current user, use ", end="")
```
nbarbettini, line 78:
```suggestion
        console.print("\nTo log out and delete your locally-stored credentials, use ", end="")
```

nbarbettini (approved):
These improvements look great, thanks for fixing these papercuts @EricGustin!

PR #247, opened 2025-02-13
Add `@tool.deprecated`

EricGustin:
## PR Description
Add the ability to mark a tool as deprecated and display the warning in the user's runtime. This PR also lays the foundation for future work for emitting other levels of logs (debug, info, etc) that occur during the tool's execution.

NOTE: Updates to the Arcade Clients (Python and JS) still need to be done before the deprecation warning is emitted, but this PR needs to be merged before those updates!

Let's cross our fingers that we'll never need to deprecate `@tool.deprecated`!

### Example

1. Mark your tool as deprecated
```python
from typing import Annotated

from arcade.sdk import tool


@tool.deprecated("Use the 'Math.AddInt' tool instead.") # order of decorators does not matter
@tool
def add(
    a: Annotated[int, "The first number"], b: Annotated[int, "The second number"]
) -> Annotated[int, "The sum of the two numbers"]:
"""
Add two numbers together
"""
return a + b
```

2. Call the deprecated tool
```python
from arcadepy import Arcade

client = Arcade()

tool_input = {"a": 9001, "b": 42}

response = client.tools.execute(
    tool_name="Math.Add",
    input=tool_input,
    user_id="me@example.com",
)
print(f"The result of adding {tool_input['a']} and {tool_input['b']} is: {response.output.value}")
```

3. Observe the DeprecationWarning:
``` 
❯ python examples/call_a_tool_directly.py 
/Users/ericgustin/repos/Team/arcade-ai/examples/call_a_tool_directly.py:22: DeprecationWarning: 'Math.Add' is deprecated: Use the `Math.AddInt` tool instead.
  response = client.tools.execute(
The result of adding 9001 and 42 is: 9043
```

### DIFF
--- a/arcade/arcade/core/executor.py
+++ b/arcade/arcade/core/executor.py
@@ -12,7 +12,7 @@ from arcade.core.errors import (
     ToolSerializationError,
 )
 from arcade.core.output import output_factory
-from arcade.core.schema import ToolCallOutput, ToolCallWarning, ToolContext, ToolDefinition
+from arcade.core.schema import ToolCallLog, ToolCallOutput, ToolContext, ToolDefinition
 
 
 class ToolExecutor:
@@ -29,6 +29,13 @@ class ToolExecutor:
         """
         Execute a callable function with validated inputs and outputs via Pydantic models.
         """
+        # only gathering deprecation log for now
+        tool_call_logs = []
+        if definition.deprecation_message:
+            tool_call_logs.append(
+                ToolCallLog(message=definition.deprecation_message, level="deprecation")
+            )
+
         try:
             # serialize the input model
             inputs = await ToolExecutor._serialize_input(input_model, **kwargs)
@@ -49,17 +56,8 @@ class ToolExecutor:
             # serialize the output model
             output = await ToolExecutor._serialize_output(output_model, results)
 
-            # gather any warnings
-            warnings = []
-            if definition.deprecation_message:
-                warnings.append(
-                    ToolCallWarning(
-                        message=definition.deprecation_message, warning_type="deprecation"
-                    )
-                )
-
             # return the output
-            return output_factory.success(data=output, warnings=warnings)
+            return output_factory.success(data=output, logs=tool_call_logs)
 
         except RetryableToolError as e:
             return output_factory.fail_retry(
--- a/arcade/arcade/core/executor.py
+++ b/arcade/arcade/core/executor.py
@@ -12,7 +12,7 @@ from arcade.core.errors import (
     ToolSerializationError,
 )
 from arcade.core.output import output_factory
-from arcade.core.schema import ToolCallOutput, ToolContext, ToolDefinition
+from arcade.core.schema import ToolCallOutput, ToolCallWarning, ToolContext, ToolDefinition
 
 
 class ToolExecutor:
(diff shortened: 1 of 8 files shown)


arcade/arcade/core/executor.py
```diff
@@ -49,8 +49,17 @@ async def run(
...
+            if definition.deprecation_message:
+                warnings.append(
+                    ToolCallWarning(
+                        message=definition.deprecation_message, warning_type="deprecation"
+                    )
+                )
+
             # return the output
-            return output_factory.success(data=output)
+            return output_factory.success(data=output, warnings=warnings)
```
EricGustin, line 62: only pass on the warnings if tool execution was successful. Thoughts?

nbarbettini: This makes sense strictly for this deprecation use case.

But thinking ahead: we will eventually want capture any number of logs (debug, info, warning, up to error) that could help someone understand why a particular tool failed. For example, a complex tool could emit logs as it steps through chunks of business logic, which would help someone pinpoint why it failed.
I'm OK if this isn't addressed in this PR, as long as it's filed for the future. 👍

arcade/arcade/core/output.py
```diff
@@ -14,9 +14,10 @@ def success(
         self,
         *,
         data: T | None = None,
+        warnings: list[ToolCallWarning] | None = None,
     ) -> ToolCallOutput:
         value = getattr(data, "result", "") if data else ""
-        return ToolCallOutput(value=value)
+        return ToolCallOutput(value=value, warnings=warnings)
```
nbarbettini, line 20: What about `logs` of type `ToolCallLog` which contain a `.level`?

EricGustin: Ok my initial thought was that we would have `ToolCallOutput(value=value, debugs, logs, warnings)` but your idea makes much more sense. Especially because it would maintain order between types of logs.

arcade/arcade/core/executor.py
EricGustin: addressed

arcade/arcade/core/schema.py
```diff
@@ -262,14 +262,20 @@ class ToolCallRequest(BaseModel):
...
     """The user-facing warning message."""
 
-    warning_type: Literal["deprecation"]
-    """The type of warning that occurred."""
+    level: Literal[
+        "debug",
+        "info",
+        "warning",
+        "error",
+        "deprecation",
```
nbarbettini, line 276: Why should `deprecation` be its own level, rather than being a `warning`?

EricGustin: I chose this way because deprecation felt like it is a special type of warning given that it is a warning about the tool itself, and not a warning that occurred during the execution of the tool.

I see your concern since typically the levels are debug, info, warning, error.

If we don't think it warrants its own level, then changing this would be very easy.

nbarbettini: Ah, I see what you mean by it being a special type. It could be useful to handle that type differently downstream.

To me, that feels like something that could go into a `type` field, rather than overloading `level` (since loglevels are a pretty well-defined thing in most code)

EricGustin: yep, an optional type field was the alternative. I'll see what I can come up with

EricGustin: I added `subtype` field to `ToolCallLog`.

nbarbettini (approved):
LGTM

PR #249, opened 2025-02-13
Google toolkit: Search and create contacts POC

nbarbettini:
This is an initial sketch of what Contacts (People) API tools could look like. But I haven't yet thought like an MX Engineer @byrro 😉

### DIFF
--- a/toolkits/google/arcade_google/tools/contacts.py
+++ b/toolkits/google/arcade_google/tools/contacts.py
@@ -51,7 +51,7 @@ async def search_contacts(
     return {"contacts": primary_results}
 
 
-async def _warmup_cache(service):
+async def _warmup_cache(service) -> None:
     """
     Warm-up the search cache for contacts by sending a request with an empty query.
     This ensures that the lazy cache is updated for both primary contacts and other contacts.
--- a/toolkits/google/arcade_google/tools/contacts.py
+++ b/toolkits/google/arcade_google/tools/contacts.py
@@ -51,7 +51,7 @@ async def search_contacts(
     return {"contacts": primary_results}
 
 
-async def _warmup_cache(service) -> None:
+async def _warmup_cache(service) -> None:  # type: ignore[no-untyped-def]
     """
     Warm-up the search cache for contacts by sending a request with an empty query.
     This ensures that the lazy cache is updated for both primary contacts and other contacts.
--- a/toolkits/google/arcade_google/tools/contacts.py
+++ b/toolkits/google/arcade_google/tools/contacts.py
@@ -1,0 +1,90 @@
+import asyncio
+from typing import Annotated
+
+from arcade.sdk import ToolContext, tool
+from arcade.sdk.auth import Google
+from google.oauth2.credentials import Credentials
+from googleapiclient.discovery import build
+
+
+@tool(requires_auth=Google(scopes=["https://www.googleapis.com/auth/contacts.readonly"]))
+async def search_contacts(
+    context: ToolContext,
+    query: Annotated[
+        str,
+        "The search query for filtering contacts.",
+    ],
+    page_size: Annotated[int, "The maximum number of contacts to return"] = 10,
+) -> Annotated[dict, "A dictionary containing the list of matching contacts"]:
+    """
+    Search the user's contacts using the People API.
+
+    This tool queries the contacts with the provided query string.
+    The API returns contacts that match based on names, email addresses, and more.
+    """
+    # Build the People API service
+    service = build(
+        "people",
+        "v1",
+        credentials=Credentials(
+            context.authorization.token
+            if context.authorization and context.authorization.token
+            else ""
+        ),
+    )
+
+    # Warm-up the cache before performing search.
+    # TODO: Ideally we should warmup only if this user (or google domain?) hasn't warmed up recently
+    _warmup_cache(service)
+
+    # Search primary contacts using searchContacts
+    primary_response = (
+        service.people()
+        .searchContacts(query=query, pageSize=page_size, readMask="names,emailAddresses")
+        .execute()
+    )
+    primary_results = primary_response.get("results", [])
+
+    return {"contacts": primary_results}
+
+
+async def _warmup_cache(service):
+    """
+    Warm-up the search cache for contacts by sending a request with an empty query.
+    This ensures that the lazy cache is updated for both primary contacts and other contacts.
+    This is a real thing: https://developers.google.com/people/v1/contacts#search_the_users_contacts
+    """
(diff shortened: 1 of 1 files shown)


toolkits/google/arcade_google/tools/contacts.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@tool(requires_auth=Google(scopes=["https://www.googleapis.com/auth/contacts.readonly"]))
+async def search_contacts(
+    context: ToolContext,
+    query: Annotated[
+        str,
+        "The search query for filtering contacts.",
+    ],
+    page_size: Annotated[int, "The maximum number of contacts to return"] = 10,
```
byrro, line 17: I see Google caps this at 30. It's a good idea to mention that in the arg annotation. Maybe also highlight that the list of contacts returned might not be exhaustive. Too bad Google won't let you paginate search results... 🤷‍♂️

toolkits/google/arcade_google/tools/contacts.py
```diff
@@ -0,0 +1,90 @@
...
+        credentials=Credentials(
+            context.authorization.token
+            if context.authorization and context.authorization.token
+            else ""
+        ),
+    )
+
+    # Warm-up the cache before performing search.
+    # TODO: Ideally we should warmup only if this user (or google domain?) hasn't warmed up recently
+    _warmup_cache(service)
```
byrro, line 38: Perhaps the warm-up could be implemented as a decorator, which could be aware of the last warm-up time and skip it, in case the index is already warm? not sure if this could be implemented easily considering our cloud engine has to support multi-tenancy... in a single tenant implementation, I imagine even an env var `GOOGLE_CONTACTS_INDEX_LAST_WARMUP_TIMESTAMP` might suffice

toolkits/google/arcade_google/tools/contacts.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    service.people().searchContacts(query="", pageSize=1, readMask="names,emailAddresses").execute()
+    await asyncio.sleep(3)  # TODO experiment with this value
+
+
+@tool(requires_auth=Google(scopes=["https://www.googleapis.com/auth/contacts"]))
+async def create_contact(
+    context: ToolContext,
+    given_name: Annotated[str, "The given name of the contact"],
+    family_name: Annotated[str, "The family name of the contact"],
```
byrro, line 65: `family_name` could be optional?

toolkits/google/arcade_google/tools/contacts.py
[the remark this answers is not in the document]
nbarbettini: Absolutely 💯

nbarbettini:
Marking this as ready for review because:
The Google verification team needs to see and test these scopes live on our staging app in order to approve them. Merging these contacts tools is the fastest way to demonstrate `contacts.readonly` and `contacts`.
I've already [updated](https://github.com/ArcadeAI/Team/pull/147/files#diff-4b0811905dd5319b97feebf01b5b959c3aba6f9278b962aa825e63e2f9d17e31) our prod config so that these tools will not appear on prod.

Merging this does mean that we need to clean up/finish the tools before our next official Google toolkit release.
cc @Spartee

toolkits/google/arcade_google/tools/contacts.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+async def _warmup_cache(service) -> None:  # type: ignore[no-untyped-def]
+    """
+    Warm-up the search cache for contacts by sending a request with an empty query.
+    This ensures that the lazy cache is updated for both primary contacts and other contacts.
+    This is a real thing: https://developers.google.com/people/v1/contacts#search_the_users_contacts
+    """
+    service.people().searchContacts(query="", pageSize=1, readMask="names,emailAddresses").execute()
+    await asyncio.sleep(3)  # TODO experiment with this value
```
Spartee, line 61: what does this function accomplish?

toolkits/google/arcade_google/tools/contacts.py
```diff
@@ -0,0 +1,99 @@
...
+        service.people()
+        .searchContacts(query=query, pageSize=limit, readMask="names,emailAddresses")
+        .execute()
+    )
+    primary_results = primary_response.get("results", [])
+
+    return {"contacts": primary_results}
+
+
+async def _warmup_cache(service) -> None:  # type: ignore[no-untyped-def]
```
Spartee, line 54: type this?

toolkits/google/arcade_google/tools/contacts.py
```diff
@@ -0,0 +1,99 @@
...
+
+@tool(requires_auth=Google(scopes=["https://www.googleapis.com/auth/contacts"]))
+async def create_contact(
+    context: ToolContext,
+    given_name: Annotated[str, "The given name of the contact"],
+    family_name: Annotated[Optional[str], "The family name of the contact"],
+    email: Annotated[Optional[str], "The email address of the contact"],
+) -> Annotated[dict, "A dictionary containing the details of the created contact"]:
+    """
+    Create a new contact in the user's Google Contacts using the People API.
```
Spartee, line 72:
```python
"""Create a new contact record in Google Contacts.

This uses the Google People API to create and populate a new record with contact information
"""
```
Something like this might perform better. it's good to be concise so as to remove the possibility of a tool selection mismatch for a user query like "create a new user for the following people" which means to go hit posthog or something, but semantically seems to match well to the description. Eval this and try it out. be adversarial

toolkits/google/arcade_google/tools/contacts.py
nbarbettini: Fixes the Google API being dumb. It is explained in the linked comment: https://developers.google.com/people/v1/contacts#search_the_users_contacts

nbarbettini:
@Spartee Note that I am intentionally excluding tests and evals in this PR because these tools are unfinished. I just need to get them to staging so that we can prove to Google that we are using the `contacts` and `contacts.readonly` scopes.

If we do not want the `contacts/contacts.readonly` scopes right now, we can move this PR back to draft and not merge it right now.

toolkits/google/arcade_google/tools/contacts.py
nbarbettini: Open to suggestions here; the Google SDK is not typed which made it difficult to type the `service` arg here.

Spartee (approved):

nbarbettini:
Merging work in progress. Filed [Finish Google Contacts toolkit](https://app.clickup.com/t/86b3znk2t) to finish it up.

PR #252, opened 2025-02-14
feat: Tool secrets

nbarbettini:
SDK support for tool secrets (stored and managed by the engine):
- [x] New `requires_secrets=` option in the `@tool` decorator
- [x] Internal plumbing in the catalog and `ToolContext`
- [x] Full test coverage of all added code
- [x] Bumped minor version (new feature)

This PR can be merged without waiting for Engine changes, because it is additive only (no breaking changes).

After this is merged, I will open another PR to update existing toolkits that will benefit from this feature!


arcade/tests/tool/test_create_tool_definition_errors.py
```diff
@@ -41,6 +41,14 @@ def func_with_multiple_context_params(context: ToolContext, context2: ToolContex
...
 
 
+@tool(
+    desc="A function with a required secret with a missing key (illegal)",
+    requires_secrets=[""],
+)
+def func_with_missing_secret_key(context: ToolContext):
+    pass
+
+
```
EricGustin, line 51: Within `Catalog.create_tool_definition`, we should ensure `isinstance(secret.key_id, str)`

Suggesting adding a test for this:
```py
@tool(
    desc="A function that requires a secret",
    requires_secrets=[True],
)
def func_with_invalid_type_secret_requirement():
    pass
```

arcade/tests/tool/test_create_tool_definition_errors.py
```diff
@@ -79,6 +87,11 @@ def func_with_multiple_context_params(context: ToolContext, context2: ToolContex
             ToolDefinitionError,
             id=func_with_multiple_context_params.__name__,
         ),
+        pytest.param(
+            func_with_missing_secret_key,
+            ToolDefinitionError,
+            id=func_with_missing_secret_key.__name__,
+        ),
```
EricGustin, line 94: In addition to the above comment
```python
pytest.param(
    func_with_invalid_type_secret_requirement,
    ToolDefinitionError,
    id=func_with_invalid_type_secret_requirement.__name__,
),
  ```

File: arcade/tests/tool/test_create_tool_definition_errors.py (hunk 1 above)
Comment (nbarbettini) [reply arcade/tests/tool/test_create_tool_definition_errors.py:51]: Great catch, thanks. Added test

File: arcade/arcade/core/catalog.py [hunk 3]
```diff
@@ -374,6 +375,17 @@ def create_tool_definition(
...
+        secrets_requirement = getattr(tool, "__tool_requires_secrets__", None)
+        if isinstance(secrets_requirement, list):
+            if any(not isinstance(secret, str) for secret in secrets_requirement):
+                raise ToolDefinitionError("Secrets keys must be strings.")
+
+            secrets_requirement = to_tool_secret_requirements(secrets_requirement)
+            if any(
+                secret.key is None or secret.key.strip() == "" for secret in secrets_requirement
+            ):
+                raise ToolDefinitionError("Secrets must have a non-empty key.")
```
EricGustin, line 387: could be useful to provide the tool's name in the error for easier debugging

EricGustin (approved):
looks good! one minor comment

PR #254, opened 2025-02-18
Slack tools to retrieve messages & metadata from multi-person DM conversation

byrro:


toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -669,8 +758,66 @@ async def get_direct_message_conversation_metadata_by_username(
...
+            user_ids=[
+                current_user["user_id"],
+                *[user["id"] for user in other_users if user["id"] != current_user["user_id"]],
+            ],
+            exact_match=True,
+            limit=1,
+            next_cursor=next_cursor,
+        )
+
+        return None if not conversations_found else conversations_found[0]
```
EricGustin, line 815: Oddly enough, returning `None` makes the model think that it successfully retrieved the metadata, also technically not the return type expected. Models seem to like empty `dict` instead.

For example, when returning `None`:
```
MODEL: 
I've successfully retrieved the metadata for the conversation with <personA>, <personB>, 
<personC>, and <personD>. Let me know if you need any additional   
information about this conversation! 

DEBUG INFO:
Called tool 'Slack_GetMultiPersonDmConversationMetadataByUsernames' with parameters: 
{"usernames":["personA","personB","personC","personD"]}
'Slack_GetMultiPersonDmConversationMetadataByUsernames' tool returned: 
Slack.GetMultiPersonDmConversationMetadataByUsernames called successfully
```

And then when returning empty `dict` (implies no metadata is found):
```
MODEL: 
It seems there isn't any available metadata about a conversation including 
<personA>, <personB>, <personC>, and <personD>. Please check if the conversation 
exists or if the usernames are correct. 


DEBUG INFO: 
Called tool 'Slack_GetMultiPersonDmConversationMetadataByUsernames' with parameters: 
{"usernames":["personA","personB","personC","personD"]}
'Slack_GetMultiPersonDmConversationMetadataByUsernames' tool returned: {}
```

byrro: interesting.. the return type hint is `Optional[dict]`, so `None` is a possibility, but it's far from ideal in terms of consistency, indeed

I'm thinking now, raising a `RetryableToolError` would make more sense here? there are many other tools/toolkits where we raise it to signal a resource was not found.

byrro: moved the raising of `RetryableToolError` to the metadata tool when no conversation is found with the usernames provided. this prevents the LLM confusion and is more consistent with how other tools handle similar cases

EricGustin (approved):

PR #264, opened 2025-02-21
Update scope on Google.ListDocuments

nbarbettini:
Updating the scope on this unreleased tool to be more granular.


EricGustin (approved):
Good catch

PR #265, opened 2025-02-25
Search Google Drive documents and retrieve contents

byrro:
This tool will be useful in scenarios akin to RAG, where someone wants to ask questions or request the production of a summary, for instance, about a bunch of documents related to a particular topic. Currently, to fulfill such requests, the LLM needs to first `list_documents`, then `get_document_by_id` for each document.

We also implement a utility functions to return documents in Markdown and HTML, since the Drive API JSON is verbose and would waste too many tokens unnecessarily.

Limitations: the Markdown/HTML utilities do not handle table of contents (which I think aren't really useful here), headers, footers, or footnotes.

---
This PR deprecates `list_documents` and implements `search_documents`, apart from `search_and_retrieve_documents`). This configuration makes it easier for LLMs to understand when to call each tool.

Both tools had their interfaces refactored to remove Google API-specific arguments that were confusing LLMs sometimes, such as "corpora" and "support_all_drives". It now accepts arguments that better relate to expected user requests.


EricGustin:
@byrro I'm unable to run this locally. Looks like an issue with all of the double and single quotes in the `arcade_google/utils.py`

image

toolkits/google/arcade_google/tools/drive.py
```diff
@@ -20,18 +22,28 @@
 async def list_documents(
```
EricGustin, line 22: my suggestion:
1. deprecate `list_documents`
2. place your new implementation inside a new tool named `search_documents`
3. rename `search_and_retrieve_documents_in_markdown` to `search_and_retrieve_documents` and let the return format be an input parameter

toolkits/google/arcade_google/tools/drive.py
```diff
@@ -20,18 +22,28 @@
...
+        Optional[list[str]],
+        "Keywords or phrases that must NOT be in the document title or body. Provide a list of "
+        "phrases if needed",
     ] = None,
     order_by: Annotated[
-        OrderBy,
+        Optional[list[OrderBy]],
         "Sort order. Defaults to listing the most recently modified documents first",
-    ] = OrderBy.MODIFIED_TIME_DESC,
+    ] = None,
```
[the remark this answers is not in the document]
byrro: The worker won't let a mutable default

toolkits/google/arcade_google/tools/drive.py
byrro: That makes sense. I'll change just the interface and see if the LLM understands it better. Likely will

toolkits/google/arcade_google/tools/drive.py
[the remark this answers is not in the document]
EricGustin: brain fart. yeah definitely don't use a list as default!

toolkits/google/arcade_google/tools/drive.py
byrro: it does understand it better. pushing the changes now.

toolkits/google/arcade_google/utils.py
```diff
@@ -592,6 +592,82 @@ def build_drive_service(auth_token: Optional[str]) -> Resource:  # type: ignore[
     return build("drive", "v3", credentials=Credentials(auth_token))
 
 
+def build_files_list_query(
+    document_contains: Optional[list[str]] = None,
+    document_not_contains: Optional[list[str]] = None,
+) -> str:
+    query = ["mimeType = 'application/vnd.google-apps.document' and trashed = false"]
```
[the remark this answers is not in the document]
EricGustin: thank you!

toolkits/google/arcade_google/utils.py
```diff
@@ -592,6 +592,85 @@ def build_drive_service(auth_token: Optional[str]) -> Resource:  # type: ignore[
...
+
+    if isinstance(document_not_contains, str):
+        document_not_contains = [document_not_contains]
+
+    if document_contains:
+        for keyword in document_contains:
+            name_contains = keyword.replace("'", "\\'")
+            full_text_contains = keyword.replace("'", "\\'")
+            keyword_query = (
+                f"name contains '{name_contains}' or fullText contains '{full_text_contains}'"
```
EricGustin, line 613: we need to group the left and right side of the `or` inside parentheses, otherwise Google interprets the query in a way that we don't intend. See slack dm for more details

byrro: good catch! just pushed a fix

EricGustin (approved):
Looks great!

PR #267, opened 2025-02-25
Fix package install for modal example

sdreyer:


examples/serving-tools/modal/run-arcade-worker.py
```diff
@@ -26,8 +26,8 @@ def fastapi_app():
 
     # Register toolkits we've installed
     installed_toolkits = Toolkit.find_all_arcade_toolkits()
-    for toolkit in toolkits:
```
sdreyer, line 29: Was previously comparing a Toolkit to a str

examples/serving-tools/modal/run-arcade-worker.py
```diff
@@ -5,9 +5,9 @@
 # Define the FastAPI app
 app = App("arcade-worker")
 
-toolkits = ["arcade-google", "arcade-slack"]
+toolkits = ["arcade_google", "arcade_slack"]
```
sdreyer, line 8: Need to use underscores to compare with the package name loaded in

examples/serving-tools/modal/run-arcade-worker.py
```diff
@@ -5,9 +5,9 @@
...
 
-image = Image.debian_slim().pip_install("arcade-ai").pip_install(toolkits)
+image = Image.debian_slim().pip_install("arcade-ai[fastapi]").pip_install(toolkits)
```
EricGustin, line 10: shouldn't be needed for arcade-ai >=1.0.2. Were you facing issues with this?

Spartee:
Looks good to me

examples/serving-tools/modal/run-arcade-worker.py
Spartee: modal had a weird bug here at one point where they needed you to install FastAPI again after the initial spinup. not sure if this was related or not.

sdreyer: Will check it out

EricGustin (approved):

PR #269, opened 2025-02-26
 Tool to retrieve file tree structure from Google 'My Drive' and 'Shared Drives'

byrro:
#da-fastest-day-1


toolkits/google/arcade_google/tools/drive.py
```diff
@@ -82,3 +82,92 @@ async def list_documents(
...
 
     return {"documents_count": len(files), "documents": files}
+
+
+@tool(
+    requires_auth=Google(
+        scopes=["https://www.googleapis.com/auth/drive.file"],
+    )
+)
+async def get_file_tree_structure(
```
EricGustin, line 92: Thinking out loud, but let me know what you think...

--------

I think we can reduce the number of API calls being made in this and it might be worth it since the API call count could potentially be very high and getting rate limited would become a serious issue.

For example, if I have 400 files and 5 folders, then this implementation will call the Google API 400 + 5 + 1 = 406 times. (correct me if I'm wrong about that)

-------

If we instead start with the the query `"'root' in parents and trashed = false"`,
then we can then iterate on the returned list and **only** call the Google API if the file's MIME type is `application/vnd.google-apps.folder`. The query would then change to `"'curr_folder_id' in parents and trashed = false"`. Otherwise, if the MIME type is NOT `application/vnd.google-apps.folder`, then we can add it to the folder structure that we are actively constructing.

Using the previous example, this would reduce the number of calls to the Google API from 406 to 6.

-----
TLDR: 
Searching based on the `parent` (starting with root) and only calling the API for MIME type `application/vnd.google-apps.folder` would reduce API calls drastically.

toolkits/google/arcade_google/tools/drive.py
```diff
@@ -82,3 +82,92 @@ async def list_documents(
...
+    context: ToolContext,
+    include_shared_drives: Annotated[
+        bool, "Whether to include shared drives in the file tree structure. Defaults to True."
+    ] = True,
+) -> Annotated[
```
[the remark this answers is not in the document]
byrro: the files.list endpoint returns a flat list of files and folders, we need to recur only if there're more items than the page size, not in depth

toolkits/google/arcade_google/tools/drive.py
byrro: Turns out we can provide a `fields` to the `files.list` endpoint and ask it to return `parents` attribute with each item. It's not documented in the endpoint docs, but it works. With that, we only need a single call (for each page) to build the entire file tree. Apart from this request, the tool also queries `drives.get` to get Shared Drives' names, if any.

toolkits/google/arcade_google/tools/drive.py
```diff
@@ -82,3 +88,96 @@ async def list_documents(
...
+@tool(
+    requires_auth=Google(
+        scopes=["https://www.googleapis.com/auth/drive.file"],
+    )
+)
+async def get_file_tree_structure(
+    context: ToolContext,
+    include_shared_drives: Annotated[
+        bool, "Whether to include shared drives in the file tree structure. Defaults to True."
+    ] = True,
```
EricGustin, line 102: I'm wary of defaulting to including shared drives as someone who is a part of a large organization with many shared drives could have tens of thousands of documents in those shared drives.

Up to you though - I see both pros and cons

EricGustin (approved):
LGTM - one comment that you can choose to ignore or address

toolkits/google/arcade_google/tools/drive.py
byrro: let's be safe and make it default to `False` then

EricGustin (approved):

Spartee:
Really great work y'all. Keep it up

Spartee:
Great work team, keep it up!

Spartee:
Great work team, keep it up!

Spartee:
Great job. Looks good to me.

Spartee:
Looks good to me, let's meet about this soon.

PR #271, opened 2025-02-28
fix: Use tool secrets in toolkits

nbarbettini:
~~Note: Don't merge until the correct secrets have been added to Arcade Cloud.~~

Ready to merge, the feature is already on its way to prod.


toolkits/code_sandbox/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "arcade_code_sandbox"
-version = "0.1.9"
+version = "1.0.0"
```
nbarbettini, line 3: I'm considering this a breaking change, because the required configuration has changed in a backwards-incompatible way.

cc @EricGustin for thoughts.

nbarbettini:
Note: Because tests install the SDK from pypi, tests will fail until we publish #252 in a public release.

toolkits/code_sandbox/pyproject.toml
EricGustin: sgtm. We also need to make the minimum arcade-ai version stricter for these toolkits

nbarbettini: Doh! fixed

EricGustin (approved):

PR #273, opened 2025-03-04
Fix parameterized type bug & better error message

EricGustin:
### 1. Bug Fix
A bug was observed where tools that were using [parameterized generics](https://docs.python.org/3/library/stdtypes.html#types-genericalias) (e.g., `dict[str, Any]`) for their input parameters or output, then this would cause the Worker to fail on startup with the error `issubclass() arg 1 must be a class` for Python3.13+. This error would not occur for Python versions less than 3.13.

In Python <3.13, parameterized generics would implicitly be treated as their underlying type (its origin), so it was possible to treat `dict[str, Any]` as a class and pass it to `issubclass`. This is not the case for Python 3.13, so we need to explicitly strip the GenericAlias (e.g., `dict[str, Any]`) down to its origin (e.g., `dict`), before checking if it is a subclass of `Enum`.

### 2. Better Error Message
When a tool used an unsupported parameter/output type, then Arcade would display the following message:
```
Failed to start Arcade Worker: 
Type error encountered while adding tool get_website_map from arcade_web.tools.firecrawl. 
Reason: issubclass() arg 1 must be a class
```

But now it displays
```
Failed to start Arcade Worker: 
Error encountered while adding tool get_website_map from arcade_web.tools.firecrawl. 
Reason: Unsupported type: tuple[str, str]
```

### DIFF
--- a/arcade/arcade/core/catalog.py
+++ b/arcade/arcade/core/catalog.py
@@ -745,7 +745,7 @@ def get_wire_type(
     if isinstance(_type, type) and issubclass(_type, BaseModel):
         return "json"
 
-    raise ToolDefinitionError(f"Unsupported type: {_type}")
+    raise ToolDefinitionError(f"Unsupported parameter type: {_type}")
 
 
 def create_func_models(func: Callable) -> tuple[type[BaseModel], type[BaseModel]]:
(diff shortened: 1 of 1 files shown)


nbarbettini (approved):
Thanks for making the error message much better for humans. 👍

PR #277, opened 2025-03-06
Add Notion Auth Provider

EricGustin:
Screenshot 2025-03-06 at 11 01 24 AM


arcade/arcade/core/auth.py
```diff
@@ -60,46 +60,46 @@ def __init__(self, *, id: Optional[str] = None, scopes: Optional[list[str]] = No
         super().__init__(id=id, scopes=scopes)
 
 
-class Google(OAuth2):
```
EricGustin, line 63: Reorganized the classes so that they are in alphabetical order

wdawson (approved):

PR #279, opened 2025-03-07
Google Calendar tool to find time slots when one or more people (within the same org domain) are simultaneously free

byrro:


Spartee:
@EricGustin @byrro I worry about the timezone being too based on the location of the caller. In many cases it may come from the client or it may be in the prompt. this may or may not matter, but the reason Archer almost always gets the time right is because of the prompt work I did [here](https://github.com/ArcadeAI/SlackAgent/blob/main/archer/utils.py) and [here](https://github.com/ArcadeAI/SlackAgent/blob/main/archer/defaults.py)

Spartee:
also... what happens if they aren't in the same org? what's the failure mode? This seems incredibly specific as a tool and requires lots of previous information in the context. Have we considered the adversarial cases here?

byrro:
> @EricGustin @byrro I worry about the timezone being too based on the location of the caller. In many cases it may come from the client or it may be in the prompt. this may or may not matter, but the reason Archer almost always gets the time right is because of the prompt work I did [here](https://github.com/ArcadeAI/SlackAgent/blob/main/archer/utils.py) and [here](https://github.com/ArcadeAI/SlackAgent/blob/main/archer/defaults.py)

it uses the default timezone from the user's primary calendar. one version had an optional `timezone` argument, but it was frequently messing up the LLM understanding of how to call it... in the majority of cases, I think people will want to see times in their own timezone. do you think it's important to try to get the `timezone` argument back?

byrro:
> also... what happens if they aren't in the same org? what's the failure mode? This seems incredibly specific as a tool and requires lots of previous information in the context. Have we considered the adversarial cases here?

for sure, that was not being handled properly. I just added exception handling when the user asks for a calendar they don't have access to or any other error returned by the google api.

toolkits/google/arcade_google/tools/calendar.py
```diff
@@ -356,3 +361,140 @@ async def delete_event(
...
+    if user_info["email"] not in email_addresses:
+        email_addresses.append(user_info["email"])
+
+    # Get the timezone of the currently logged in user
+    calendar = calendar_service.calendars().get(calendarId="primary").execute()
+    timezone_name = calendar.get("timeZone")
+
+    try:
+        tz = ZoneInfo(timezone_name)
+    # If the calendar timezone name is not supported by Python's zoneinfo, use UTC
```
EricGustin, line 432: which calendar timezone names are not supported? If we had this list then couldn't we write the `timezone_name` to `tz` conversion for those unsupported timezones? i.e., we would never need to default to UTC?

toolkits/google/arcade_google/tools/calendar.py
```diff
@@ -356,3 +361,140 @@ async def delete_event(
...
+)
+async def find_time_slots_when_everyone_is_free(
+    context: ToolContext,
+    email_addresses: Annotated[
+        Optional[list[str]],
+        "The list of email addresses from people in the same organization domain (apart from the "
+        "currently logged in user) to search for free time slots. Defaults to None, which will "
+        "return free time slots for the current user only.",
+    ] = None,
+    start_date: Annotated[
```
EricGustin, line 381: why expose the search start (00:00:00) to the caller? Doesn't the caller/model only care about the time boundaries of the returned free slots?

byrro: Good question. When we don't expose, it affects how the LLM sets the `end_date`. For instance, if I prompt for free slots "today", it'd frequently set the `start_date` to today's date and the `end_date` to tomorrow's date. When we expose we'll search until 23:59:59, the LLM correctly sets both arguments to today's date.

toolkits/google/arcade_google/tools/calendar.py
byrro: In principle, both Google Calendar and Python zoneinfo module implement the [IANA tz database](https://www.iana.org/time-zones) and this exception should never be raised.

But we don't control Google responses (and I learned to not trust what 3pty APIs say in their docs). And we don't control the system's tz database implementation (where the engine is running). This is usually in `/usr/share/zoneinfo` or `/usr/share/zoneinfo.default`.

We can simply let an exception be raised when the two tz database implementations (Google & system) have any incompatibility. Or we can default to UTC and carry on the user request...

EricGustin (approved):
LGTM

PR #289, opened 2025-03-12
Fix examples

EricGustin:
Somehow these have slipped through the cracks. It hasn't been `authorization_url` for some time.


sdserranog (approved):
Good catch

PR #293, opened 2025-03-13
Math tools expanded

torresmateo:
Migrated all interfaces to get and return strings.

Added tests and evals for all functions (except the random generation)

### DIFF
--- a/toolkits/math/arcade_math/tools/arithmetic.py
+++ b/toolkits/math/arcade_math/tools/arithmetic.py
@@ -1,6 +1,4 @@
-import math
 from decimal import Decimal
-from statistics import median as stats_median
 from typing import Annotated
 
 from arcade.sdk import tool
@@ -54,25 +52,12 @@ def divide(
     b: Annotated[str, "The second number as a string"],
 ) -> Annotated[str, "The quotient of the two numbers as a string"]:
     """
-    Divide two numbers.
+    Divide two numbers
     """
     # Use Decimal for arbitrary precision
     a_decimal = Decimal(a)
     b_decimal = Decimal(b)
     return str(a_decimal / b_decimal)
-
-
-@tool
-def sqrt(
-    a: Annotated[str, "The number to square root as a string"],
-) -> Annotated[str, "The square root of the number as a string"]:
-    """
-    Get the square root of a number
-    If
-    """
-    # Use Decimal for arbitrary precision
-    a_decimal = Decimal(a)
-    return str(a_decimal.sqrt())
 
 
 @tool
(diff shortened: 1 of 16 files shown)


toolkits/math/arcade_math/tools/arithmetic.py
```diff
@@ -1,75 +1,268 @@
...
+    """
+    Calculate one number raised to the power of another.
+    """
+    # Use Decimal for arbitrary precision
+    return str(Decimal(a)**Decimal(b))
+
+
+@tool
+def abs_val(
+    a: Annotated[int, "The number as a string"],
```
EricGustin, line 131: int or str?

torresmateo: it should be str, fixed now, together with linter checks

EricGustin:
Feels weird to do `from math.arithmetic import my_trigonmetry_tool` or `from math.arithmetic import my_statistics_tool` 

Consider moving the non-arithmetic tools to their own file.

Otherwise, this looks good

torresmateo:
That's a good point. I'll reorganize the functions by general mathematical area

toolkits/math/arcade_math/tools/arithmetic.py
```diff
@@ -1,75 +1,268 @@
...
 @tool
 def divide(
-    a: Annotated[int, "The first number"], b: Annotated[int, "The second number"]
-) -> Annotated[float, "The quotient of the two numbers"]:
+    a: Annotated[str, "The first number as a string"],
+    b: Annotated[str, "The second number as a string"],
+) -> Annotated[str, "The quotient of the two numbers as a string"]:
     """
-    Divide two numbers
+    Divide two numbers.
```
nbarbettini, line 57: nit: None of the other descriptions have a `.` suffix

toolkits/math/arcade_math/tools/arithmetic.py
```diff
@@ -1,75 +1,268 @@
...
     """
     Sum all numbers from start through end
     """
-    return sum(list(range(start, end + 1)))
+    return str(sum(list(range(int(start), int(end) + 1))))
+
+
+# -----------------------
+# Additional useful tools
+# -----------------------
```
nbarbettini, line 102: nit: remove -- we don't really need "sections" in the file per se.

toolkits/math/tests/test_arithmetic.py
```diff
@@ -2,49 +2,366 @@
...
-    assert subtract(2, 1) == 1
-    assert subtract(2, 3.5) == -1.5
+@pytest.mark.parametrize(
+    "a, b, expected",
+    [
+        ("-1", "2", "-2"),
+        # TODO(mateo): ok to return negative 0?
+        ("-10", "0", "-0"),
+        ("0.5", "10.9", "5.45"),
+        # TODO(mateo): is scientific notation is acceptable as output?
```
nbarbettini, line 76: probably OK. I believe JS can parse these natively?

toolkits/math/tests/test_arithmetic.py
```diff
@@ -2,49 +2,366 @@
...
+    assert floor("1.0") == "1"
+    assert floor("-1.0") == "-1"
+    assert floor("3.14") == "3"
+    assert floor("0.4") == "0"
+    assert floor("-1.3") == "-2"
+
+
+def test_round_num():
+    # TODO(mateo): ok with scientific notatin? ok with negative round digits?
+    assert round_num("1.2345", "-2") == "0E+2"
```
nbarbettini, line 330: In this case I think scientific notation looks silly, because it's such a small number. Is this a side effect of using `Decimal`?

torresmateo: It is, it will return an integer if we use `float`. I will change it together with reorganizing the functions into math categories

toolkits/math/tests/test_arithmetic.py
torresmateo: Yes, JS will handle -0 and float variants gracefully

torresmateo:
> Feels weird to do `from math.arithmetic import my_trigonmetry_tool` or `from math.arithmetic import my_statistics_tool`
> 
> Consider moving the non-arithmetic tools to their own file.
> 
> Otherwise, this looks good

I changed this and separated files to follow [numpy's categories](https://numpy.org/doc/stable/reference/routines.math.html#rational-routines)

nbarbettini (approved):
LGTM!

PR #294, opened 2025-03-14
feat: Add Reddit and Twitch named providers

nbarbettini:
Adding auth support for future Reddit and Twitch tools.


EricGustin (approved):
woohoo!

PR #301, opened 2025-03-17
Deployment Secrets

sdreyer:
Allows environment variables to be used as secrets

### DIFF
--- a/arcade/arcade/worker/config/deployment.py
+++ b/arcade/arcade/worker/config/deployment.py
@@ -89,7 +89,7 @@ class Config(BaseModel):
         else:
             raise TypeError("Secret must be a string or a Secret object")
         # Check that the secret is not the default dev secret or empty
-        if secret.value.strip("") == "" or secret.value == "dev":
+        if secret.value.strip() == "" or secret.value == "dev":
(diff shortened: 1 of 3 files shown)


arcade/arcade/worker/config/deployment.py
```diff
@@ -63,33 +64,58 @@ class Pypi(PackageRepository):
...
+        # If the secret is a string, attempt to parse it as an environment variable or return the secret
+        if isinstance(v, str):
+            secret = get_env_secret(v)
+        # If the secret has been manually set, return it
+        elif isinstance(v, Secret):
+            secret = v
+        else:
+            raise TypeError("Secret must be a string or a Secret object")
+        # Check that the secret is not the default dev secret or empty
+        if secret.value.strip("") == "" or secret.value == "dev":
```
EricGustin, line 92: Should be
```py
if secret.value.strip() == "" or secret.value == "dev":
```

if the intended behavior is to strip whitespace from the string

EricGustin: i.e., `secret.value.strip("") == ""` isn't doing anything

EricGustin (approved):
lgtm

PR #303, opened 2025-03-17
Update Publish Toolkit Workflow

EricGustin:


.github/workflows/publish-toolkit.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         # Attempt to publish the toolkit to PyPI.
         # If the toolkit version already exists, skip publishing.
         # If another error occurs, exit with the error code so that the step fails.
-        PUBLISH_OUTPUT=$(poetry publish --skip-existing 2>&1)
+        PUBLISH_OUTPUT=$(poetry publish --skip-existing 2>&1 || true)
```
EricGustin, line 65: need to make it a conditional, otherwise when there is an error, the step will just exit without logging the error

.github/workflows/publish-toolkit.yml
```diff
@@ -77,7 +77,7 @@ jobs:
         fi
 
     - name: Send status to Slack
-      if: steps.Publish_Toolkit.outputs.skip_publish != 'true'
+      if: always() && steps.Publish_Toolkit.outputs.skip_publish != 'true'
```
EricGustin, line 80: need to add always() so that this step will run even when the previous step has a non-zero exit code

sdreyer (approved):

PR #313, opened 2025-03-18
Fix Slack bug when sending msg to a private channel

byrro:


EricGustin (approved):

EricGustin:
merging so that we can get this into the next release

PR #315, opened 2025-03-19
increased precision of decimal-based math tools to 100 digits

torresmateo:
Some edge cases of a benchmark were failing due to low precision, increased the precision to 100 digits.


torresmateo:
I tested this in a FastAPIWorker on modal and the increased precision works @nbarbettini

nbarbettini (approved):

PR #319, opened 2025-03-20
Fix Search.SearchJobs for Anthropic Models

EricGustin:
Fixes this error:

```bash
openai.InternalServerError: Error code: 503 - {'name': 'all_models_unavailable', 'message': "failed to chat with models; Attempt 0: anthropic - anthropic (type = error, message = tools.75.custom.input_schema.properties: Property keys should match pattern '^[a-zA-Z0-9_-]{1,64}$'); Attempt 1: anthropic - anthropic (type = error, message = tools.75.custom.input_schema.properties: Property keys should match pattern '^[a-zA-Z0-9_-]{1,64}$'); Attempt 2: anthropic - anthropic (type = error, message = tools.75.custom.input_schema.properties: Property keys should match pattern '^[a-zA-Z0-9_-]{1,64}$')"}
```

### DIFF
--- a/toolkits/search/arcade_search/tools/google_jobs.py
+++ b/toolkits/search/arcade_search/tools/google_jobs.py
@@ -23,7 +23,7 @@ async def search_jobs(
     ] = None,
     language: Annotated[
         str,
-        "2-character language code to use in the Google Jobs search. ",
+        "2-character language code to use in the Google Jobs search. "
         f"E.g. 'en' for English. Defaults to '{DEFAULT_GOOGLE_JOBS_LANGUAGE}'.",
     ] = DEFAULT_GOOGLE_JOBS_LANGUAGE,
     limit: Annotated[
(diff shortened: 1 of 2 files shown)


toolkits/search/arcade_search/tools/google_jobs.py
```diff
@@ -23,7 +23,7 @@ async def search_jobs(
     ] = None,
     language: Annotated[
         str,
-        "2-character language code to use in the Google Jobs search. ",
+        "2-character language code to use in the Google Jobs search. "
         f"E.g. 'en' for English. Defaults to '{DEFAULT_GOOGLE_JOBS_LANGUAGE}'.",
     ] = DEFAULT_GOOGLE_JOBS_LANGUAGE,
```
nbarbettini, line 28: For followup: What was this ultimately producing in the tool definition JSON? Could we catch this in the SDK and error gracefully?

nbarbettini (approved):

toolkits/search/arcade_search/tools/google_jobs.py
byrro: ah, this was making the annotation a tuple instead of a str, good catch!

PR #321, opened 2025-03-21
Google Sheets Tools

EricGustin:
| Name                     | Description                                                                           |
|--------------------------|---------------------------------------------------------------------------------------|
| Google.CreateSpreadsheet | Create a new spreadsheet with the provided title and data in its first sheet            |
| Google.GetSpreadsheet    | Get the user entered and formatted data for all sheets in the spreadsheet                           |
| Google.WriteToCell       | Write a value to a single cell in a spreadsheet.                                      |

## Google.CreateSpreadsheet
This tool can create a new spreadsheet with data in its first sheet
This tool takes in the data as a JSON string. Here's an example input: 
```
// Good at large payloads, sparse payloads, and contiguous data payloads.
// For example data[1]["D"] represents the value of the cell in the first row in the D column
{
  // All data in row 1
  1: {
    "A": 42, 
    "B": 2, 
    "D":"=A1+B1"
  },
  // All data in row 54
  54: {
    "A": "my string",
    "QQ": "my far away string"
  }
}
```
The above data format performed better on evals than the other two that I tested:
```
// Performed poorly at sparse data and also at larger amounts of data
[
  [42, 2, "", "=A1+B1"], 
  [], 
  [],
  ..., 
  ["A": "my string", "", "", ..., "my far away string"]
]
```
```
// Good at small payloads and sparse payloads, but very bad at payloads with contiguous data
{
  "A1": 42", "B1": 2, "D1": "=A1+B1", "A54": "my string", "QQ": "my far away string"
}
```

## Google.GetSpreadsheet
Gets the formatted values for all non empty cells in all sheets of the spreadsheet. The data returned is in a similar format as the `Google.CreateSpreadsheet` tool's `data` input parameter. The difference is that `get_spreadsheet` will return the user entered value (=A1+B1) and also the formatted value (23.4) for each cell.

## Google.WriteToCell
Writes to a single cell. At this point in time we do not support batch updating a sheet.


toolkits/google/arcade_google/tools/sheets.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+@tool(
+    requires_auth=Google(
+        scopes=["https://www.googleapis.com/auth/drive.file"],
+    )
+)
+def write_to_cell(
+    context: ToolContext,
+    spreadsheet_id: Annotated[str, "The id of the spreadsheet to write to"],
```
EricGustin, line 110: In the future we can also use `spreadsheet_name` if there is a Google Drive tool added for getting the name of spreadsheet by ID

PR #332, opened 2025-03-26
Dropbox Toolkit

byrro:
Basic tools to list items in a folder, search for files / folders, and download a file content.


toolkits/dropbox/arcade_dropbox/constants.py
EricGustin, line 1: `import`ing inside a constants file introduces the risk of circular imports (Importing `arcade_dropbox.enums` here means that `arcade_dropbox.enums` can't use any constants!)

This may not be an issue with the current implementation, but will be a pain for the next dev that may extend this toolkit.

I ran into this exact problem when I was extending the arcade-google toolkit.

byrro: True. In my mind, enums are actually like a collection of constant values. I would declare them in the `constants` module. Does that make sense?

toolkits/dropbox/arcade_dropbox/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        url += "/continue"
+        # If cursor is provided, every other argument must be ignored to avoid API error
+        json_data = {"cursor": json_data["cursor"]}
+
+    if endpoint_type == EndpointType.CONTENT:
+        headers["Dropbox-API-Arg"] = json.dumps(json_data)
+        json_data = {}
+
+    async with httpx.AsyncClient() as client:
+        response = await client.post(url, headers=headers, json=json_data)
```
EricGustin, line 48: Getting this error:
Screenshot 2025-03-26 at 3 09 36 PM


```python
# Setting json to None if the json_data is {} fixes the error
# Probably better to change the logic above such that json_data is set to None if {}
response = await client.post(url, headers=headers, json=json_data or None)
```

toolkits/dropbox/arcade_dropbox/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        "size_in_bytes": entry.get("size"),
+        "modified_datetime": entry.get("server_modified"),
+    }
+
+
+def clean_dropbox_entries(entries: list[dict]) -> list[dict]:
+    return [clean_dropbox_entry(entry) for entry in entries]
+
+
+def parse_dropbox_path(path: Optional[str]) -> Optional[str]:
```
EricGustin, line 86: Also, normalize `/` to empty string, else we get this error:
Screenshot 2025-03-26 at 3 15 20 PM

toolkits/dropbox/arcade_dropbox/tools/browse.py
```diff
@@ -0,0 +1,138 @@
...
+    """Provides a dictionary containing the list of items in the specified folder path.
+
+    Note 1: when paginating, it is not necessary to provide any other argument besides the cursor.
+    Note 2: when paginating, any given item (file or folder) may be returned in multiple pages.
+    """
+    limit = min(limit, 2000)
+
+    try:
+        result = await send_dropbox_request(
+            None if not context.authorization else context.authorization.token,
```
EricGustin, line 50: migrate to using [get_auth_token_or_empty](https://github.com/ArcadeAI/arcade-ai/blob/887ded441edbc0c31f8d38be9b1c99adf0887394/arcade/arcade/core/schema.py#L265)

toolkits/dropbox/evals/eval_download_file.py
```diff
@@ -0,0 +1,89 @@
...
+                func=download_file,
+                args={
+                    "file_path": "test.txt",
+                    "file_id": None,
+                },
+            ),
+        ],
+        critics=[
+            DropboxPathCritic(critic_field="file_path", weight=0.5),
+            NoneCritic(critic_field="file_id", weight=0.5),
```
EricGustin, line 49: as discussed offline, these should not use NoneCritic if you want them to be critisized

toolkits/dropbox/arcade_dropbox/constants.py
EricGustin: yeah, unless they [look like this enum](https://github.com/ArcadeAI/arcade-ai/blob/887ded441edbc0c31f8d38be9b1c99adf0887394/toolkits/google/arcade_google/models.py#L13)

toolkits/dropbox/arcade_dropbox/constants.py
```diff
@@ -0,0 +1,10 @@
+from arcade_dropbox.enums import Endpoint, EndpointType
+
+API_BASE_URL = "https://{endpoint_type}.dropboxapi.com"
+API_VERSION = "2"
+ENDPOINT_URL_MAP = {
+    Endpoint.LIST_FOLDER: (EndpointType.API, "files/list_folder"),
+    Endpoint.SEARCH_FILES: (EndpointType.API, "files/search_v2"),
+    Endpoint.DOWNLOAD_FILE: (EndpointType.CONTENT, "files/download"),
+}
+MAX_RESPONSE_BODY_SIZE = 10 * 1024 * 1024  # 10 MB
```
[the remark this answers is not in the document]
byrro: imagine the catastrophic consequences if you had not caught that bug 😂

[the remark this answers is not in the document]
byrro: @nbarbettini engine also needs urgent fix: https://github.com/ArcadeAI/Engine/blob/ab0a90f3b5161e23f3f3a3adac57cfeeba360a64/internal/directors/workers/config/http.go#L20

byrro:
@EricGustin addressed all comments from your review 👍

toolkits/dropbox/arcade_dropbox/constants.py
EricGustin, line 1: enums are defined in both constants.py and enums.py?

toolkits/dropbox/arcade_dropbox/utils.py
```diff
@@ -0,0 +1,106 @@
...
+    if not isinstance(path, str):
+        return None
+
+    if not path:
+        return ""
+
+    if path in ["/", "\\"]:
+        return None
```
EricGustin, line 100: should be empty string, yeah?

Screenshot 2025-03-27 at 9 56 29 AM

toolkits/dropbox/arcade_dropbox/constants.py
byrro: removed

toolkits/dropbox/arcade_dropbox/utils.py
byrro: fixed

toolkits/dropbox/arcade_dropbox/utils.py
```diff
@@ -0,0 +1,106 @@
...
+    if json_data.get("cursor"):
+        url += "/continue"
+        # If cursor is provided, every other argument must be ignored to avoid API error
+        json_data = {"cursor": json_data["cursor"]}
+
+    if endpoint_type == EndpointType.CONTENT:
+        headers["Dropbox-API-Arg"] = json.dumps(json_data)
+        json_data = {}
+
+    async with httpx.AsyncClient() as client:
```
EricGustin, line 48: why not use the [Offical Dropbox Python SDK](https://github.com/dropbox/dropbox-sdk-python)?

EricGustin (approved):

toolkits/dropbox/arcade_dropbox/utils.py
byrro: it should probably work fine on newer Python versions. but the main issue I think is that all HTTP requests are blocking. using httpx will provide more freedom in future tools that require more complex networking patterns without compromising performance

PR #345, opened 2025-04-04
Common changes in all toolkits

EricGustin:
Addresses general improvements to all toolkits including changing ruff from python 3.9 to python 3.10 which is the reason for the removal of Optional[] among others.

Also, turns out that our `make install` for toolkits wasn't correctly checking for whether poetry was installed (&> /dev/null syntax isn't supported by our check-toolkits GitHub action, so we were installing poetry twice. I replaced with the more portable >/dev/null 2>&1)

Question: Should we also change ruff to py310 for the `arcade/` package in a later PR?

-------------------

CU-86b4gzyp6

### DIFF
--- a/arcade/arcade/templates/{{ toolkit_name }}/.ruff.toml
+++ b/arcade/arcade/templates/{{ toolkit_name }}/.ruff.toml
@@ -1,4 +1,4 @@
-target-version = "py39"
+target-version = "py310"
 line-length = 100
 fix = true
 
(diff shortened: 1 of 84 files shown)


Spartee:
Task linked: [CU-86b4gzyp6 Arcade Toolkit Package Improvements](https://app.clickup.com/t/86b4gzyp6)

wdawson (approved):
LGTM

PR #347, opened 2025-04-04
Support PEP 604 Union Types

EricGustin:
⚠️ Local Worker will not work with Slack toolkit until this is merged 

Support [PEP 604](https://peps.python.org/pep-0604/) Union Types for tool input param types and return types.

https://github.com/ArcadeAI/arcade-ai/pull/345 introduced PEP 604 Union Types into our toolkits, but the worker did not support this syntax, so it was failing on start up if a tool used bar syntax in its return type (Slack).

Additionally, optionals defined with `Union[T, None]` were already supported, but didn't have any unit tests, so I added them.

### DIFF
--- a/arcade/arcade/core/catalog.py
+++ b/arcade/arcade/core/catalog.py
@@ -48,6 +48,7 @@ from arcade.core.toolkit import Toolkit
 from arcade.core.utils import (
     does_function_return_value,
     first_or_none,
+    is_strict_optional,
     is_string_literal,
     is_union,
     snake_to_pascal_case,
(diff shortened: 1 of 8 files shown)


arcade/tests/core/test_catalog.py
```diff
@@ -94,14 +94,10 @@ class InvalidTool:
...
-        # Assert that ToolDefinitionError is raised with the correct message
-        with pytest.raises(ToolDefinitionError) as exc_info:
+        # Assert that ToolDefinitionError is raised
+        with pytest.raises(ToolDefinitionError):
             catalog.add_toolkit(mock_toolkit)
 
-        assert "Type error encountered while adding tool invalid_tool from mock_module" in str(
-            exc_info.value
-        )
-
```
EricGustin, line 104: fragile test - we only care that the `ToolDefinitionError` is raised

arcade/arcade/core/catalog.py
```diff
@@ -481,10 +482,10 @@ def create_output_definition(func: Callable) -> ToolOutput:
         return_type = return_type.__origin__
 
     # Unwrap Optional types
-    is_optional = False
-    if get_origin(return_type) is Union and type(None) in get_args(return_type):
+    # Both Optional[T] and T | None are supported
+    is_optional = is_strict_optional(return_type)
+    if is_optional:
```
nbarbettini, line 487: nit: could be combined
```suggestion
    if is_strict_optional(return_type):
```

nbarbettini (approved):

arcade/arcade/core/catalog.py
EricGustin: saving it in variable `is_optional` so that I don't have to call `is_strict_optional` again on line 493

PR #352, opened 2025-04-07
Add More Reddit Tools

EricGustin:
| Name                             | Description                                                             |
|----------------------------------|-------------------------------------------------------------------------|
| Reddit.CheckSubredditAccess      | Checks whether the specified subreddit exists and also if it is accessible |
| Reddit.GetSubredditRules         | Gets the rules of the specified subreddit                               |
| Reddit.GetMyUsername             | Get the Reddit username of the authenticated user                       |
| Reddit.GetMyPosts                | Get posts that were created by the authenticated user sorted by newest  |

Four new tools that I had to create for my demo app.

### DIFF
--- a/toolkits/reddit/arcade_reddit/tools/read.py
+++ b/toolkits/reddit/arcade_reddit/tools/read.py
@@ -2,6 +2,7 @@ from typing import Annotated
 
 from arcade.sdk import ToolContext, tool
 from arcade.sdk.auth import Reddit
+from arcade.sdk.errors import ToolExecutionError
 
 from arcade_reddit.client import RedditClient
 from arcade_reddit.enums import (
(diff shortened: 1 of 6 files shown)


toolkits/reddit/arcade_reddit/tools/read.py
```diff
@@ -118,8 +120,83 @@ async def get_top_level_comments(
...
+    return username
+
+
+@tool(requires_auth=Reddit(scopes=["identity", "history", "read"]))
+async def get_my_posts(
+    context: ToolContext,
+    limit: Annotated[
+        int, "The maximum number of posts to fetch. Default is 10. Maximum is 100"
+    ] = 10,
+    include_body: Annotated[
```
EricGustin, line 184: I'm on the fence as to whether we should include this param or just always return the body of the posts

byrro: I don't see a reason that justifies this argument. But the tool interface is very simple, so keeping it will not create any unmanageable complexity for LLMs to deal with. Your call...

byrro (approved):
LGTM

PR #353, opened 2025-04-08
Add example for calling a tool that requires a secret

EricGustin:


Spartee (commented):
I think this example should be simpler. Show the tool, tell them to use arcade deploy, set secret,... etc

I thikn showing ``requires_secret`` is more important here.

PR #357, opened 2025-04-10
Tool Metadata

EricGustin:
Adds the ability to specify required metadata for a tool which is injected into the tool's context before execution.

Example of a tool that requires metadata:
```python
from typing import Annotated

from arcade.sdk import ToolContext, ToolMetadataKey, tool
from arcade.sdk.auth import Google


@tool(
    requires_auth=Google(),
    requires_metadata=[ToolMetadataKey.CLIENT_ID, ToolMetadataKey.COORDINATOR_URL])
def my_metadata_tool(context: ToolContext) -> Annotated[str, "Some metadata"]:
    """Gets some metadata"""
    google_client_id = context.get_metadata(ToolMetadataKey.CLIENT_ID)
    cloud_coordinator_url = context.get_metadata(ToolMetadataKey.COORDINATOR_URL)
    
    return f"My Google Client ID: {google_client_id}, My Cloud Coordinator URL: {cloud_coordinator_url}"

### DIFF
--- a/arcade/arcade/core/schema.py
+++ b/arcade/arcade/core/schema.py
@@ -109,6 +109,13 @@ class ToolSecretRequirement(BaseModel):
     """The ID of the secret."""
 
 
+class ToolMetadataRequirement(BaseModel):
+    """A requirement for a tool to run."""
+
+    key: str
+    """The ID of the metadata."""
+
+
 class ToolRequirements(BaseModel):
     """The requirements for a tool to run."""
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: arcade/arcade/core/schema.py [hunk 1]
```diff
@@ -109,13 +110,37 @@ class ToolSecretRequirement(BaseModel):
...
 
 
+class ToolMetadataKey(str, Enum):
+    """Convience enum for commonly used metadata keys."""
+
+    CLIENT_ID = "client_id"
+    COORDINATOR_URL = "coordinator_url"
+
+    @staticmethod
+    def requires_auth(key: str) -> bool:
```
nbarbettini, line 120: naming nit: since `requires_auth` is already an important name used elsewhere in the project, maybe `needs_auth_requirement`? Just so devs don't get confused by this name.

nbarbettini (approved):

PR #360, opened 2025-04-14
Salesforce toolkit

byrro:
Reviewer: DM me on slack so that I can send you client ID/Secret and a sample user account to try out the toolkit


toolkits/salesforce/arcade_salesforce/models.py
```diff
@@ -0,0 +1,344 @@
...
+            return None
+
+    async def get_account_contacts(self, account_id: str, limit: Optional[int] = 10) -> list[dict]:
+        contacts = await self._get_related_objects(
+            SalesforceObject.CONTACT, SalesforceObject.ACCOUNT, account_id, limit
+        )
+
+        return [
+            clean_contact_data(contact)
+            for contact in await asyncio.gather(*[
```
EricGustin, line 168: I worry about getting rate limited with this sort of stuff. Although it decreases tool execution time, it increases the risk of being rate limited.

If we are invoking multiple API calls concurrently like we are here, then implementing a rate limiting strategy would be preferred

toolkits/salesforce/arcade_salesforce/models.py
```diff
@@ -0,0 +1,344 @@
...
+                except ValueError:
+                    obj_type = get_object_type(item) + "s"
+
+                if obj_type not in account_data:
+                    account_data[obj_type] = []
+                account_data[obj_type].append(item)
+
+        return await self.expand_account_associations(account_data)
+
+    async def expand_account_associations(self, account: dict) -> dict:
```
EricGustin, line 271: docstrings for this, and others, would be helpful

toolkits/salesforce/arcade_salesforce/models.py
```diff
@@ -0,0 +1,344 @@
...
+        account_id: Optional[str] = None,
+        account_data: Optional[dict[str, Any]] = None,
+        limit_per_association: Optional[int] = 10,
+    ) -> dict:
+        """Enrich account dictionary with notes, leads, contacts, etc.
+
+        Must provide either `account_id` or `account_data`.
+        """
+        if (account_id and account_data) or (not account_id and not account_data):
+            raise ValueError("Must provide either `account_id` or `account_data`")
```
EricGustin, line 240: Remove ambiguity between OR and XOR
```suggestion
            raise ValueError("Must provide exactly one of `account_id` or `account_data`")
```

toolkits/salesforce/arcade_salesforce/tools/crm/account.py
```diff
@@ -0,0 +1,127 @@
...
+            clean_account_data(account_data)
+            for account_data in await asyncio.gather(*[
+                client.enrich_account(
+                    account_data=account,
+                    limit_per_association=10,
+                )
+                for account in accounts
+            ])
+        ]
+    }
```
EricGustin, line 86: fancy, but very difficult to debug. any way we can break this out into more than one block?

toolkits/salesforce/arcade_salesforce/tools/crm/account.py
```diff
@@ -0,0 +1,127 @@
...
+            "read_opportunity",
+            "read_task",
+        ],
+    )
+)
+async def get_account_data_by_keywords(
+    context: ToolContext,
+    query: Annotated[
+        str,
+        "The query to search for accounts. It will match the keywords against all account fields, "
```
EricGustin, line 32: note: i was being adversarial when trying this tool out, but sometimes LLM would provide empty string for `query`. Maybe we can add an extra note in the annotation like "Must be longer than one character" to help combat this

toolkits/salesforce/arcade_salesforce/tools/crm/account.py
```diff
@@ -0,0 +1,127 @@
...
+    "The accounts matching the query with related info: contacts, leads, notes, calls, "
+    "opportunities, tasks, emails, and events (up to 10 items of each type)",
+]:
+    """Searches for accounts in Salesforce and returns them with related info: contacts, leads,
+    notes, calls, opportunities, tasks, emails, and events (up to 10 items of each type).
+
+    An account is an organization (such as a customer, supplier, or partner, though more commonly
+    a customer). In some Salesforce account setups, an account can also represent a person.
+    """
+    if not query:
```
EricGustin, line 53: also need to check for single character query.

```
{'name': 'non_retryable_tool_call_error', 'message': "tool call failed - 
Salesforce.GetAccountDataByKeywords failed: Bad request. Errors: ['search term must be longer than one character: ']: Bad 
request. Errors: ['search term must be longer than one character: ']"}
```

toolkits/salesforce/arcade_salesforce/models.py
byrro: I added a [semaphore in the `SalesforceClient` class](https://github.com/ArcadeAI/arcade-ai/blob/34742b4970dc189f87665dd83a8881f3c757e126/toolkits/salesforce/arcade_salesforce/models.py#L40) limiting concurrency. The [default is 3 requests, configurable through an env var](https://github.com/ArcadeAI/arcade-ai/blob/34742b4970dc189f87665dd83a8881f3c757e126/toolkits/salesforce/arcade_salesforce/constants.py#L5).

EricGustin (approved):
lgtm

PR #361, opened 2025-04-15
Add Google File Picker Tool

EricGustin:
## Google File Picker
The Google Picker lets users select or upload Google Drive files. Users can grant permission to your apps to access their Drive data, providing a secure and authorized way to interact with their files.

The `generate_google_file_picker_url` returns a URL to a Google File Picker for the user.


toolkits/google/arcade_google/tools/__init__.py
```diff
@@ -1 +1,94 @@
-__all__ = ["gmail", "calendar", "drive", "docs"]
+from arcade_google.tools.calendar import (
```
EricGustin, line 1: sneaking in these re-exports

sdserranog (approved):

PR #365, opened 2025-04-17
Bump Python version for Worker's Dockerfile

EricGustin:
[build-and-publish](https://github.com/ArcadeAI/arcade-ai/actions/runs/14522118587/job/40745392086) is failing to build the Worker container because the arcade-stripe toolkit requires Python >=3.11. The arcade-stripe toolkit has a dependency that requires >=3.11, so we can't loosen the toolkit's requirement.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim
+FROM python:3.11-slim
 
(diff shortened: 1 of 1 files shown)


nbarbettini (approved):
Seems OK to me 👍

PR #368, opened 2025-04-18
Support for MCP stdio transport

Spartee:
MCP stdio Implementation:

The PR adds support for standard input/output (stdio) as a transport mechanism for the Message Control Protocol. This is a replacement to the SSE (Server-Sent Events) transport that was worked on in PR #359 but will not be merged as it's not deprecated.

This will allow developers to use Arcade tools (written by the dev or Arcade) in Claude, Cursor, windsurf, etc. 

The engine Gateway already supports adding HTTPS streamable (replacement for SSE) MCP servers as tool servers, and will soon support full gateway capability in the client API as well. 

To use any existing Toolkit just 

### Quickstart setup with existing toolkits

```bash
pip install arcade-ai
pip install <name of toolkit> # ex. arcade-google
arcade serve --mcp
```

### Run with Claude

Just add the following to the Claude config

```json
{
  "mcpServers": {
    "arcade": {
      "command": "bash",
      "args": ["-c", "export ARCADE_API_KEY=arc_xxxx && /path/to/python /path/to/arcade serve --mcp"]
    }
  }
}
```

### Customizing the Tool Server

Developers can customize their served tools and server furthermore by importing the worker sdk

```python

import arcade_google  # pip install arcade_google
import arcade_search  # pip install arcade_search

from arcade.core.catalog import ToolCatalog
from arcade.worker.mcp.stdio import StdioServer

# 2. Create and populate the tool catalog
catalog = ToolCatalog()
catalog.add_module(arcade_google)  # Registers all tools in the package
catalog.add_module(arcade_search)


# 3. Main entrypoint
async def main():
    # Create the worker with the tool catalog
    worker = StdioServer(catalog)

    # Run the worker
    await worker.run()


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
    
 ```
 
 Then to run with claude, just run this python file instead of the prebuilt server used in ``arcade serve --mcp``

### CODE REVIEW

File: arcade/arcade/core/schema.py [hunk 1]
```diff
@@ -324,10 +324,18 @@ def _get_item(
...
+            if item.key.lower() == normalized_key or item.key.upper() == normalized_key:
                 return item.value
+
         raise ValueError(f"{item_name.capitalize()} {key} not found in context.")
 
+    def set_secret(self, key: str, value: str) -> None:
+        """Set a secret for the tool invocation."""
+        if not self.secrets:
+            self.secrets = []
+        secret = ToolSecretItem(key=str(key).lower(), value=str(value))
```
EricGustin, line 336: the `key` should remain the same casing as what the environment variable is. i.e., remove the `.lower()`.

The current behavior is equivalent to the following:
```
>export ABC=123
>echo $ABC // cannot find the envvar because it was converted to lower case
>echo $abc // echoes back 123
```

arcade/arcade/worker/core/common.py
```diff
@@ -1,10 +1,20 @@
...
 
+primitives = TypeVar("primitives", bound=str | int | float | bool | list)
+CatalogResponse = list[ToolDefinition]
+ToolCallResponse = ToolCallResponse
+HealthCheckResponse = dict[str, str]
+MetricsResponse = dict[str, Any]
+JSONResponse = dict[str, Any]
+ResponseData = (
+    CatalogResponse | ToolCallResponse | HealthCheckResponse | MetricsResponse | JSONResponse
+)
```
EricGustin, line 16: some of these might not be needed. as discussed in person

EricGustin (commented):
I got tools w/ secret requirements and also tools with auth requirements working with Claude Desktop. Super cool

EricGustin (approved):

PR #377, opened 2025-04-23
Microsoft Outlook Mail Tools

EricGustin:
| Name                         | Description                                                                    |
|------------------------------|--------------------------------------------------------------------------------|
| Microsoft.CreateDraftEmail   | Compose a new draft email in Outlook                                           |
| Microsoft.UpdateDraftEmail   | Update an existing draft email in Outlook                                      |
| Microsoft.CreateAndSendEmail | Create and immediately send a new email in Outlook to the specified recipients |
| Microsoft.SendDraftEmail     | Send an existing draft email in Outlook                                        |
| Microsoft.ReplyToEmail       | Reply only to the sender of an existing email in Outlook                       |
| Microsoft.ReplyAllToEmail    | Reply to all recipients of an existing email in Outlook                        |
| Microsoft.ListEmails         | List emails in the user's mailbox across all folders                           |
| Microsoft.ListEmailsInFolder | List the user's emails in the specified folder                                 |


toolkits/microsoft/arcade_microsoft/outlook_mail/tools/read.py
```diff
@@ -0,0 +1,89 @@
...
+        str | None, "The pagination token to continue a previous request"
+    ] = None,
+) -> Annotated[
+    dict, "A dictionary containing a list of emails and a pagination token, if applicable"
+]:
+    """List the user's emails in the specified folder.
+
+    Exactly one of `well_known_folder_name` or `folder_id` MUST be provided.
+    """
+    if not (bool(well_known_folder_name) ^ bool(folder_id)):
```
byrro, line 70: not really important, but isn't it more readable to just to do `if not well_known_folder_name and not folder_id:`?

I believe the `not` operator uses the `__bool__` magic method

toolkits/microsoft/arcade_microsoft/outlook_mail/tools/read.py
```diff
@@ -0,0 +1,89 @@
...
+        raise ToolExecutionError(
+            message="Exactly one of `well_known_folder_name` or `folder_id` must be provided."
+        )
+    folder_name = well_known_folder_name.value if well_known_folder_name else folder_id
+    client = get_client(context.get_auth_token_or_empty())
+    request_config = _prepare_list_emails_request_config(limit)
```
byrro, line 76: Warning of pedantic comment ahead: conventionally, starting a variable or function with `_` indicates it's a non-public one (not meant to be used outside its own scope). In this case, the `_prepare_list_emails_request_config` (and a few others) has the leading underscore, which seems to invoke the convention. But they are used outside the scope of the `utils` module. Perhaps removing the leading `_` would keep it consistent with the convention.

toolkits/microsoft/arcade_microsoft/outlook_mail/tools/read.py
EricGustin: but that doesn't cover the case for both being provided

toolkits/microsoft/arcade_microsoft/outlook_mail/tools/read.py
EricGustin: i was trying to convey that anyone using the package should not import anything in utils.py. I think I'll just rename the file to `_utils.py` and then get rid of the underscores in the function names

EricGustin: updated

toolkits/microsoft/arcade_microsoft/outlook_mail/utils.py
```diff
@@ -0,0 +1,78 @@
...
+async def _send_reply_email(
+    context: ToolContext,
+    message_id: str,
+    body: str,
+    reply_type: Literal["reply", "reply_all"],
+) -> dict:
+    """Send a reply email to the sender of an existing email."""
+    client = get_client(context.get_auth_token_or_empty())
+
+    if reply_type == "reply":
```
byrro: this `if / elif` block could be simplified with:

```
await getattr(client.me.messages.by_message_id(message_id), reply_type).post(reply_request_body)
```

or, perhaps more readable:

```
reply = getattr(client.me.messages.by_message_id(message_id), reply_type)
await reply.post(reply_request_body)
```

EricGustin: i like it, but I would still need the `if / elif` for `ReplyPostRequestBody` vs `ReplyAllPostRequestBody`, right?

toolkits/microsoft/arcade_microsoft/outlook_mail/tools/send.py
```diff
@@ -0,0 +1,101 @@
...
+    await client.me.messages.by_message_id(message_id).send.post()
+
+    return {
+        "success": True,
+        "message": "Email sent successfully",
+    }
+
+
+@tool(requires_auth=Microsoft(scopes=["Mail.Send"]))
+async def reply_to_email(
```
byrro, line 77: It would be more concise (and perhaps less room for LLM confusion also?) if we had a single tool `reply_to_email` with a `reply_to` argument accepting an enum:

```
class ReplyTo(enum):
    ALL_RECIPIENTS = "all_recipients"
    ONLY_THE_SENDER = "only_the_sender"
```

byrro: `SENDER_AND_ALL_RECIPIENTS = "sender_and_all_recipients" ` would be even more clear to LLMs

EricGustin: looking forward, would you expect forwarding an email to be included in this tool's capabilities?

EricGustin: I combined the two into a single tool

byrro (approved):

PR #378, opened 2025-04-23
fix: Typo in Sqrt tool description

nbarbettini:


nbarbettini:
@torresmateo Was there something else you wanted to add to this description?

torresmateo (approved):

torresmateo:
I originally had "If the number is negative, raises an error". But then I removed it (except the "If") as no other tool had that.

PR #382, opened 2025-04-28
Add Outlook Calendar tools

EricGustin:
| Name                         | Description                                                              |
|-------------------------------|--------------------------------------------------------------------------|
| Microsoft.GetEvent            | Get an event by its ID from the user's calendar.                         |
| Microsoft.ListEventsInTimeRange| List events in the user's calendar in a specific time range.             |
| Microsoft.CreateEvent         | Create an event in the authenticated user's default calendar.            |


toolkits/microsoft/arcade_microsoft/outlook_calendar/tools/list_events_in_time_range.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Ignores timezone offsets provided in the start_date_time and end_date_time parameters.
+    Instead, uses the user's default calendar timezone to filter events.
+    If the user has not set a timezone for their calendar, then the timezone will be UTC.
+    """
+    client = get_client(context.get_auth_token_or_empty())
+
+    time_zone = await get_default_calendar_timezone(client)
+    time_zone_offset = convert_timezone_to_offset(time_zone)
+    start_date_time = replace_timezone_offset(start_date_time, time_zone_offset)
```
byrro, line 44: perhaps adding a datetime format validation before using the arg value? would allow to raise an error that is more descriptive of what needs to be corrected

toolkits/microsoft/arcade_microsoft/outlook_calendar/tools/list_events_in_time_range.py
```diff
@@ -0,0 +1,56 @@
...
+        str,
+        "The start date and time of the time range, represented in "
+        "ISO 8601 format. Timezone offset is ignored. For example, 2025-04-24T19:00:00",
+    ],
+    end_date_time: Annotated[
+        str,
+        "The end date and time of the time range, represented in "
+        "ISO 8601 format. Timezone offset is ignored. For example, 2025-04-24T19:30:00",
+    ],
+    limit: Annotated[int, "The maximum number of events to return. Max 1000. Defaults to 10"] = 10,
```
byrro, line 32: Based on the [Outlook docs sample response](https://learn.microsoft.com/en-us/graph/api/calendar-list-calendarview?view=graph-rest-1.0&tabs=python#response-1), an event dict has ~300 chars. If a user searches for a long enough period and retrieves 1000 events, the response may potentially contain 300,000 chars, which seems too big.

Perhaps 100 would be more appropriate?

Although the API doesn't offer pagination, we could paginate inside the tool and return `events[offset:offset+limit]`. That would keep response size relatively small, while still allowing to retrieve up to 1,000 events.

toolkits/microsoft/arcade_microsoft/outlook_calendar/tools/create_event.py
```diff
@@ -0,0 +1,68 @@
...
+        "The datetime of the event's end, represented in "
+        "ISO 8601 format. Timezone offset is ignored. For example, 2025-04-25T13:30:00",
+    ],
+    location: Annotated[str | None, "The location of the event"] = None,
+    attendee_emails: Annotated[
+        list[str] | None,
+        "The email addresses of the attendees of the event. "
+        "Must be valid email addresses e.g., username@domain.com.",
+    ] = None,
+    is_online_meeting: Annotated[
```
byrro, line 40: should we also support adding [online meeting info](https://learn.microsoft.com/en-us/graph/api/resources/onlinemeetinginfo?view=graph-rest-1.0)? perhaps at least an `online_meeting_url`, which is the most common scenario nowadays?

I can see people asking the LLM to book a conference in Zoom or Meet and then asking to set up the meeting with the conference room url, for intance.

toolkits/microsoft/arcade_microsoft/outlook_calendar/tools/create_event.py
```diff
@@ -0,0 +1,68 @@
...
+    """Create an event in the authenticated user's default calendar.
+
+    Ignores timezone offsets provided in the start_date_time and end_date_time parameters.
+    Instead, uses the user's default calendar timezone to filter events.
+    If the user has not set a timezone for their calendar, then the timezone will be UTC.
+    """
+    client = get_client(context.get_auth_token_or_empty())
+
+    time_zone = await get_default_calendar_timezone(client)
+    start_date_time = remove_timezone_offset(start_date_time)
```
byrro, line 53: Same comment as before. Would be good to validate the datetime values before using them. LLMs sometimes get confused about datetime string formats. Having a more descriptive `RetryableToolError` could potentially improve performance in a second try.

toolkits/microsoft/arcade_microsoft/outlook_calendar/tools/create_event.py
```diff
@@ -0,0 +1,68 @@
...
+    end_date_time: Annotated[
+        str,
+        "The datetime of the event's end, represented in "
+        "ISO 8601 format. Timezone offset is ignored. For example, 2025-04-25T13:30:00",
+    ],
+    location: Annotated[str | None, "The location of the event"] = None,
+    attendee_emails: Annotated[
+        list[str] | None,
+        "The email addresses of the attendees of the event. "
+        "Must be valid email addresses e.g., username@domain.com.",
```
byrro, line 38: perhaps add an email validation step before trying to create the event with this arg values?

EricGustin: good point. I had previously assumed that microsoft api would return a helpful error in this case, but turns out it doesn't

byrro (approved):

PR #387, opened 2025-05-02
Fix local worker toolkit registration

EricGustin:
This PR fixes two bugs:
1. `serve_default_worker` was not registering toolkits to the worker (non mcp). This caused workers created during `arcade workerup` and `arcade serve` to have 0 tools.
2. We were logging the number of files in the toolkit as the number of tools in the toolkit.

### DIFF
--- a/arcade/arcade/cli/serve.py
+++ b/arcade/arcade/cli/serve.py
@@ -239,13 +239,20 @@ def serve_default_worker(
 
     toolkits = discover_toolkits()
     logger.info("Serving the following toolkits:")
+    toolkit_tool_counts = {}
+    for toolkit in toolkits:
+        for _, tools in toolkit.tools.items():
+            toolkit_tool_counts[toolkit.name] = toolkit_tool_counts.get(toolkit.name, 0) + len(
+                tools
+            )
     for toolkit in toolkits:
         if debug:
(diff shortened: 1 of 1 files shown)


arcade/arcade/cli/serve.py
```diff
@@ -239,13 +239,20 @@ def serve_default_worker(
 
     toolkits = discover_toolkits()
     logger.info("Serving the following toolkits:")
+    toolkit_tool_counts: dict[str, int] = {}
```
EricGustin, line 242: correctly count the number of tools in each toolkit

Spartee:
Good catch

arcade/arcade/cli/serve.py
```diff
@@ -267,18 +274,23 @@ def serve_default_worker(
...
         app, enable=enable_otel, log_level=logging.DEBUG if debug else logging.INFO
     )
-    _ = FastAPIWorker(
+    worker = FastAPIWorker(
         app=app,
         secret=secret,
         disable_auth=disable_auth,
         otel_meter=otel_handler.get_meter(),
     )
+    for toolkit in toolkits:
```
EricGustin, line 291: we need to register the toolkits so that the worker is aware of them :)

nbarbettini (approved):

PR #404, opened 2025-05-20
Standardize notion toolkit name

byrro:
The toolkit was named `arcade_notion_toolkit`. This renames it to `arcade_notion`, in line with the standard followed by every other toolkit.


wdawson (approved):

EricGustin:
My understanding is that the PyPI account `spartee` does not own the PyPI package `arcade_notion`, so publishing this toolkit will fail.

EricGustin:
The Arcade.dev PyPI organization now owns the `arcade_notion` PyPI package.

PR #405, opened 2025-05-20
Ericgustin/arcade expert tk

EricGustin:
This is a unique type of toolkit and is still in a nascent stage, but it has plenty of potential. This PR introduces the first tool for an Arcade Toolkit for Arcade (meta, I know).

This PR introduces `ArcadeExpert.SearchDocumentation`. Given a provided query, it leverages Arcade documentation's `llm.txt` to extract up to 5 relevant links and scrape their content. It uses an LLM to determine the most relevant links. Using similarity search instead of an LLM to determine the most relevant links would probably be the more idiomatic way to do this,  but simplicity wins for now.

Ideally, this toolkit would be added as an MCP server to an Arcade user's IDE.

### DIFF
--- a/toolkits/arcade_expert/arcade_arcade_expert/tools/search_documentation.py
+++ b/toolkits/arcade_expert/arcade_arcade_expert/tools/search_documentation.py
@@ -4,6 +4,7 @@ import httpx
 from arcade.sdk import ToolContext, tool
 from markdownify import markdownify
 from openai import OpenAI
+from openai.types.chat import ParsedChatCompletion
 
 from arcade_arcade_expert.models import Links
 
@@ -33,7 +34,7 @@ async def search_documentation(
         data = markdownify(response.text)
 
     # Get relevant links from the llms.txt file
-    response = openai_client.beta.chat.completions.parse(
+    chat_response: ParsedChatCompletion = openai_client.beta.chat.completions.parse(
         model="gpt-4o",
         messages=[
             {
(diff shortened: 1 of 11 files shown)


toolkits/arcade_expert/pyproject.toml
```diff
@@ -0,0 +1,43 @@
+[tool.poetry]
+name = "arcade_arcade_expert"
```
EricGustin, line 2: Given that the toolkit name starts with `arcade` and official Arcade toolkits start with `arcade`, we end up with `arcade_arcade_expert`. If we were to change it to `arcade_expert`, then the toolkit name would be `Expert` instead of `ArcadeExpert`.

Pip installing:
`pip install arcade_arcade_expert`

Calling an ArcadeExpert tool:
`arcade_client.tools.execute("ArcadeExpert.SearchDocumentation", ...)`

toolkits/arcade_expert/arcade_arcade_expert/tools/search_documentation.py
```diff
@@ -0,0 +1,67 @@
...
+import httpx
+from arcade.sdk import ToolContext, tool
+from markdownify import markdownify
+from openai import OpenAI
+from openai.types.chat import ParsedChatCompletion
+
+from arcade_arcade_expert.models import Links
+
+
+@tool(requires_secrets=["OPENAI_API_KEY"])
```
EricGustin, line 12: If we decide to have this in the prod worker, then we will need to add an `OPENAI_API_KEY` secret for all users

EricGustin:
@evantahler thoughts on getting this merged now as a 'community' toolkit?

evantahler (approved):

PR #406, opened 2025-05-21
Add Confluence toolkit

EricGustin:
12 tools for the new Confluence toolkit.

| Name                             | Description                                                                        |
|----------------------------------|------------------------------------------------------------------------------------|
| Confluence.CreatePage            | Create a new page at the root of the given space.                                  |
| Confluence.UpdatePageContent     | Update a page's content.                                                           |
| Confluence.RenamePage            | Rename a page by changing its title.                                               |
| Confluence.GetPage               | Retrieve a SINGLE page's content by its ID or title.                               |
| Confluence.GetPagesById          | Get the content of MULTIPLE pages by their ID in a single efficient request.       |
| Confluence.ListPages             | Get the content of multiple pages by their ID.                                     |
| Confluence.ListAttachments       | List attachments in a workspace.                                                   |
| Confluence.GetAttachmentsForPage | Get attachments for a page by its ID or title.                                     |
| Confluence.SearchContent         | Search for content in Confluence.                                                  |
| Confluence.GetSpace              | Get the details of a space by its ID or key.                                       |
| Confluence.ListSpaces            | List all spaces sorted by name in ascending order.                                 |
| Confluence.GetSpaceHierarchy     | Retrieve the full hierarchical structure of a Confluence space as a tree structure.|

### Confluence Clients
Confluence has deprecated most of their V1 endpoints, so most of the tools use V2. However, we still need a V1 client to support search. The V1 search API has not been deprecated yet, because there is no V2 equivalent. But we need to be aware in the future for when this deprecation happens.

### Future work
* Content of pages are returned in the Confluence `storage` format. This is the format that Confluence uses to store pages internally.  We should understand the storage format more deeply, and write utility function to transform this format into plain  text.
* Better protections against extremely large pages. I've tested up to 6,000 word pages.
* Tools for blog posts
* Tool for getting the children of a page. (`get_space_hierarchy` will suffice for now)
* Allow for numerical titles

### DIFF
--- a/toolkits/confluence/arcade_confluence/client.py
+++ b/toolkits/confluence/arcade_confluence/client.py
@@ -10,7 +10,6 @@ from arcade_confluence.utils import (
     build_child_url,
     build_hierarchy,
     remove_none_values,
-    split_by_space,
 )
 
 
@@ -74,46 +73,114 @@ class ConfluenceClientV1(ConfluenceClient):
     def __init__(self, token: str):
         super().__init__(token, api_version=ConfluenceAPIVersion.V1)
 
-    def construct_cql(
-        self, terms: list[str] | None, phrases: list[str] | None, enable_fuzzy: bool = False
-    ) -> str:
-        """Construct a CQL query from a list of terms and phrases.
+    def _build_query_cql(self, query: str, enable_fuzzy: bool) -> str:
+        """Build CQL for a single query (term or phrase).
 
         Args:
-            terms: list of single words to search for
-            phrases: list of multiple words to search for
-            enable_fuzzy: enable fuzzy matching to find similar terms (e.g. 'roam' will find 'foam')
+            query: The search query (single word term or multi-word phrase)
+            enable_fuzzy: Whether to enable fuzzy matching for single terms
+
+        Returns:
+            CQL string for the query
+        """
+        query = query.strip()
+        if not query:
+            return ""
+
+        # For phrases (multiple words), don't use fuzzy matching
+        if " " in query:
+            return f'(text ~ "{query}" OR title ~ "{query}" OR space.title ~ "{query}")'
+        else:
+            # For single terms, optionally use fuzzy matching
+            term_suffix = "~" if enable_fuzzy else ""
+            return f'(text ~ "{query}{term_suffix}" OR title ~ "{query}{term_suffix}" OR space.title ~ "{query}{term_suffix}")'  # noqa: E501
+
+    def _build_and_cql(self, queries: list[str], enable_fuzzy: bool) -> str:
+        """Build CQL for queries that must ALL be present (AND logic).
+
+        Args:
+            queries: List of queries that must all be present
+            enable_fuzzy: Whether to enable fuzzy matching for single terms
+
+        Returns:
+            CQL string with AND logic
+        """
+        and_parts = []
+        for query in queries:
+            query_cql = self._build_query_cql(query, enable_fuzzy)
+            if query_cql:
+                and_parts.append(query_cql)
+
+        if not and_parts:
+            return ""
+
+        return f"({' AND '.join(and_parts)})"
+
+    def _build_or_cql(self, queries: list[str], enable_fuzzy: bool) -> str:
+        """Build CQL for queries where ANY can be present (OR logic).
+
+        Args:
+            queries: List of queries where any can be present
+            enable_fuzzy: Whether to enable fuzzy matching for single terms
+
+        Returns:
+            CQL string with OR logic
+        """
+        or_parts = []
+        for query in queries:
+            query_cql = self._build_query_cql(query, enable_fuzzy)
+            if query_cql:
+                or_parts.append(query_cql)
+
+        if not or_parts:
+            return ""
+
+        return f"({' OR '.join(or_parts)})"
+
+    def construct_cql(
+        self,
+        must_contain_all: list[str] | None,
+        can_contain_any: list[str] | None,
+        enable_fuzzy: bool = False,
+    ) -> str:
+        """Construct CQL query with AND/OR logic.
 
         Learn about advanced searching using CQL here: https://developer.atlassian.com/cloud/confluence/advanced-searching-using-cql/
+
+        Args:
+            must_contain_all: Queries that must ALL be present (AND logic)
+            can_contain_any: Queries where ANY can be present (OR logic)
+            enable_fuzzy: Whether to enable fuzzy matching for single terms
+
+        Returns:
+            CQL query string
+
+        Raises:
+            ToolExecutionError: If no search parameters are provided
         """
         cql_parts = []
 
-        # Process single terms
-        if terms:
-            terms = split_by_space(terms)
-            term_queries = []
-            for term in terms:
-                term_suffix = "~" if enable_fuzzy else ""
-                term_queries.append(
-                    f'(text ~ "{term}{term_suffix}" OR title ~ "{term}{term_suffix}" OR space.title ~ "{term}{term_suffix}")'  # noqa: E501
-                )
+        # Handle must_contain_all (AND logic)
+        if must_contain_all:
+            and_cql = self._build_and_cql(must_contain_all, enable_fuzzy)
+            if and_cql:
+                cql_parts.append(and_cql)
 
-            if term_queries:
-                cql_parts.append(f"({' OR '.join(term_queries)})")
+        # Handle can_contain_any (OR logic)
+        if can_contain_any:
+            or_cql = self._build_or_cql(can_contain_any, enable_fuzzy)
+            if or_cql:
+                cql_parts.append(or_cql)
 
-        # Process phrases
-        if phrases:
-            phrase_queries = []
-            for phrase in phrases:
-                phrase_queries.append(
-                    f'(text ~ "{phrase}" OR title ~ "{phrase}" OR space.title ~ "{phrase}")'
-                )
+        # If there's only one part, return it
+        if len(cql_parts) == 1:
+            return cql_parts[0]
 
-            if phrase_queries:
-                cql_parts.append(f"({' OR '.join(phrase_queries)})")
+        # AND the must_contain_all with the can_contain_any
+        if len(cql_parts) > 1:
+            return f"({' AND '.join(cql_parts)})"
 
-        cql = " OR ".join(cql_parts)
-        return cql
+        raise ToolExecutionError(message="At least one search parameter must be provided")
 
     def transform_search_content_response(
         self, response: dict[str, Any]
@@ -358,7 +425,14 @@ class ConfluenceClientV2(ConfluenceClient):
         params = remove_none_values({
             "body-format": content_format.to_api_value(),
         })
-        page = await self.get(f"pages/{page_id}", params=params)
+        try:
+            page = await self.get(f"pages/{page_id}", params=params)
+        except httpx.HTTPStatusError as e:
+            # If the page is not found, return an empty page object
+            if e.response.status_code in [400, 404]:
+                return self.transform_page_response({})
+            raise
+
         return self.transform_page_response(page)
 
     async def get_page_by_title(
@@ -382,7 +456,8 @@ class ConfluenceClientV2(ConfluenceClient):
         response = await self.get("pages", params=params)
         pages = response.get("results", [])
         if not pages:
-            raise ToolExecutionError(message=f"No page found with title: '{page_title}'")
+            # If the page is not found, return an empty page object
+            return self.transform_page_response({})
         return self.transform_page_response(pages[0])
 
     async def get_space_by_id(self, space_id: str) -> dict[str, Any]:
@@ -445,6 +520,10 @@ class ConfluenceClientV2(ConfluenceClient):
         else:
             page = await self.get_page_by_title(page_identifier)
             page_id = page.get("page", {}).get("id")
+
+        if not page_id:
+            raise ToolExecutionError(message=f"No page found with identifier: '{page_identifier}'")
+
         return page_id
 
     async def get_space_id(self, space_identifier: str) -> str:
--- a/toolkits/confluence/arcade_confluence/client.py
+++ b/toolkits/confluence/arcade_confluence/client.py
@@ -1,0 +1,535 @@
+from enum import Enum
+from typing import Any
+from urllib.parse import parse_qs, urlparse
+
+import httpx
+from arcade.sdk.errors import ToolExecutionError
+
+from arcade_confluence.enums import BodyFormat, PageUpdateMode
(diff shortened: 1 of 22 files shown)


toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+    requires_auth=Atlassian(
+        scopes=["read:page:confluence", "write:page:confluence"],
+    )
+)
+async def create_page(
+    context: ToolContext,
+    space_identifier: Annotated[str, "The ID or title of the space to create the page in"],
+    title: Annotated[str, "The title of the page"],
+    content: Annotated[str, "The content of the page. Can be plain text"],
+    parent_id: Annotated[
```
EricGustin, line 22: I'm not supporting title for the parent because this would require support for databases, whiteboards, smart links, etc. This initial version of the toolkit is mainly concerned with spaces and pages.

toolkits/confluence/arcade_confluence/client.py
```diff
@@ -0,0 +1,535 @@
...
+    V2 = "wiki/api/v2"
+
+
+class ConfluenceClient:
+    ACCESSIBLE_RESOURCES_URL = "https://api.atlassian.com/oauth/token/accessible-resources"
+    BASE_URL = "https://api.atlassian.com/ex/confluence"
+
+    def __init__(self, token: str, api_version: ConfluenceAPIVersion):
+        self.token = token
+        self.cloud_id = self._get_cloud_id()
```
byrro, line 28: Nit: ideally, I wouldn't want a class to issue an HTTP request when it's being instantiated. What I did for Jira client was smth like:

```
def __init__(self, cloud_id = None):
        self._cloud_id = cloud_id

@property
def cloud_id(self):
    if not self._cloud_id:
        self._cloud_id = self._get_cloud_id()
    return self._cloud_id
```

toolkits/confluence/arcade_confluence/client.py
```diff
@@ -0,0 +1,535 @@
...
+        using the OAuth2 3LO accessible-resources endpoint.
+
+        For details on why this is necessary, see: https://developer.atlassian.com/cloud/oauth/getting-started/making-calls-to-api
+        """
+        headers = {"Authorization": f"Bearer {self.token}"}
+        resp = httpx.get(self.ACCESSIBLE_RESOURCES_URL, headers=headers)
+        resp.raise_for_status()
+        resp_json = resp.json()
+
+        if len(resp_json) == 0:
```
byrro, line 43: Since the `available-resources` looks like it may return multiple workspaces in some future implementation, I think it'd be safe to add a case for `elif len(resp_json) > 1` here, raising a descriptive error.

toolkits/confluence/arcade_confluence/client.py
```diff
@@ -0,0 +1,535 @@
...
+
+    async def put(self, path: str, **kwargs: Any) -> Any:
+        return await self.request("PUT", path, **kwargs)
+
+
+class ConfluenceClientV1(ConfluenceClient):
+    def __init__(self, token: str):
+        super().__init__(token, api_version=ConfluenceAPIVersion.V1)
+
+    def construct_cql(
```
byrro, line 77: Should constructing CQL strings be a part of the HTTP client? I feel like this belongs to some utilities module.

toolkits/confluence/arcade_confluence/client.py
```diff
@@ -0,0 +1,535 @@
...
+                "status": content.get("status"),
+                "excerpt": result.get("excerpt"),
+                "url": f"{base_url}{result.get('url')}",
+            }
+            transformed_results.append(transformed_result)
+
+        return {"results": transformed_results}
+
+
+class ConfluenceClientV2(ConfluenceClient):
```
byrro, line 142: It looks to me the HTTP client is overloaded with stuff that belongs somewhere else, like transforming request and response payloads according to specific endpoints.

I don't see why we need two Client classes. Just instantiate the same client with a different `api_version`. The stuff that is not HTTP-related should ideally be outside the scope of the HTTP-client.

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+@tool(
+    requires_auth=Atlassian(
+        scopes=["read:page:confluence", "write:page:confluence"],
+    )
+)
+async def create_page(
+    context: ToolContext,
+    space_identifier: Annotated[str, "The ID or title of the space to create the page in"],
+    title: Annotated[str, "The title of the page"],
+    content: Annotated[str, "The content of the page. Can be plain text"],
```
byrro, line 21: Should the annotation be "Only plain text is supported", like in the `update_page_content` tool?

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+        PageUpdateMode,
+        "The mode of update. Defaults to 'append'.",
+    ] = PageUpdateMode.APPEND,
+) -> Annotated[dict, "The page"]:
+    """Update a page's content."""
+    # Get the page to update
+    client = ConfluenceClientV2(context.get_auth_token_or_empty())
+    page_id = await client.get_page_id(page_identifier)
+
+    page = await get_page(context, page_id, content_format=BodyFormat.STORAGE)
```
byrro, line 83: What happens if the page identifier doesn't match any page? Perhaps there should be a `if not page` check before trying to update the page?

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+    requires_auth=Atlassian(
+        scopes=["read:page:confluence", "write:page:confluence"],
+    )
+)
+async def update_page_content(
+    context: ToolContext,
+    page_identifier: Annotated[
+        str, "The ID or title of the page to update. Numerical titles are NOT supported."
+    ],
+    content: Annotated[str, "The content of the page. Only plain text is supported"],
```
byrro, line 72: Is there a reason for not allowing to update the page title as well?

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+    """Retrieve a SINGLE page's content by its ID or title.
+
+    If a title is provided, then the first page with an exact matching title will be returned.
+
+    IMPORTANT: For retrieving MULTIPLE pages, use `get_pages_by_id` instead
+    for a massive performance and efficiency boost. If you call this function multiple times
+    instead of using `get_pages_by_id`, then the universe will explode.
+    """
+    client = ConfluenceClientV2(context.get_auth_token_or_empty())
+    if page_identifier.isdigit():
```
byrro, line 163: I see the annotation warns about numerical titles, but perhaps we should have a more robust implementation here. Instead of checking `isdigit`, why not go ahead and try to first `get_page_by_id` and, if nothing comes up, try `get_page_by_title` next? If none return anything, raise an error or return a 404 message response.

This would allow the tool caller to specify any `page_identifier` without the risk for the tool confusing a numerical title with a page ID.

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+) -> Annotated[dict, "The pages"]:
+    """Get the content of MULTIPLE pages by their ID in a single efficient request.
+
+    IMPORTANT: Always use this function when you need to retrieve content from more than one page,
+    rather than making multiple separate calls to get_page, because this function is significantly
+    more efficient than calling get_page multiple times.
+    """
+    if not page_ids:
+        raise ToolExecutionError(message="The 'page_ids' parameter must be non-empty")
+    page_ids = page_ids[:250]
```
byrro, line 190: As a general rule, I think it could be dangerous to do things like this without warning the tool caller. They'll get back a response thinking that every page_id was considered in the search, but not actually.

I tend to lean on being safe with:

```
if len(page_ids) > 250:
    raise RetryableToolError(f"page_ids must have up 250 items. Got {len(page_ids)}, instead.")
```

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+        "The order of the pages to sort by. Defaults to created-date-newest-to-oldest",
+    ] = PageSortOrder.CREATED_DATE_DESCENDING,
+    limit: Annotated[int, "The maximum number of pages to return. Defaults to 25. Max is 250"] = 25,
+    pagination_token: Annotated[
+        str | None,
+        "The pagination token to use for the next page of results",
+    ] = None,
+) -> Annotated[dict, "The pages"]:
+    """Get the content of multiple pages by their ID"""
+    space_ids = space_ids[:100] if space_ids else None
```
byrro, line 224: Same as the comment about `page_ids` in the previous tool

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+    validate_ids(space_ids)
+    limit = max(1, min(limit, 250))
```
byrro, line 226: I think limit is one exception where capping without warning the tool caller is fine. They'll see the number of results in the response and have a `next_page_token` to paginate anyway, so it's safe.

When we cap the other arguments, like `space_ids`, the tool caller has no way to notice that we actually manipulated their input.

toolkits/confluence/arcade_confluence/tools/search.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@tool(
+    requires_auth=Atlassian(
+        scopes=["search:confluence"],
+    )
+)
+async def search_content(
+    context: ToolContext,
+    terms: Annotated[
```
byrro, line 17: In a previous toolkit I noticed that LLMs sometimes have a hard time understanding they can provide multiple strings in a `list[str]` argument.

Both `terms` and `phrases` could have type `str` and we simply do `terms = terms.split()` internally.

toolkits/confluence/arcade_confluence/tools/search.py
```diff
@@ -0,0 +1,51 @@
...
+    phrases: Annotated[
+        list[str] | None,
+        "A list of groups of words that match content containing all the words in any order. "
+        "For example, ['software developer']",
+    ] = None,
+    enable_fuzzy: Annotated[
+        bool,
+        "Enable fuzzy matching to find similar terms (e.g. 'roam' will find 'foam'). "
+        "Defaults to False",
+    ] = False,
```
byrro, line 31: In general, I think people expect search to be fuzzy by default. It's the experience of most modern search engines and search features in SaaS GUIs

toolkits/confluence/arcade_confluence/client.py
EricGustin: `accessible-resources` returns all workspaces that the token provided can work for. The token that is being used is from the Engine, which can only work for a single workspace (since the oauth flow forces you to select a specific workspace).

toolkits/confluence/arcade_confluence/client.py
EricGustin: Going to keep ALL logic related to the V1 API encapsulated in the same class so that refactoring is easy in the future when the V12 search API is deprecated.

toolkits/confluence/arcade_confluence/client.py
```diff
@@ -0,0 +1,535 @@
...
+            base_url: The base URL for the Confluence space
+
+        Returns:
+            None (modifies root_children in place)
+        """
+        descendent_params = {"limit": 250, "depth": 5}
+
+        for i, child in enumerate(root_children):
+            page_id = child["id"]
+            descendants = await self.get(f"pages/{page_id}/descendants", params=descendent_params)
```
byrro, line 515: Since the number of children could be somewhat large, it'd be interesting to take advantage of async and retrieve multiple at a time here to speed up the process.

Might be desirable to have some sort of timeout control here as well, to prevent the tool from running for too long in the case of large Confluence spaces...

toolkits/confluence/arcade_confluence/tools/page.py
EricGustin: Broke them into two tools because the interface got too complex. `update_page_content` can be an append or a replace operation whereas rename is only a replace operation. Purely preference though

toolkits/confluence/arcade_confluence/client.py
```diff
@@ -0,0 +1,535 @@
...
+            terms = split_by_space(terms)
+            term_queries = []
+            for term in terms:
+                term_suffix = "~" if enable_fuzzy else ""
+                term_queries.append(
+                    f'(text ~ "{term}{term_suffix}" OR title ~ "{term}{term_suffix}" OR space.title ~ "{term}{term_suffix}")'  # noqa: E501
+                )
+
+            if term_queries:
+                cql_parts.append(f"({' OR '.join(term_queries)})")
```
byrro, line 102: While testing the tool, I realized it's OR'ing each term and phrase provided. My intuition is that if I search using `["hello", "world"]`, I'd like to see documents that contain both words.

I think using OR to concatenate will yield poor search results, a low signal to noise ratio. My preference would be to AND all terms and phrases.

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+        str, "The ID or title of the page to rename. Numerical titles are NOT supported."
+    ],
+    title: Annotated[str, "The title of the page"],
+) -> Annotated[dict, "The page"]:
+    """Rename a page by changing its title."""
+    # Get the page to rename
+    client = ConfluenceClientV2(context.get_auth_token_or_empty())
+    page_id = await client.get_page_id(page_identifier)
+
+    page = await get_page(context, page_id, content_format=BodyFormat.STORAGE)
```
byrro, line 123: This raises `TypeError: get_page() got an unexpected keyword argument 'content_format'`

toolkits/confluence/arcade_confluence/tools/page.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+@tool(
+    requires_auth=Atlassian(
+        scopes=["read:page:confluence"],
+    )
+)
+async def get_page(
+    context: ToolContext,
+    page_identifier: Annotated[
```
byrro, line 150: If the `page_identifier` doesn't match any page, the tool raises `httpx.HTTPStatusError: Client error '404 Not Found'`. Perhaps it'd be better to return a `{"page": None, "message": f"Page not found with the identifier '{page_identifier}'"}`.

I noticed this in a few other tools as well. If you end up changing the response, tools that use the changed ones should be adjusted to handle the case of `response["page"] is None`.

PR #407, opened 2025-05-22
Jira Toolkit

byrro:
Jira is quite complicated... and annoying. Every little thing requires a mountain of boilerplate (and auxiliary tools) to make tools easy for LLMs. Hence the unusually big PR.


toolkits/jira/arcade_jira/tools/attachments.py
```diff
@@ -0,0 +1,146 @@
...
+
+    if not filename:
+        raise ToolExecutionError("Must provide a filename.")
+
+    client = JiraClient(context.get_auth_token_or_empty())
+
+    response = await client.post(
+        f"/issue/{issue}/attachments",
+        headers={
+            # "Content-Type": "multipart/form-data",
```
EricGustin: remove?

toolkits/jira/arcade_jira/tools/attachments.py
```diff
@@ -0,0 +1,146 @@
...
+        str,
+        "The name of the file to add as an attachment. The filename should contain the "
+        "file extension (e.g. 'test.txt', 'report.pdf'), but it is not mandatory.",
+    ],
+    file_content_str: Annotated[
+        str | None,
+        "The string content of the file to attach. Use this if the file is a text file. "
+        "Defaults to None.",
+    ] = None,
+    file_content_base64: Annotated[
```
EricGustin, line 27: Can a language model actually provide a valid base64-encoded str reliably?

toolkits/jira/arcade_jira/tools/issues.py
```diff
@@ -0,0 +1,551 @@
...
+        pagination["next_page_token"] = response["nextPageToken"]
+
+    return {
+        "issues": [clean_issue_dict(issue) for issue in response["issues"]],
+        "pagination": pagination,
+    }
+
+
+@tool(requires_auth=Atlassian(scopes=["read:jira-work"]))
+async def search_issues_with_jql(
```
EricGustin, line 389: this tool gives the LLM a lot more room to mess up than our typical tools. We'll need to be very aware of how it is performing over time.

toolkits/jira/arcade_jira/tools/issues.py
```diff
@@ -0,0 +1,551 @@
...
+    labels: Annotated[
+        list[str] | None,
+        "The labels of the issue. Defaults to None (no labels).",
+    ] = None,
+    parent_issue: Annotated[
+        str | None,
+        "The ID or key of the parent issue. Defaults to None (no parent issue). "
+        "Must provide at least one of `parent_issue` or `project` arguments.",
+    ] = None,
+    priority: Annotated[
```
EricGustin, line 489: why not use IssuePriority enum here? Or can priority be customized?

toolkits/jira/arcade_jira/tools/issues.py
```diff
@@ -0,0 +1,551 @@
...
+            "summary": title,
+            "labels": labels,
+            "duedate": due_date,
+            "parent": {"id": parent_data["id"]} if isinstance(parent_data, dict) else None,
+            "project": {"id": project_data["id"]} if isinstance(project_data, dict) else None,
+            "priority": {"id": priority_data["id"]} if isinstance(priority_data, dict) else None,
+            "assignee": {"id": assignee_data["id"]} if isinstance(assignee_data, dict) else None,
+            "reporter": {"id": reporter_data["id"]} if isinstance(reporter_data, dict) else None,
+            "issuetype": {"id": issue_type_data["id"]}
+            if isinstance(issue_type_data, dict)
```
EricGustin: feels like these could all use a `extract_id` helper function if you feel inclined to do so

toolkits/jira/arcade_jira/tools/issues.py
```diff
@@ -0,0 +1,551 @@
...
+            "next_page_token": api_response.get("nextPageToken"),
+        }
+    else:
+        response["pagination"] = {"is_last_page": True}
+
+    return response
+
+
+@tool(requires_auth=Atlassian(scopes=["write:jira-work"]))
+async def create_issue(
```
EricGustin, line 447: this tool does not return a viewable url at `response["issue"]["url"]`.

For example, if it creates this issue: `https://renatoarcade.atlassian.net/jira/software/projects/CRM/list?selectedIssue=CRM-19`, then it returns this url: `https://api.atlassian.com/ex/jira/ea994008-a7d1-4568-8e6e-d28bc5717f20/rest/api/3/issue/10132`

toolkits/jira/arcade_jira/tools/priorities.py
```diff
@@ -0,0 +1,196 @@
...
+    projects = [clean_project_dict(project) for project in api_response["values"]]
+    response = {
+        "projects": projects,
+        "isLast": api_response.get("isLast"),
+    }
+    return add_pagination_to_response(response, projects, limit, offset)
+
+
+@tool(requires_auth=Atlassian(scopes=["manage:jira-configuration"]))
+async def list_priorities_available_to_a_project(
```
EricGustin, line 147: Shouldn't this at the very least return the default priorities (verly low, low, medium, etc)? Or is this something different?

toolkits/jira/arcade_jira/tools/issues.py
```diff
@@ -0,0 +1,551 @@
...
+            f"issue/{issue}",
+            params={"expand": "renderedFields"},
+        )
+    except NotFoundError:
+        return {"error": f"Issue not found with ID/key '{issue}'."}
+    return {"issue": clean_issue_dict(response)}
+
+
+@tool(requires_auth=Atlassian(scopes=["read:jira-work"]))
+async def get_issues_without_id(
```
EricGustin, line 124: this tool needs to protect against an empty query, otherwise the API returns bad request `Unbounded JQL queries are not allowed here. Please add a search restriction to your query`

toolkits/jira/arcade_jira/tools/priorities.py
```diff
@@ -0,0 +1,196 @@
...
+    except asyncio.TimeoutError:
+        return {"error": f"The operation timed out after {JIRA_API_REQUEST_TIMEOUT} seconds."}
+    except JiraToolExecutionError as error:
+        return {"error": error.message}
+
+
+@tool(requires_auth=Atlassian(scopes=["manage:jira-configuration"]))
+async def list_priorities_available_to_an_issue(
+    context: ToolContext,
+    issue: Annotated[str, "The ID, key or name of the issue to retrieve priorities for."],
```
EricGustin: annotation says it can be ID, key, or name. But the tool uses `get_issue_by_id` which looks to be only for ID & key; not name

EricGustin (commented):
Nice, very comprehensive toolkit. Most of my comments are nits.

A couple improvements that are nice to haves:
1. The URLs returned by the tools don't seem to work. It would be nice if there was some internal function that constructs urls that end-users could click on that takes them to their issue, comment, project, etc.
2. Very simple queries like 'Show me 1 issue' fail (400 response) because it needs to be tied to a project. Perhaps if there's no project provided we list available projects and get the first issue we can find? My reasoning behind this is that typically simple queries like this could be the reason why someone decides to never use the tools again & miss out on all of the cool features you've built.

toolkits/jira/arcade_jira/tools/attachments.py
byrro: I trust even smaller models with writing Python code to `base64.b64encode` a binary. If the model is capable of executing this code and getting the results, I trust this can work very well.

The thing is: if we don't support such an argument, it's impossible to attach any binary (PDF, images, DOC, etc), which renders the tool practically useless. The tool checks whether the string provided is a valid base64 (raising a descriptive error if not) before proceeding with the attach request.

toolkits/jira/arcade_jira/tools/issues.py
byrro: Yep, I agree that an arg accepting a full query str is not ideal. JQL is read-only, which eliminates a good chunk of hairy issues. Jira is so configurable and complex that we can't parameterize a lot of stuff without some serious gymnastics in our tools.

In the evals, every time the LLM is asked to find an issue without providing a JQL, it always chooses the `GetIssuesWithoutId` tool (which is parameterized). The `SearchIssuesWithJql` tool is only selected when the prompt explicitly gives a JQL.

So I guess it's relatively safe to keep it, and it's going to give freedom to retrieve data that our parameterized tools can't.

toolkits/jira/arcade_jira/tools/issues.py
byrro: Ah the `IssuePriority` enum is from an early age when I was still in wonderland... before I got to know Jira 🤣 

We can't parameterize priorities, it's configurable and it varies even in the same account, depending on the project. Had to write the tools in `arcade_jira.priorities` to take care of that.

To abstract the mess, every tool that takes a priority arg will accept a priority name. When the user mentions "get all high priority issues", the LLM can provide "high" and the tool will figure the correct priority ID to filter.

byrro: Just removed the `IssuePriority` enum, btw, it's not used anymore...

toolkits/jira/arcade_jira/tools/issues.py
byrro: yep, good idea, looks much better now

toolkits/jira/arcade_jira/tools/priorities.py
byrro: @EricGustin What does it return to you? For me, it does return the default ones: lowest, low, medium, high, highest

toolkits/jira/arcade_jira/tools/issues.py
byrro: good point, I just realized those URLs are for API endpoints, not for their GUI.. I'll adapt them to be viewable from a browser.

byrro:
> Very simple queries like 'Show me 1 issue' fail (400 response) because it needs to be tied to a project. Perhaps if there's no project provided we list available projects and get the first issue we can find?

@EricGustin That's a very good point. I don't feel comfortable pre-selecting the first project if the user has multiple. Creates the potential for tricking the user into interacting with the wrong issues unknowingly.

But if there's only one project, we can proactively select it. In case of multiple projects, we can raise an error listing the available projects. Would that be a good approach?

toolkits/jira/arcade_jira/tools/priorities.py
EricGustin: It was returning nothing for the project 'Arcade'

toolkits/jira/arcade_jira/tools/transitions.py
```diff
@@ -6,6 +6,45 @@
...
+                f"No transition found for the issue '{issue}' with ID '{transition_id}'. "
+                "To get all transitions available for the issue, use the "
+                f"`Jira.{get_transitions_available_for_issue.__tool_name__}` tool."
+            ),
+        }
+
+    if len(transitions) == 1:
+        return {"transition": transitions[0]}
+
+    return {
```
EricGustin, line 42: I'm interested in why we choose to return a dict with an error message instead of raising an error here. Was there an observation that made this the ideal route?

byrro: The idea was to respond with more context for the LLM to know what to do next. Most tools abstract away a lot of Jira's complexity. I wanted to also keep it simple for LLMs/callers when something goes wrong too.

Like in the case you highlighted, it includes a list of available transitions, so that the LLM can ask/pick one and call again correctly. Other times, like a few lines above, it'll indicate another tool that can be used to get the information needed for a correct call.

I could shove all this data into an exception message too or use the additional prompt in a retryable exception, but this kind of exception is at risk of deprecation, right?...

toolkits/jira/arcade_jira/client.py
```diff
@@ -38,16 +38,18 @@ async def get_cloud_id(self) -> str:
...
+                await cache.async_set_cloud_name(self.auth_token, cloud["name"])
 
         return self._cloud_id
 
     async def _build_url(self, endpoint: str) -> str:
         cloud_id = await self.get_cloud_id()
         return f"{self.base_url}/{cloud_id}/rest/api/{self.api_version}/{endpoint.lstrip('/')}"
 
-    async def _get_cloud_id_from_available_resources(self) -> str:
+    async def _get_cloud_id_from_available_resources(self) -> dict[str, Any]:
```
EricGustin: nit: after you changed this from returning a just the ID to returning a dict, maybe rename it since it isn't just returning an ID

toolkits/jira/arcade_jira/tools/issues.py
```diff
@@ -394,6 +598,8 @@ async def add_labels_to_issue(
     issue_data = await get_issue_by_id(context, issue)
     if issue_data.get("error"):
         return cast(dict, issue_data)
+
+    labels = [label.replace(" ", "_") for label in labels]
```
EricGustin: nit/opinion: consider `strip()` before `replace(" ", "_")`

toolkits/jira/arcade_jira/utils.py
```diff
@@ -38,7 +39,7 @@ def is_valid_date_string(date_string: str) -> bool:
 
 
 def quote(v: str) -> str:
-    return f'"{v.replace('"', '\\"')}"'
+    return repr(v)  #  f'"{v.replace('"', '\\"')}"'
```
EricGustin: What is the intent of this function?

Function is named `quote` and _repr() is a built-in function in Python used to obtain the string representation of an object_

toolkits/jira/arcade_jira/tools/issues.py
```diff
@@ -0,0 +1,780 @@
...
+    if issue_data.get("error"):
+        return cast(dict, issue_data)
+
+    project = issue_data["issue"]["project"]["id"]
+
+    error, _, issue_type_data, priority_data, parent_issue_data = await validate_issue_args(
+        context, due_date, project, issue_type, priority, parent_issue
+    )
+    if error:
+        return cast(dict, error)
```
EricGustin, line 756: how golang of you

byrro:
> * The URLs returned by the tools don't seem to work. It would be nice if there was some internal function that constructs urls that end-users could click on that takes them to their issue, comment, project, etc.

this was addressed by [this commit](https://github.com/ArcadeAI/arcade-ai/pull/407/commits/b23a09b06b865c5194092d0e41144cdc077bd037). it's really useful to have browser-accessible links while chatting

byrro:
> * Very simple queries like 'Show me 1 issue' fail (400 response) because it needs to be tied to a project

addressed by [this commit](https://github.com/ArcadeAI/arcade-ai/pull/407/commits/add7c1ce25cebf4fc03a735ed333e840e0f63cad): makes `project` non-required arg and tries to determine a project when not specified, or returns a list of available projects when not possible to determine

toolkits/jira/arcade_jira/utils.py
byrro: good catch, this change should have been reverted before that commit

EricGustin (approved):
LGTM

PR #408, opened 2025-05-23
Get Salesforce org subdomain from secret

byrro:


EricGustin:
Do we want to add the secret as a requirement for the tools?

sdreyer:
The engine side of this is in. Assuming this needs to be added?

sdreyer (approved):

EricGustin:
@byrro closing this PR, but we should still add Subdomain as a secret that the tools requires

PR #414, opened 2025-05-30
Toolkit docs generator command for Arcade CLI

byrro:
This PR adds a new command to Arcade CLI for automated toolkit docs generation. E.g. `arcade generate-toolkit-docs --toolkit-name Jira --docs-section productivity`.

It retrieves a list of tools from the Engine and auto-generates a full toolkit documentation MDX page with headers, description, table of contents, list of tools, as well as tool-call example scripts in Python and JavaScript for each tool.

Use the `--docs-root-dir` parameter to specify the path to your local docs repo.

### DIFF
--- a/.github/workflows/test-toolkits.yml
+++ b/.github/workflows/test-toolkits.yml
@@ -11,25 +11,31 @@ jobs:
   setup:
     runs-on: ubuntu-latest
     outputs:
-      tool_matrix: ${{ steps.get-toolkits.outputs.toolkits }}
+      toolkits_with_gha_secrets: ${{ steps.load_toolkits.outputs.toolkits_with_gha_secrets }}
+      toolkits_without_gha_secrets: ${{ steps.load_toolkits.outputs.toolkits_without_gha_secrets }}
     steps:
       - name: Check out
         uses: actions/checkout@v4
 
-      - name: Get toolkits
-        id: get-toolkits
+      - name: determine toolkits with and without GHA secrets
+        id: load_toolkits
         run: |
           # Find all directories in toolkits/ that have a pyproject.toml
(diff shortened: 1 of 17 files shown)


arcade/arcade/cli/main.py
```diff
@@ -702,6 +703,59 @@ def dashboard(
...
+        "--docs-section",
+        "-s",
+        help=(
+            "The section of the docs to generate documentation for. E.g. 'productivity', 'sales'. "
+            "Defaults to an empty string (generate the docs in the root of /pages/toolkits)"
+        ),
+        show_default=True,
+    ),
+    docs_root_dir: str = typer.Option(
+        "~/arcade/Team/docs",
```
evantahler, line 722: That's not a great default for something public - only us employees will have that.  How about no default?

evantahler: or maybe CWD?

arcade/arcade/cli/main.py
```diff
@@ -702,6 +703,59 @@ def dashboard(
...
+        "--docs-root-dir",
+        "-r",
+        help="The path to the documentation root directory.",
+        show_default=True,
+    ),
+    engine_base_url: str = typer.Option(
+        "http://127.0.0.1:9099",
```
evantahler, line 729: [nit] I understand you are re-using the swagger endpoint to get the info you need, but it seems a little odd to have to run the service to get the annotation info.  Would there be a way to pass a directory path to the tool instead, and then look for methods with the `@tool` annotation?  I think could be possible with `inspect` and `ast`

evantahler (changes requested):
Just the default path should change.  I like that there are no new imports needed!

arcade/arcade/cli/main.py
byrro: the worker certainly has some method to collect this data, but I was too lazy to look for it 🙈

it was faster to get it from the engine.. I'll look for a simple way to get it without the need for running the engine

byrro: @evantahler it now retrieves tools from the local Python environment, instead of a running Engine

PR #415, opened 2025-06-04
Prototype error handling with Slack example

byrro:
This is a prototype proposal for better error handling in our tools and SDKs. In principle, I don't think this woud require any changes to the Engine, but would have to double check with Nate/Wils.

I'll be adding some comments in the code to explain the reasoning.

### DIFF
--- a/arcade/arcade/core/errors.py
+++ b/arcade/arcade/core/errors.py
@@ -1,6 +1,8 @@
 import traceback
 from typing import Optional
 
+from arcade.core.schema import HttpResponse
+
 
 class ToolkitError(Exception):
     """
@@ -38,15 +40,26 @@ class ToolDefinitionError(ToolError):
 
 
 class ToolRuntimeError(RuntimeError):
+    _can_retry: bool = False
+
     def __init__(
         self,
         message: str,
         developer_message: Optional[str] = None,
+        additional_prompt_content: Optional[str] = None,
+        retry_after_ms: Optional[int] = None,
     ):
         super().__init__(message)
         self.message = message
         self.developer_message = developer_message
+        self.additional_prompt_content = additional_prompt_content
+        self.retry_after_ms = retry_after_ms
 
+    @property
+    def can_retry(self) -> bool:
+        return self._can_retry
+
+    @property
     def traceback_info(self) -> str | None:
         # return the traceback information of the parent exception
         if self.__cause__:
@@ -62,21 +75,7 @@ class ToolExecutionError(ToolRuntimeError):
     pass
 
 
-class RetryableToolError(ToolExecutionError):
-    """
-    Raised when a tool error is retryable.
-    """
-
-    def __init__(
-        self,
-        message: str,
-        developer_message: Optional[str] = None,
-        additional_prompt_content: Optional[str] = None,
-        retry_after_ms: Optional[int] = None,
-    ):
-        super().__init__(message, developer_message)
-        self.additional_prompt_content = additional_prompt_content
-        self.retry_after_ms = retry_after_ms
+# ------  serialization errors ------
 
 
 class ToolSerializationError(ToolRuntimeError):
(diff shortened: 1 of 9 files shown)


arcade/arcade/core/errors.py
```diff
@@ -38,15 +40,26 @@ class ToolDefinitionError(ToolError):
 
 
 class ToolRuntimeError(RuntimeError):
+    _can_retry: bool = False
```
byrro, line 43: I think `can_retry` should be a property of the error, so we can use it in different types of errors when needed.

arcade/arcade/core/errors.py
```diff
@@ -38,15 +40,26 @@ class ToolDefinitionError(ToolError):
...
+
     def __init__(
         self,
         message: str,
         developer_message: Optional[str] = None,
+        additional_prompt_content: Optional[str] = None,
+        retry_after_ms: Optional[int] = None,
```
byrro, line 50: Since `can_retry` is a property of any error, we need to accept `additional_prompt` and `retry_after_ms` when instantiating any class.

arcade/arcade/core/errors.py
```diff
@@ -101,3 +110,40 @@ class ToolOutputError(ToolSerializationError):
...
+
+# ------  retry-able errors ------
+
+
+class RetryableToolError(ToolExecutionError):
+    """
+    Raised when a tool error is retryable.
+    """
+
+    _can_retry: bool = True
```
byrro, line 123: To make an error retryable, we simply set `_can_retry` to `True`.

This makes it easier to handle tool execution errors: the developer simply checks for `error.can_retry`, instead of having to differentiate by the type of exception.

arcade/arcade/core/output.py
```diff
@@ -11,8 +12,8 @@ class ToolOutputFactory:
     Singleton pattern for unified return method from tools.
     """
 
+    @staticmethod
```
byrro, line 15: Made these two methods `staticmethod`, so we don't have to instantiate a global singleton

arcade/arcade/core/output.py
```diff
@@ -21,44 +22,22 @@ def success(
...
-                can_retry=False,
-                traceback_info=traceback_info,
+                name=error.__class__.__name__,
+                message=error.message,
+                developer_message=error.developer_message,
+                can_retry=error.can_retry,
+                additional_prompt_content=error.additional_prompt_content,
+                retry_after_ms=error.retry_after_ms,
+                traceback_info=error.traceback_info,
+                http_response=getattr(error, "http_response", None),
```
byrro: If I was a developer (Arcade customer) executing an Arcade tool and the downstream API request failed, I would like to see the HTTP response. In many cases, the error will be connected to a misconfig in my oauth app, or smth else that I need to act upon.

Currently, most (if not all) the data returned by the third-party API is lost, which makes debugging much harder for developers.

Someone will raise a concern about security, data privacy, or whatever. But think about it: the developer can get the auth-token and call the API themselves, right? There's nothing stopping them from getting access to the full API response. Why make it harder for them to get it in the context of a tool call? I can see no logical reason, tbh...

arcade/arcade/core/errors.py
```diff
@@ -38,15 +40,36 @@ class ToolDefinitionError(ToolError):
...
         super().__init__(message)
+
+        if (
+            not self._can_retry
+            and retry_after_ms is not None
+            or additional_prompt_content is not None
+        ):
+            raise AttributeError(
+                "retry_after_ms and additional_prompt_content are only allowed for errors where `_can_retry` is True"
+            )
```
evantahler, line 61: Can this check be done with stronger types instead?  A runtime error, when we are already in the middle of throwing an error, seems tricky.  This would be seen by a user, not a developer

arcade/arcade/core/errors.py
```diff
@@ -101,3 +110,40 @@ class ToolOutputError(ToolSerializationError):
...
+
+class RetryableToolError(ToolExecutionError):
+    """
+    Raised when a tool error is retryable.
+    """
+
+    _can_retry: bool = True
+
+
+class ThirdPartyApiError(ToolExecutionError):
```
evantahler, line 126: It's a nuance, but I don't particularly like the term "Third Party" - it puts the API providers in a 'lower class' than Arcade, but our users/developers might not be thinking of it that way.  It's a value judgment. How about "Remote" or "Upstream"?

arcade/arcade/core/errors.py
```diff
@@ -101,3 +110,40 @@ class ToolOutputError(ToolSerializationError):
...
+        super().__init__(*args, **kwargs)
+        self.http_response = http_response.model_dump()
+
+
+class ThirdPartyApiServerError(ThirdPartyApiError, RetryableToolError):
+    """
+    Raised when there is a 500 Server Error in a downstream request to a third-party service.
+    """
+
+    pass
```
evantahler, line 141: I think there's also a class of `ThirdPartyApiServerError` that isn't retryable, e.g. "user not found".   I think that implies that this is `RetryableThirdPartyApiServerError` and there's a missing `FatalThirdPartyApiServerError` or something like that

arcade/arcade/core/output.py
evantahler: How about making this an option that can be configured on the `@tool` decorator, or a global setting on the Arcade TDK?  This way in development, folks can see the errors, but in production, if they don't want to, we won't return something that might be sensitive?

toolkits/slack/arcade_slack/utils.py
```diff
@@ -444,3 +450,44 @@ def convert_relative_datetime_to_unix_timestamp(
...
+        error = ThirdPartyApiRateLimitError(
+            http_response=http_response,
+            error_message=error_message,
+            retry_after_ms=retry_after_ms,
+        )
+    else:
+        error = ThirdPartyApiError(
+            http_response=http_response,
+            error_message=error_message,
+        )
```
evantahler, line 480: cc @EricGustin - I'll bet we can come up with a utility that makes this part part simple too, including a common form for `get_retry_after_ms`

PR #420, opened 2025-06-09
Add GitLab OAuth provider

byrro:


byrro:
closing until we figure the best way to support multi-app with a single toolkit

PR #429, opened 2025-06-12
Add PyPI release workflow

EricGustin:
This is the first of a few PRs. Deploy to staging will fail until we have `arcade-core`, `arcade-serve`, and `arcade-ai` released to PyPI. This PR will release `arcade-core` to PyPI.

### PR Description
* Adds workflow that checks for changes in any pyproject.toml, and if its version has changed, then tests, builds wheel, then publishes to PyPI
* Updates the Dockerfile for our new structure
* Updates porter yamls
* Updates `make full-dist`
* Removes a couple unused workflows

Check out https://github.com/ArcadeAI/arcade-ai/actions/runs/15622059209 to see how the new workflow works (note that it failed publishing to PyPI on purpose)


.github/workflows/porter_app_worker_4828.yml
```diff
@@ -18,18 +18,9 @@ jobs:
...
-              with:
-                version: 1.8.5
-
             - name: Build Dist
               run: make full-dist
 
-            - name: Export Requirements
-              run: |
-                cd arcade && poetry export --output ../dist/requirements.txt
-
```
EricGustin, line 32: I don't think porter needs requirements.txt?

Makefile
```diff
@@ -1,7 +1,3 @@
-CLI_VERSION ?= "2.0.0"
-TDK_VERSION ?= "2.0.0"
-SERVE_VERSION ?= "2.0.0"
-CORE_VERSION ?= "2.0.0"
```
EricGustin, line 4: pyproject.toml should be source of truth, not this

Makefile
```diff
@@ -101,28 +97,6 @@ coverage: ## Generate coverage report
 	@echo "Generating coverage report"
 	@uv run coverage html
 
-.PHONY: set-version
```
EricGustin, line 104: not needed since pyproject.toml version is source of truth

Makefile
```diff
@@ -219,19 +193,18 @@ publish-ghcr: ## Publish to the GHCR
 full-dist: clean-dist ## Build all projects and copy wheels to ./dist
```
EricGustin, line 193: output of `make full-dist`:
image

.github/workflows/porter_app_worker_4828.yml
sdreyer: we need a requirements.txt in the ./dist/ directory. Porter does need that but unsure if theres a better way to get it in there

.github/workflows/porter_app_worker_4879.yml
```diff
@@ -18,18 +18,9 @@ jobs:
...
-              uses: snok/install-poetry@v1
-              with:
-                version: 1.8.5
-
             - name: Build Dist
               run: make full-dist
 
-            - name: Export Requirements
-              run: |
-                cd arcade && poetry export --output ../dist/requirements.txt
```
sdreyer, line 31: same as above

.github/workflows/release-on-version-change.yml
```diff
@@ -0,0 +1,173 @@
...
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46
+        with:
+          files: |
+            **/pyproject.toml
+
+      - name: Check for version changes or new packages
+        id: check-versions
+        if: steps.changed-files.outputs.any_changed == 'true'
+        run: |
+          echo "Changed pyproject.toml files:"
```
sdreyer: may want to put this in a script or makefile so its easier to test

Makefile
sdreyer: agreed

.github/workflows/porter_app_worker_4828.yml
sdreyer: you can try `docker build -f <dockerfilename> .` locally to see if it still needs it

EricGustin: ok so the question now is which package does it need the requirements.txt for? `arcade-serve`? @sdreyer

.github/workflows/release-on-version-change.yml
EricGustin: turned into a script

.github/workflows/porter_app_worker_4828.yml
sdreyer: well if we just install the whl file, do we still need reqs?

EricGustin: updated Dockerfile - requirements.txt no longer needed

.github/scripts/get_toolkits.sh
```diff
@@ -1,12 +0,0 @@
-#!/bin/bash
-
-# Get all directories and format as JSON array
-echo -n '['
```
EricGustin, line 4: I don't think this was being used

.github/workflows/check-toolkits.yml
```diff
@@ -1,68 +0,0 @@
-name: Check Toolkits
```
EricGustin, line 1: no longer used

.github/workflows/promote.yml
```diff
@@ -8,6 +8,15 @@ on:
         type: string
         required: false
         default: ''
+      worker_container_increment:
```
EricGustin, line 11: only adding the option right now. TODO: Implement logic that actually does the version increment logic

.github/workflows/publish-toolkit.yml
```diff
@@ -1,93 +0,0 @@
-name: Publish Toolkit
```
EricGustin, line 1: no longer used

.github/workflows/promote.yml
sdreyer: Thats fine

libs/arcade-core/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "arcade-core"
-version = "2.0.0rc1"
+version = "2.0.0"
```
sdreyer, line 3: we going straight out with 2.0.0, no rc?

sdreyer (approved):

libs/arcade-core/pyproject.toml
EricGustin: only for core, tdk, and serve. `arcade-ai` will be an rc

PR #431, opened 2025-06-13
Temp version change for core

EricGustin:
Forgot to set up on the PyPI side. Temporarily changing back to 1.0.0, so that I can trigger a release when going back to 2.0.0


EricGustin:
not needed, just reran workflow instead

PR #437, opened 2025-06-13
Small fixes before arcade-ai v2.0.0

EricGustin:
As I continue to test the pre-release I will place fixes in this PR

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,9 +33,8 @@ EXPOSE $PORT
 # List wheel files for debugging purposes
 RUN ls -la /app/dist/
 
-# Install the core Arcade packages (each has its own version)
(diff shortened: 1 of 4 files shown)


evantahler (approved):

docker/Dockerfile
```diff
@@ -33,9 +33,8 @@ EXPOSE $PORT
 # List wheel files for debugging purposes
 RUN ls -la /app/dist/
 
-# Install the core Arcade packages (each has its own version)
+# Install the CLI and worker package
 RUN python -m pip install \
-    /app/dist/arcade_core-*.whl \
```
EricGustin, line 38: since arcade-core is a transitive dependency through arcade-serve, I don't think we need to explicitly install it here.

PR #439, opened 2025-06-14
Update release containers GH Action

EricGustin:
1. Patch, Minor, or Major is selected during workflow dispatch (defaults to patch)
2. Get most recent release version
3. Calculate new version based on findings from step 2
4. Build wheels
5. Build worker and base-worker
6. Push to GHCR
7. Create new release with release notes

Also, I removed everything related to ECR


docker/Dockerfile
```diff
@@ -33,9 +33,8 @@ EXPOSE $PORT
 # List wheel files for debugging purposes
 RUN ls -la /app/dist/
 
-# Install the core Arcade packages (each has its own version)
+# Install the worker and CLI package
 RUN python -m pip install \
-    /app/dist/arcade_core-*.whl \
```
EricGustin: I don't think this is actually needed since arcade_core is already a transitive dependency through arcade_serve

.github/workflows/promote.yml
```diff
@@ -170,3 +170,202 @@ jobs:
           echo "### Branches" >> $GITHUB_STEP_SUMMARY
           echo "- **Source:** main" >> $GITHUB_STEP_SUMMARY
           echo "- **Target:** production" >> $GITHUB_STEP_SUMMARY
+
+  calculate-version:
+    runs-on: ubuntu-latest
+    needs: promote
```
EricGustin, line 176: Technically we could release containers perfectly fine without first needing "promote" to succeed, but I think released containers should reflect what is on prod, so I'm adding this prerequisite in case "promote" fails. I'm open to the opposition.

.github/workflows/promote.yml
```diff
@@ -170,3 +170,202 @@ jobs:
...
+          echo "" >> release_notes.md
+          echo "## What's Changed" >> release_notes.md
+          echo "" >> release_notes.md
+
+          # Get all commits in production branch since it was created
+          echo "This release includes all changes promoted to the production branch." >> release_notes.md
+          echo "" >> release_notes.md
+          echo "### Recent commits:" >> release_notes.md
+          # Show last 20 commits in production branch
+          git log --oneline -20 --pretty=format:"- %s" >> release_notes.md
```
EricGustin, line 346: Not confident about this line.

sdreyer (approved):
LGTM, we can clean it up later if needed

PR #449, opened 2025-06-17
Update check unauthorized toolkit updates workflow

EricGustin:
Forgot to add the change for the target versions too


sdserranog (approved):
:shipit:

PR #453, opened 2025-06-20
Slack tools to get user by email / username & performance improvements

byrro:
Adds four new tools to retrieve a user: by username and email.

- `get_user_by_username`
- `get_multiple_users_by_username`
- `get_user_by_email`
- `get_multiple_users_by_email`

Retrieving by email is the most efficient way. Retrieving by username still requires scanning the list of users, but it does so more efficiently now (especially the `get_multiple_users_by_username` tool is much more efficient than previous implementations).

### DIFF
--- a/toolkits/slack/arcade_slack/models.py
+++ b/toolkits/slack/arcade_slack/models.py
@@ -1,5 +1,7 @@
+from abc import ABC, abstractmethod
+from contextlib import suppress
 from enum import Enum
-from typing import Literal, TypedDict
+from typing import Any, Literal, TypedDict
 
 from arcade_slack.custom_types import (
     SlackOffsetSecondsFromUTC,
@@ -204,3 +206,57 @@ class SlackConversationsToolResponse(TypedDict, total=True):
 
     conversations: list[ConversationMetadata]
     next_cursor: SlackPaginationNextCursor | None
+
+
+class PaginationSentinel(ABC):
+    """Base class for pagination sentinel classes."""
+
+    def __init__(self, **kwargs: Any) -> None:
+        self.kwargs = kwargs
+
+    @abstractmethod
+    def __call__(self, last_result: Any) -> bool:
+        """Determine if the pagination should stop."""
+        raise NotImplementedError
+
+
+class FindUserByUsernameSentinel(PaginationSentinel):
+    """Sentinel class for finding a user by username."""
+
+    def __call__(self, last_result: Any) -> bool:
+        for user in last_result:
+            if not isinstance(user.get("name"), str):
+                continue
+            if user.get("name").casefold() == self.kwargs["username"].casefold():
+                return True
+        return False
+
+
+class FindMultipleUsersByUsernameSentinel(PaginationSentinel):
+    """Sentinel class for finding multiple users by username."""
+
+    def __init__(self, usernames: list[str]) -> None:
+        if not usernames:
+            raise ValueError("usernames must be a non-empty list of strings")
+        super().__init__(usernames=usernames)
+        self.usernames_pending = {username.casefold() for username in usernames}
+
+    def flag_username_found(self, username: str) -> None:
+        with suppress(KeyError):
+            self.usernames_pending.remove(username.casefold())
+
+    def all_usernames_found(self) -> bool:
+        return not self.usernames_pending
+
+    def __call__(self, last_result: Any) -> bool:
+        if not self.usernames_pending:
+            return True
+        for user in last_result:
+            username = user.get("name")
+            if not isinstance(username, str):
+                continue
+            if username.casefold() in self.usernames_pending:
+                self.flag_username_found(username)
+                if self.all_usernames_found():
+                    return True
+        return False
--- a/toolkits/slack/arcade_slack/tools/chat.py
+++ b/toolkits/slack/arcade_slack/tools/chat.py
@@ -1,6 +1,6 @@
 import asyncio
 from datetime import datetime, timezone
-from typing import Annotated, cast
+from typing import Annotated, Any, cast
 
 from arcade_tdk import ToolContext, tool
 from arcade_tdk.auth import Slack
@@ -9,15 +9,12 @@ from slack_sdk.errors import SlackApiError
 from slack_sdk.web.async_client import AsyncWebClient
 
 from arcade_slack.constants import MAX_PAGINATION_TIMEOUT_SECONDS
-from arcade_slack.exceptions import (
-    ItemNotFoundError,
-    UsernameNotFoundError,
-)
+from arcade_slack.exceptions import ItemNotFoundError
 from arcade_slack.models import (
     ConversationType,
     SlackUserList,
 )
-from arcade_slack.tools.users import get_user_info_by_id, list_users
+from arcade_slack.tools.users import get_user_info_by_id
 from arcade_slack.utils import (
     async_paginate,
     convert_conversation_type_to_slack_name,
@@ -26,7 +23,6 @@ from arcade_slack.utils import (
     enrich_message_datetime,
     extract_conversation_metadata,
     format_users,
-    get_user_by_username,
     retrieve_conversations_by_user_ids,
 )
 
@@ -324,7 +320,7 @@ async def get_messages_in_conversation_by_id(
     )
     slackClient = AsyncWebClient(token=token)
 
-    datetime_args = {}
+    datetime_args: dict[str, Any] = {}
     if oldest_timestamp:
         datetime_args["oldest"] = oldest_timestamp
     if latest_timestamp:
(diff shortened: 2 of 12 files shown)


toolkits/slack/arcade_slack/models.py
```diff
@@ -204,3 +206,57 @@ class SlackConversationsToolResponse(TypedDict, total=True):
...
+            raise ValueError("usernames must be a non-empty list of strings")
+        super().__init__(usernames=usernames)
+        self.usernames_pending = {username.casefold() for username in usernames}
+
+    def flag_username_found(self, username: str) -> None:
+        with suppress(KeyError):
+            self.usernames_pending.remove(username.casefold())
+
+    def all_usernames_found(self) -> bool:
+        return not self.usernames_pending
```
EricGustin: nit: make these private methods?

toolkits/slack/arcade_slack/models.py
```diff
@@ -204,3 +206,57 @@ class SlackConversationsToolResponse(TypedDict, total=True):
...
+
+    def __call__(self, last_result: Any) -> bool:
+        if not self.usernames_pending:
+            return True
+        for user in last_result:
```
EricGustin, line 269: possibly we could reuse `FindUserByUsernameSentinel` here

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -474,7 +476,12 @@ async def get_messages_in_direct_message_conversation_by_username(
         "when there are additional messages to retrieve."
     ),
 ]:
-    """Get the messages in a direct conversation by the user's name.
+    """Get the messages in a direct conversation by the username or email.
+
+    IF YOU HAVE A USERNAME OR EMAIL ADDRESS, DO NOT CALL THE `Slack.GetUserByUsername`
+    AND/OR `Slack.GetUserByEmail` TOOLS FIRST. PASS THE USERNAME OR EMAIL ADDRESS DIRECTLY TO THIS
+    TOOL. IF YOU CALL ANOTHER TOOL FIRST UNNECESSARILY, YOU WILL GENERATE TOO MUCH CO2 AND
```
EricGustin: the consequences are immense

toolkits/slack/arcade_slack/utils.py
```diff
@@ -444,3 +443,94 @@ def convert_relative_datetime_to_unix_timestamp(
...
+    elif user.get("email"):
+        data["email"] = user["email"]
+    return data
+
+
+def short_human_users_info(users: list[dict]) -> list[dict[str, str | None]]:
+    return [short_user_info(user) for user in users if not user.get("is_bot")]
+
+
+def is_valid_email(email: str) -> bool:
```
EricGustin, line 412: using a regex would probably be more robust

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -705,50 +727,58 @@ async def get_direct_message_conversation_metadata_by_username(
...
     This tool does not return the messages in a conversation. To get the messages, use the
-    `get_messages_in_direct_message_conversation_by_username` tool."""
-    try:
-        token = (
-            context.authorization.token
-            if context.authorization and context.authorization.token
-            else ""
-        )
-        slack_client = AsyncWebClient(token=token)
+    `get_messages_in_direct_message_conversation_by_username` tool.
```
EricGustin: should this docstring be changed to `get_messages_in_direct_message_conversation_by_user`?

toolkits/slack/arcade_slack/utils.py
EricGustin: also, I think there are actually 3 cases, not 2.

1. Is a valid username
2. Is a valid email
3. Is not a valid username or valid email

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -705,50 +727,58 @@ async def get_direct_message_conversation_metadata_by_username(
...
+    current_user, other_user = await asyncio.gather(
+        slack_client.auth_test(), get_user(context, username_or_email)
+    )
 
+    conversations_found = await retrieve_conversations_by_user_ids(
+        list_conversations_func=list_conversations_metadata,
+        get_members_in_conversation_func=get_members_in_conversation_by_id,
+        context=context,
+        conversation_types=[ConversationType.DIRECT_MESSAGE],
+        user_ids=[current_user["user_id"], other_user["user"]["id"]],
```
EricGustin: before doing `current_user["user_id"]` we should ensure that `current_user["ok"]` is true otherwise a KeyError will be thrown

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -757,58 +787,49 @@ async def get_multi_person_dm_conversation_metadata_by_usernames(
...
-    except UsernameNotFoundError as e:
+    if not conversations_found:
         raise RetryableToolError(
-            f"Username '{e.username_not_found}' not found",
-            developer_message=f"User with username '{e.username_not_found}' not found.",
-            additional_prompt_content=f"Available users: {e.usernames_found}",
-            retry_after_ms=500,
+            "Conversation not found with the usernames provided",
+            developer_message="Conversation not found with the usernames provided",
+            retry_after_ms=100,
```
EricGustin: Since you removed the `additional_prompt_content` this feels more like a `ToolExecutionError` now

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -940,3 +961,251 @@ async def list_direct_message_conversations_metadata(
     )
 
     return response  # type: ignore[no-any-return]
+
+
+@tool(requires_auth=Slack(scopes=["im:history", "im:read", "users:read", "users:read.email"]))
```
EricGustin, line 742: why does this tool need the users:read.email scope? Comment applies to the other "username" tools too

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -940,3 +961,251 @@ async def list_direct_message_conversations_metadata(
...
+    limit: Annotated[int | None, "The maximum number of messages to return."] = None,
+    next_cursor: Annotated[str | None, "The cursor to use for pagination."] = None,
+) -> Annotated[
+    dict,
+    (
+        "The messages in a direct message conversation and next cursor for paginating results "
+        "when there are additional messages to retrieve."
+    ),
+]:
+    """Get the messages in a direct conversation by the user's name.
```
EricGustin, line 783: hmm this tool description is more than 1024 characters which was openai's limit, but it looks like they've quietly increased that limit. Are there other model providers that this tool description would be too long for?

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -940,3 +961,251 @@ async def list_direct_message_conversations_metadata(
...
+            current_user["user_id"],
+            *[user["id"] for user in other_users["users"] if user["id"] != current_user["user_id"]],
+        ],
+        exact_match=True,
+        limit=1,
+        next_cursor=next_cursor,
+    )
+
+    if not conversations_found:
+        raise RetryableToolError(
```
EricGustin: why Retryable here? maybe a ToolExecutionError or return empty list with a message saying no such conversation exists

toolkits/slack/arcade_slack/tools/users.py
```diff
@@ -52,23 +58,19 @@ async def get_user_info_by_id(
...
-    requires_auth=Slack(
-        scopes=["users:read", "users:read.email"],
-    )
-)
+@tool(requires_auth=Slack(scopes=["users:read", "users:read.email"]))
 async def list_users(
     context: ToolContext,
     exclude_bots: Annotated[bool | None, "Whether to exclude bots from the results"] = True,
-    limit: Annotated[int | None, "The maximum number of users to return."] = None,
+    limit: Annotated[
```
EricGustin, line 53: I wonder how this would perform on a large slack workspace (e.g., 20,000 users). Will it always timeout and fail for these big workspaces? Should we default to a reasonable limit instead of not having a limit?

toolkits/slack/arcade_slack/models.py
byrro: 💡

toolkits/slack/arcade_slack/utils.py
byrro: In the tools this function is used, if the value is not an email, it assumes it's a username. If the username is invalid, the tool will raise a user not found error. Does that make sense?

toolkits/slack/arcade_slack/tools/chat.py
byrro: It depends on the `get_user_by_username` tool, which [requires this scope](https://api.slack.com/changelog/2017-04-narrowing-email-access) to access the users email address through the `slack.users_list` endpoint.

toolkits/slack/arcade_slack/tools/users.py
byrro: yea, we should take this opportunity and fix that too, working on it

toolkits/slack/arcade_slack/tools/chat.py
byrro: I summarized and simplified the docstring to 732 chars.

toolkits/slack/arcade_slack/models.py
byrro: actually wouldn't be efficient: would need to loop through `last_result` for each pending username.. better to keep the current independent implementation

toolkits/slack/arcade_slack/utils.py
EricGustin: makes sense. regex comment still stands as a suggestion though

toolkits/slack/arcade_slack/tools/chat.py
```diff
@@ -8,27 +7,371 @@
...
+        ),
+    ] = None,
+    latest_datetime: Annotated[
+        str | None,
+        (
+            "The latest message to include in the results, specified as a datetime object in the "
+            "format 'YYYY-MM-DD HH:MM:SS'"
+        ),
+    ] = None,
+    limit: Annotated[int | None, "The maximum number of messages to return."] = None,
```
EricGustin: can we limit to maybe 50 or 100 messages by default? I just asked to get dms between us & apparently we talk a lot b/c it exceeded the model's maximum context length

byrro:
> Question: why not have a single tool that can get a user based on their id, username, or email? Same question for getting multiple users.

@EricGustin I underestimated the refactoring journey this comment would lead me to 😆

EricGustin:
> > Question: why not have a single tool that can get a user based on their id, username, or email? Same question for getting multiple users.
> 
> 
> 
> @EricGustin I underestimated the refactoring journey this comment would lead me to 😆 

"In doing so, I also found..." 🤣

PR #454, opened 2025-06-23
`--debug` for CLI commands

evantahler:
```
(base) ➜  arcade-ai git:(debug) ✗ uv run arcade busted-command --host invalid-host
❌ Login failed with error: Not implemented
^C
Aborted.


(base) ➜  arcade-ai git:(debug) ✗ uv run arcade busted-command --host invalid-host --debug
❌ Login failed with error: Traceback (most recent call last):
  File "/Users/evan/workspace/arcade/Team/arcade-ai/libs/arcade-cli/arcade_cli/main.py", line 106, in login
    raise Exception("Not implemented")
Exception: Not implemented
```

### DIFF
--- a/libs/arcade-cli/arcade_cli/main.py
+++ b/libs/arcade-cli/arcade_cli/main.py
@@ -70,12 +70,19 @@ cli.add_typer(
 console = Console()
 
 
-def handle_error_with_debug(error: Exception, message: str, debug: bool) -> None:
-    """Handle error reporting with optional debug traceback."""
-    if debug:
+def handle_cli_error(
+    message: str, error: Exception | None = None, debug: bool = True, should_exit: bool = True
+) -> None:
+    """Handle CLI error reporting with optional debug traceback and exit."""
+    if error and debug:
         console.print(f"❌ {message}: {traceback.format_exc()}", style="bold red")
-    else:
+    elif error:
         console.print(f"❌ {message}: {escape(str(error))}", style="bold red")
+    else:
+        console.print(f"❌ {message}", style="bold red")
+
+    if should_exit:
+        raise typer.Exit(code=1)
 
 
 @cli.command(help="Log in to Arcade Cloud", rich_help_panel="User")
@@ -126,8 +133,7 @@ def login(
     except KeyboardInterrupt:
         auth_server.shutdown_server()
     except Exception as e:
-        handle_error_with_debug(e, "Login failed", debug)
-        raise typer.Exit(code=1)
+        handle_cli_error("Login failed", e, debug)
     finally:
         if server_thread.is_alive():
             server_thread.join()  # Ensure the server thread completes and cleans up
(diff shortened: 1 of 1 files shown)


libs/arcade-cli/arcade_cli/main.py
```diff
@@ -69,6 +70,14 @@
 console = Console()
 
 
+def handle_error_with_debug(error: Exception, message: str, debug: bool) -> None:
```
Spartee: might as well call this ``handle_cli_error(message: str, error: Exception | None = None, debug: bool = True, exit : bool = True``

would get rid of alot of typer.exits()

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -69,6 +70,14 @@
...
+    """Handle error reporting with optional debug traceback."""
+    if debug:
+        console.print(f"❌ {message}: {traceback.format_exc()}", style="bold red")
+    else:
+        console.print(f"❌ {message}: {escape(str(error))}", style="bold red")
+
+
 @cli.command(help="Log in to Arcade Cloud", rich_help_panel="User")
```
Spartee, line 88: we turn off the typer/rich tracebacks. have you seen those?

evantahler: yeah, in the setup [here](https://github.com/ArcadeAI/arcade-ai/blob/main/libs/arcade-cli/arcade_cli/main.py#L52-L60), and I think that's still the correct configuration 99% of the time.  This came up when there was something wrong with our code that we wanted to debug and get the stack trace, but I think this will be rare

evantahler:
Bump @byrro - I think this is what you did to help a customer debug, and I wanted to add it to our CLI for real

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -376,6 +398,7 @@ def evals(
         "--no-tls",
         help="Whether to disable TLS for the connection to the Arcade Engine.",
     ),
+    debug: bool = typer.Option(False, "--debug", help="Show debug information"),
```
EricGustin, line 401:
```suggestion
    debug: bool = typer.Option(False, "--debug", "-d", help="Show debug information"),
```

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -620,6 +642,7 @@ def deploy(
...
```
EricGustin, line 645:
```suggestion
    debug: bool = typer.Option(False, "--debug", "-d", help="Show debug information"),
```

EricGustin (approved):

libs/arcade-cli/arcade_cli/main.py
evantahler: I didn't want to have a short command because this is more for us then customers... I hope.

PR #458, opened 2025-06-25
Add Zendesk Toolkit

lpetralli:
This PR adds a new toolkit for integrating with Zendesk Help Center and ticketing system.

  Features

  - Search Articles: Search Help Center knowledge base with filters for text, categories, labels, and dates
  - List Tickets: Retrieve all open support tickets
  - Get Comments: Retrieve all comments for a specific ticket to understand the context
  - Add Comments: Add public or internal comments to existing tickets
  - Solve Tickets: Mark tickets as solved with optional internal resolution comments

  Testing & Quality
  - Comprehensive test suite 
  - Evaluation scripts for real-world testing
  - All tests passing (make test)
  - Code quality checks passing (make check)
  - All evals passing


lgesuellip:
now LGTM — deployed and verified on staging.

evantahler:
+ @sdserranog do we need any assets/logos as well?

sdserranog:
Yes! We need to add the logo to our dashboard, docs and chat. We can do it before or after this PR is merged.

toolkits/zendesk/arcade_zendesk/__init__.py
```diff
@@ -0,0 +1,13 @@
+from arcade_zendesk.tools import (
+    add_ticket_comment,
+    list_tickets,
```
byrro, line 4: `get_ticket_comments` tool is missing in the imports here

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
+import logging
+from typing import Annotated, Any
+
+import httpx
+from arcade_tdk import ToolContext, tool
+from arcade_tdk.auth import OAuth2
+
+from ..utils import (
```
byrro:
```suggestion
from arcade_zendesk.utils import (
```

absolute reference `from arcade_zendesk.utils import ...` is better than relative for readability and avoids breaking imports in a future refactoring

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+@tool(
+    requires_auth=OAuth2(id="zendesk", scopes=["read"]),
+    requires_secrets=["ZENDESK_SUBDOMAIN"],
+)
+async def search_articles(
+    context: ToolContext,
+    query: Annotated[
+        str | None,
+        "Search text to match against articles. Can use quoted phrases for exact matches "
+        "(e.g., 'exact phrase') or multiple terms (e.g., 'carrot potato')",
```
byrro: We try to avoid providing content samples in tool arg docstrings. It's not common, but sometimes it can lean LLMs into using the examples we give, despite not having any actual relevance.

In this case, I think we could have a docstring like `"Search text to match against articles. Supports quoted expressions for exact matching."`.

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+    sort_by: Annotated[
+        str | None,
+        "Sort results by 'created_at'. Defaults to relevance when omitted",
+    ] = None,
+    sort_order: Annotated[
+        str | None,
+        "Sort order: 'asc' (ascending) or 'desc' (descending). Defaults to 'desc'",
+    ] = None,
+    per_page: Annotated[int, "Number of results per page (maximum 100)"] = 10,
+    all_pages: Annotated[
```
byrro: I'd suggest to return only one page per tool call and get rid of `all_pages` and `max_pages` args. Lengthy tool responses can consume too much memory in the worker env, take too long to process, and confuse LLMs or exhaust their context window.

In the tool response, we can include the `next_token` for pagination and accept a `next_token` argument in the tool interface. When the user asks "get the next page of articles", the LLM can call the tool again with the token.

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+    date_params = {
+        "created_after": created_after,
+        "created_before": created_before,
+        "created_at": created_at,
+    }
+
+    for param_name, param_value in date_params.items():
+        if param_value and not validate_date_format(param_value):
+            return {
+                "error": "InvalidDateFormat",
```
byrro: In `arcade_tdk` we have exceptions that can be used in cases like this:

`from arcade_tdk.errors import RetryableToolError, ToolExecutionError`

Both will take a `message` (returned to the LLM and potentially displayed to the end user) and `developer_message` (directed to the developer using the Arcade API/SDK).

The `RetryableToolError` also takes a `retry_after_ms` and `additional_prompt_content` (useful to include instructions for the LLM to try again correctly).

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+    ] = None,
+    created_before: Annotated[
+        str | None,
+        "Filter articles created before this date (format: YYYY-MM-DD, e.g., '2024-01-15')",
+    ] = None,
+    created_at: Annotated[
+        str | None,
+        "Filter articles created on this exact date (format: YYYY-MM-DD, e.g., '2024-01-15')",
+    ] = None,
+    sort_by: Annotated[
```
byrro, line 47: Tools support `Enum` types. We could have here:

```
from enum import Enum

class ArticleSortBy(Enum):
    CREATED_AT = "created_at"
    RELEVANCE = "relevance"

async def search_articles(
    ...,
    sort_by: Annotated[ArticleSortBy, "Field to sort articles. Defaults to relevance according to the search query."] = ArticleSortBy.RELEVANCE,
```

Note: when the Arcade Engine exposes the tool interface to the LLM, it only provides the Enum values, not the keys. In the example above, the LLM would see two possible values for this argument: `"created_at"` and `"relevance"`.

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+        str | None,
+        "Sort results by 'created_at'. Defaults to relevance when omitted",
+    ] = None,
+    sort_order: Annotated[
```
byrro, line 51: An `Enum` would also fit well here.

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+        str | None,
+        "Search text to match against articles. Can use quoted phrases for exact matches "
+        "(e.g., 'exact phrase') or multiple terms (e.g., 'carrot potato')",
+    ] = None,
+    label_names: Annotated[
+        str | None,
+        "Comma-separated list of label names (case-insensitive). Article must have at least "
+        "one matching label. Available on Professional/Enterprise plans only",
+    ] = None,
+    category: Annotated[
```
byrro: Since we don't have tools to retrieve existing categories and sections, I'm in doubt whether we should keep these two arguments.

What we try to do, when feasible, is to make the tool interface the closest possible to what a human user will ask. No one will ask an LLM for articles in a category/section ID. People will always ask `"search for 'foobar' in the tutorials section"`.

Ideally, we should accept a category/section name. Inside the tool, we lookup that category/section ID and then search. But that's going to add a lot more work on top of what you've already done.

Maybe for now we should remove these two arguments, since there's no way to retrieve a category/section ID to use them.

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+        int | None,
+        "Filter by category ID. Can specify multiple by comma-separating values (e.g., '123,456')",
+    ] = None,
+    section: Annotated[
+        int | None,
+        "Filter by section ID. Can specify multiple by comma-separating values (e.g., '789,101')",
+    ] = None,
+    created_after: Annotated[
+        str | None,
+        "Filter articles created after this date (format: YYYY-MM-DD, e.g., '2024-01-15')",
```
byrro: We try to avoid providing content samples ('2024-01-15'). When in doubt, LLMs sometimes can take the sample and use it in a tool call, despite having zero relevance.

The expected format (YYYY-MM-DD) is relevant and important to keep.

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+                pages_to_fetch = max_pages
+            else:
+                pages_to_fetch = 1  # Single page (default behavior)
+
+            data = await fetch_all_pages(client, url, headers, params, max_pages=pages_to_fetch)
+            if "results" in data:
+                data["results"] = process_search_results(data["results"], include_body=include_body)
+            logger.info(f"Article search results: {data}")
+
+        except httpx.HTTPStatusError as e:
```
byrro, line 201: There is some code duplicated across tools to handle various types of errors. Perhaps it'd be better to create a simple `ZendeskClient` class, group this logic there and abstract it away from the tool code.

[An example](https://github.com/ArcadeAI/arcade-ai/blob/d86cc1c301a35c94ec03995a2626ec908fcf1012/toolkits/jira/arcade_jira/client.py#L111) from the Jira toolkit, recently published.

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+    # Validate that at least one search parameter is provided
+    if not any([query, category, section, label_names]):
+        return {
+            "error": "MissingSearchParameter",
+            "message": "At least one of query, category, section, or label_names must be provided.",
+        }
+
+    auth_token = context.get_auth_token_or_empty()
+
+    if not auth_token:
```
byrro: The Arcade Engine already takes care of making sure the auth token is available and valid. Inside tools, we simply assume the token will be there and good to be used.

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+        return {
+            "error": "AuthenticationError",
+            "message": "No authentication token found. Please authenticate with Zendesk first.",
+        }
+
+    subdomain = context.get_secret("ZENDESK_SUBDOMAIN")
+    if not subdomain:
```
byrro: `context.get_secret` will raise a `ValueError` if the secret is not set. Instead of the `if` statement, would need to wrap the call to `context.get_secret` inside a `try / except` block to customize with a descriptive error message. Can raise a `ToolExecutionError`:

```
try:
    subdomain = context.get_secret("ZENDESK_SUBDOMAIN")
except ValueError:
    raise ToolExecutionError(
        message="Zendesk subdomain is not set.",
        developer_message="Zendesk subdomain is not set. Make sure to set the 'ZENDESK_SUBDOMAIN' secret in the Arcade Dashboard.",
    )
```

Since this is done in every tool, perhaps encapsulating the logic in a `get_zendesk_subdomain(context)` function would make sense.

toolkits/zendesk/arcade_zendesk/utils.py
```diff
@@ -0,0 +1,157 @@
...
+    clean_text = re.sub(r"\n+", "\n", clean_text)
+
+    clean_text = re.sub(r"\s+", " ", clean_text)
+
+    clean_text = "\n".join(line.strip() for line in clean_text.split("\n"))
+
+    return clean_text.strip()
+
+
+def truncate_text(
```
byrro, line 126: Truncating is a good strategy to avoid returning too much data in the `search_articles` call. But I'm concerned that we don't offer a way to dynamically override the 500 char limit nor to retrieve an entire article, if needed.

Perhaps we should have a `max_article_length` argument in the `search_articles` tool? Ideally, we'd have a `get_full_article` tool.

Otherwise, the truncation would render the tool practically useless when articles have +500 chars.

toolkits/zendesk/arcade_zendesk/utils.py
```diff
@@ -0,0 +1,157 @@
...
+        processed_results.append(processed_result)
+
+    return processed_results
+
+
+def validate_date_format(date_string: str) -> bool:
+    """Validate that a date string matches YYYY-MM-DD format."""
+    import re
+
+    pattern = r"^\d{4}-\d{2}-\d{2}$"
```
byrro: This pattern would accept dates like '2025-99-99'. Instantiating a `datetime.datetime` class and handling a potential exception would be a more robust validation.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,243 @@
...
+            tickets = data.get("tickets", [])
+
+            # Format the tickets for display
+            if not tickets:
+                return "No open tickets found."
+
+            result = []
+            for ticket in tickets:
+                ticket_info = (
+                    f"Ticket #{ticket['id']}: {ticket['subject']} (Status: {ticket['status']})"
```
byrro: I think it'd be safer to return structured data (dict, list), instead of concatenated text. Avoids confusion, especially because Zendesk content is heavy on free text.

A structured output would also make it a lot easier for developers to manipulate responses programmatically, when needed.

[suggestion applies to other tools as well]

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,243 @@
...
+import httpx
+from arcade_tdk import ToolContext, tool
+from arcade_tdk.auth import OAuth2
+
+
+@tool(
+    requires_auth=OAuth2(id="zendesk", scopes=["read"]),
+    requires_secrets=["ZENDESK_SUBDOMAIN"],
+)
+async def list_tickets(context: ToolContext) -> Annotated[str, "The tickets"]:
```
byrro: In this kind of tool, we usually support limit & pagination, so that the user can browse more items. Otherwise, the tool is limited to only a first set of tickets.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,243 @@
...
+            data = response.json()
+            comments = data.get("comments", [])
+
+            if not comments:
+                return f"No comments found for ticket #{ticket_id}."
+
+            result = [f"Comments for Ticket #{ticket_id}:\n"]
+
+            for i, comment in enumerate(comments):
+                author_id = comment.get("author_id", "Unknown")
```
byrro: Very good to respond with only what's relevant. Makes LLMs cheaper to run and less prone to confusion. The only thing I'd change here is perhaps include the list of attachments.

byrro (changes requested):
hey @lpetralli, thank you for submitting this PR, you guys did a very good job on the toolkit! After internal discussion, there are a few changes that we consider necessary for the toolkit to be production-ready.

There are a couple of comments about code encapsulation/organization that are simply suggestions, feel free to consider them.

Please let me know if there's anything unclear in any of the comments. I'm available to discuss on Slack and we can hop on a call as well, if you think it's better.

toolkits/zendesk/arcade_zendesk/utils.py
lgesuellip: Makes sense!  I’ll add the parameter to the tool so we can control truncation dynamically.
And yeah, ideally we’d also have a `get_full_article` to fetch the entire content when needed.

toolkits/zendesk/.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+files: ^.*/stripe/.*
```
EricGustin: stripe --> zendesk

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,243 @@
...
+    requires_secrets=["ZENDESK_SUBDOMAIN"],
+)
+async def search_articles(
+    context: ToolContext,
+    query: Annotated[
+        str | None,
+        "Search text to match against articles. Can use quoted phrases for exact matches "
+        "(e.g., 'exact phrase') or multiple terms (e.g., 'carrot potato')",
+    ] = None,
+    label_names: Annotated[
```
EricGustin, line 30: We're trusting that the LLM will provide a valid comma separated string without verifying. To ensure an invalid comma separated string is never provided, I would either
1. Make the type a `list[str]` and then split with `,` as the separator before sending to the API
2. or write a helper function that validates the input is correctly comma separated

toolkits/zendesk/pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+dependencies = [
+    "arcade-tdk>=2.0.0,<3.0.0",
+    "httpx>=0.25.0,<1.0.0",
+    "beautifulsoup4>=4.0.0,<5"
+]
+
+
+[project.optional-dependencies]
+dev = [
+    "arcade-ai[evals]>=2.0.0rc1,<3.0.0",
```
EricGustin: this can be updated to "arcade-ai[evals]>=2.0.0,<3.0.0" now that the arcade-ai package is past the release candidate stage

toolkits/zendesk/arcade_zendesk/tools/tickets.py
EricGustin: in what order are tickets returned? Can we make this configurable and also specify the default order in the parameter annotation

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,243 @@
...
+    # Make the API request
+    async with httpx.AsyncClient() as client:
+        headers = {
+            "Authorization": f"Bearer {token}",
+            "Content-Type": "application/json",
+        }
+
+        response = await client.get(url, headers=headers)
+
+        if response.status_code == 200:
```
EricGustin: typical convention for toolkits that use httpx is to do response.raise_for_status()

[Example](https://github.com/ArcadeAI/arcade-ai/blob/d86cc1c301a35c94ec03995a2626ec908fcf1012/toolkits/reddit/arcade_reddit/client.py#L21)

toolkits/zendesk/arcade_zendesk/tools/tickets.py
EricGustin: Same. Please make the tool outputs more structured.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,243 @@
...
+            return "\n".join(result)
+        else:
+            return f"Error fetching tickets: {response.status_code} - {response.text}"
+
+
+@tool(
+    requires_auth=OAuth2(id="zendesk", scopes=["read"]),
+    requires_secrets=["ZENDESK_SUBDOMAIN"],
+)
+async def get_ticket_comments(
```
EricGustin, line 165: Given that I haven't personally used Zendesk before, is it extremely rare for tickets to have >100 comments? If not, then this tool should have pagination support.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,243 @@
...
+@tool(
+    requires_auth=OAuth2(id="zendesk", scopes=["tickets:write"]),
+    requires_secrets=["ZENDESK_SUBDOMAIN"],
+)
+async def add_ticket_comment(
+    context: ToolContext,
+    ticket_id: Annotated[int, "The ID of the ticket to comment on"],
+    comment_body: Annotated[str, "The text of the comment"],
+    public: Annotated[
+        bool, "Whether the comment is public (visible to requester) or internal"
```
EricGustin: This comment applies to all tool parameters that have a default value: 

The default value is not provided to the LLM, so our convention is to append "Defaults to x" to the parameter annotation so that the model has this context.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,243 @@
...
+        headers = {
+            "Authorization": f"Bearer {token}",
+            "Content-Type": "application/json",
+        }
+
+        response = await client.put(url, headers=headers, json=request_body)
+
+        if response.status_code == 200:
+            comment_type = "public" if public else "internal"
+            return f"Successfully added {comment_type} comment to ticket #{ticket_id}"
```
EricGustin: This tool should also be returning the url to the ticket so that a model can say something like "I've successfully added your comment. To view the ticket where your comment has been added, **click here**.

Comment applies to the other ticket tools

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,243 @@
...
+        "Optional final comment to add when solving (e.g., resolution summary)",
+    ] = None,
+    comment_public: Annotated[bool, "Whether the comment is visible to the requester"] = False,
+) -> Annotated[str, "Result of marking ticket as solved"]:
+    """Mark a Zendesk ticket as solved, optionally with a final comment.
+
+    Args:
+        ticket_id: The ID of the ticket to mark as solved
+        comment_body: Optional final comment to add when solving (e.g., resolution summary)
+        comment_public: Whether the comment is visible to the requester (default False - internal)
```
EricGustin: No need to have this in the docstring because the parameter annotations are sent to the model. So this just ends up adding unneeded tokens to the request.

This comment applies to all tools

EricGustin (commented):
In addition to @byrro 's review I've added some comments inline

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
lgesuellip: I feel we’d handle this way better as chunks of data, using a limit + cursor pattern. It’s just more natural for an LLM, and honestly for users too:
- cursor: keeps track of where we left off
- limit: says how many more to grab

I really don’t see LLMs or people thinking in pages (like “get the next page of articles”). 
It’s much more: “Get me 50 articles", “Fetch more from last year."

Implementation:
    - Tool fetches multiple pages internally as needed
    - Cursor encodes the offset (total items already returned)
  
What do you think @lpetralli @byrro ?

toolkits/zendesk/pyproject.toml
lgesuellip: Perfect!

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
lgesuellip: Otherwise, since it’s a search tool, it might be better to just think in terms of “give me the top K best results.” The tool can handle whether that means fetching 1 page or 5 behind the scenes, so the user/llms doesn’t have to worry about it — no need to think in “pages” or “cursors” at all.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: The current implementation already returns the full comment objects from Zendesk's API, which should include the attachments field if present. I will update the docstring to explicitly document that attachments are included in the response.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: currently the tickets are sorted by ID (oldest to newest) by default, as that's Zendesk's default behavior when no sort parameters are specified. I'll add configurable sorting to the list_tickets function with the ability to sort from newest to oldest.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: I agree that it is extremely rare for tickets to have >100 comments - considering that each public comments means a mail sent or received from the customer. I do not think pagination is needed here.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: the Zendesk API already includes URLs in the ticket objects it returns:
  - url: The API endpoint URL
  - html_url: The web interface URL (what users would click to view the ticket)

  So when using these tools, the LLM can access the ticket URL directly from the response.

EricGustin:
@lgesuellip we updated the Makefile for existing toolkits today. Please update Zendesk's to match what is in main branch when you find the time.

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: fixed in commit [aba0db4](https://github.com/ArcadeAI/arcade-ai/pull/458/commits/aba0db4662bfb5a131722a83647a5bf573196482)

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: fixed in commit [fcce88c](https://github.com/ArcadeAI/arcade-ai/pull/458/commits/fcce88c9f823cc3f04bde781d90c75951c3e769c)

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: updated docstring in commit [69c379e](https://github.com/ArcadeAI/arcade-ai/pull/458/commits/69c379e6911ddabe67a65868240fa0a83aff3af5)

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: fixed in commit [f1234c2](https://github.com/ArcadeAI/arcade-ai/pull/458/commits/f1234c2d3ca638199b2a5e8ea11342071ca611f4)

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: fixed in commit [f1234c2](https://github.com/ArcadeAI/arcade-ai/pull/458/commits/f1234c2d3ca638199b2a5e8ea11342071ca611f4)

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: fixed in commit [a7d1df9](https://github.com/ArcadeAI/arcade-ai/pull/458/commits/a7d1df9d739719569c36ba03e29109ca8e4d42ac)

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: addressed in commit [5ebf273](https://github.com/ArcadeAI/arcade-ai/pull/458/commits/5ebf27343f9f6c7bbccd46b0eb70b47f99323c79)

I am removing the api url and just the html url remains so that the agent is not confused (after some testing in the playground I noticed the api url was sometimes used to direct the user to the link)

toolkits/zendesk/arcade_zendesk/tools/tickets.py
lpetralli: fixed in commit [1cd43e3](https://github.com/ArcadeAI/arcade-ai/pull/458/commits/1cd43e3898827279ad5695b627e6d8b60ede127d)

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
EricGustin: An offset & limit sounds good to me. (In your words: cursor & limit)

I'm not a fan of "top k", because it removes the ability to satisfy queries such as "get the next 5"

toolkits/zendesk/arcade_zendesk/tools/search_articles.py
```diff
@@ -0,0 +1,223 @@
...
+
+    CREATED_AT = "created_at"
+    RELEVANCE = "relevance"
+
+
+class SortOrder(Enum):
+    """Sort order direction."""
+
+    ASC = "asc"
+    DESC = "desc"
```
EricGustin: nit: we typically place enums at `toolkits/zendesk/zendesk_toolkit/enums.py` and reserve code inside `tools/` directory to only contain functions with `@tool` decorator

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,410 @@
...
+    requires_secrets=["ZENDESK_SUBDOMAIN"],
+)
+async def list_tickets(
+    context: ToolContext,
+    status: Annotated[
+        str,
+        "The status of tickets to filter by. "
+        "Valid values: 'new', 'open', 'pending', 'solved', 'closed'. "
+        "Defaults to 'open'",
+    ] = "open",
```
EricGustin: perfect opportunity for enum

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,410 @@
...
+    ] = None,
+    cursor: Annotated[
+        str | None,
+        "The cursor for pagination. Use 'after_cursor' from previous response to get next page.",
+    ] = None,
+    sort_order: Annotated[
+        Literal["asc", "desc"],
+        "Sort order for tickets by ID. 'asc' returns oldest first, 'desc' returns newest first. "
+        "Defaults to 'desc'",
+    ] = "desc",
```
EricGustin: perfect opportunity for enum

toolkits/zendesk/arcade_zendesk/tools/tickets.py
```diff
@@ -0,0 +1,410 @@
...
+            retry_after_ms=500,
+            additional_prompt_content="Please verify the ticket ID and try again",
+        )
+
+
+@tool(
+    requires_auth=OAuth2(id="zendesk", scopes=["read"]),
+    requires_secrets=["ZENDESK_SUBDOMAIN"],
+)
+async def list_tickets(
```
EricGustin, line 27: Pagination needs to be updated to be `limit` and `offset` like the search articles. If there are other tools that have pagination then make them `limit` and `offset`.

lpetralli-ncl:
- Fixed pagination error: Replaced cursor-based pagination with offset/limit pagination in list_tickets to avoid
  InvalidPaginationParameter errors
- Added type safety: Now using TicketStatus and SortOrder enums instead of string literals
- Unified approach: Both list_tickets and search_articles now use the same pagination pattern with fetch_paginated_results utility
- Updated tests/evals: All tests and evaluation cases updated to use new parameters and enum types

EricGustin (approved):
LGTM!

evantahler:
Please pull in the main branch - https://github.com/ArcadeAI/arcade-ai/pull/495 should prevent the postgres test from running on forks

evantahler:
One more merge please - https://github.com/ArcadeAI/arcade-ai/pull/496 should help

PR #459, opened 2025-06-25
Postgres Database Toolkit

evantahler:
Adds an example of a good "general case" SQL tool:
* enforces read-only mode
* hints to the LLM to discover the tables and schemas for the tables it needs before any query
* uses RetryableToolErrors to hint to the LLM about what to do next

Docs: https://github.com/ArcadeAI/docs/pull/345

For testing, `TEST_POSTGRES_DATABASE_CONNECTION_STRING` has been set in the repo (from Neon).  details in 1 password.

464977013-49aff5e5-e301-4ca0-83b5-3ea742db2283


toolkits/postgres/arcade_postgres/tools/database_engine.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # try again
+            try:
+                async with cls._engines[key].connect() as connection:
+                    await connection.execute(text(TEST_QUERY))
+                return cls._engines[key]
+            except Exception as e:
+                raise RetryableToolError(
+                    f"Connection failed: {e}",
+                    developer_message=f"Connection to '{connection_string}' failed.",
```
evantahler: Don't return the whole connection string. Just the host.

evantahler:
Yeah! 
Screenshot 2025-07-10 at 3 58 42 PM

.github/workflows/test-toolkits.yml
```diff
@@ -51,3 +51,5 @@ jobs:
       - name: Test toolkit
         working-directory: toolkits/${{ matrix.toolkit }}
         run: uv run --active pytest -W ignore -v --cov=arcade_${{ matrix.toolkit }} --cov-report=xml
+        env:
+          TEST_POSTGRES_DATABASE_CONNECTION_STRING: ${{ secrets.TEST_POSTGRES_DATABASE_CONNECTION_STRING }} # TODO: dynamically only load the `TEST_${{ matrix.toolkit }}_DATABASE_CONNECTION_STRING secret`
```
evantahler, line 54: Any tips here? I don't love this... but it's fine I guess

toolkits/postgres/arcade_postgres/tools/database_engine.py
evantahler, line 1: This class manages an auto-reconnecting singleton connection for the database.  The first request will establish it, and then all other requests will re-use the connection (being faster).  If the connection dies, it will be re-esablished on the next request

toolkits/postgres/arcade_postgres/tools/database_engine.py
```diff
@@ -0,0 +1,90 @@
...
+from arcade_tdk.errors import RetryableToolError
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
+
+MAX_ROWS_RETURNED = 1000
+DEFAULT_ISOLATION_LEVEL = "READ COMMITTED"
+TEST_QUERY = "SELECT 1"
+ERROR_REMAPPING = {
+    re.compile(r"This result object does not return rows"): "Only SELECT queries are allowed.",
+}
```
evantahler: This is fun - this makes the retryable tool errors better AND helps humans understand what's going on too!

toolkits/postgres/arcade_postgres/tools/database_engine.py
```diff
@@ -0,0 +1,90 @@
...
+    ) -> AsyncEngine:
+        parsed_url = urlparse(connection_string)
+
+        # TODO: something strange with sslmode= and friends
+        # query_params = parse_qs(parsed_url.query)
+        # query_params = {
+        #     k: v[0] for k, v in query_params.items()
+        # }  # assume one value allowed for each query param
+
+        async_connection_string = f"{parsed_url.scheme.replace('postgresql', 'postgresql+asyncpg')}://{parsed_url.netloc}{parsed_url.path}"
```
evantahler: sqlalchemy is silly, and the URL is how you say "be async"

docker/toolkits.txt
```diff
@@ -10,6 +10,7 @@ arcade-linkedin
 arcade-math
 arcade-microsoft
 arcade-notion-toolkit
+arcade-postgres
```
EricGustin: omit so that it isn't added to our hosted worker

toolkits/postgres/README.md
EricGustin, line 1: Convention is to not have a README for the toolkits. Maybe we should in the future, but I guess that's what docs.arcade.dev is for

toolkits/postgres/arcade_postgres/tools/postgres.py
```diff
@@ -0,0 +1,155 @@
...
+from sqlalchemy import inspect, text
+from sqlalchemy.ext.asyncio import AsyncEngine
+
+from .database_engine import ERROR_REMAPPING, MAX_ROWS_RETURNED, DatabaseEngine
+
+
+@tool(requires_secrets=["DATABASE_CONNECTION_STRING"])
+async def discover_tables(
+    context: ToolContext,
+    schema_name: Annotated[str, "The database schema to discover tables in"] = "public",
```
EricGustin: This comment applies to all tool inputs that have a default value:

Appending "Defaults to <value>" to the end of the annotation adds this context to the tool definition. Otherwise the model will not see the default value in the tool description. 

This helps in at least two ways:
1. The model doesn't feel pressure to try to come up with a value that it may hallucinate.
2. The model is 'anchored' by your default value. This is more apparent for params like `limit`. For example, a model's natural state might be to default to limiting to 5 results, but if your default is 100_000, then the model becomes more aligned with the developer's preference. Ideally the developer's preference is influenced by the common use case & what would be useful for an end-user.

toolkits/postgres/arcade_postgres/tools/postgres.py
```diff
@@ -0,0 +1,155 @@
...
+) -> list[str]:
+    """Discover all the tables in the postgres database when the list of tables is not known.
+
+    THIS TOOL SHOULD ALWAYS BE USED BEFORE ANY OTHER TOOL THAT REQUIRES A TABLE NAME.
```
EricGustin, line 32: I like this type of comment. I should incorporate it more in my own tool docstrings.

toolkits/postgres/arcade_postgres/tools/database_engine.py
EricGustin, line 1: nit/dogma: generally we reserve files inside of tools/ directory to only include tool functions. And then we place files like this at arcade_postgres/

toolkits/postgres/arcade_postgres/tools/database_engine.py
```diff
@@ -0,0 +1,90 @@
+import re
+from typing import Any, ClassVar
+from urllib.parse import urlparse
+
+from arcade_tdk.errors import RetryableToolError
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
+
+MAX_ROWS_RETURNED = 1000
+DEFAULT_ISOLATION_LEVEL = "READ COMMITTED"
```
EricGustin: As discussed IRL, something seems off here

toolkits/postgres/arcade_postgres/tools/postgres.py
```diff
@@ -0,0 +1,155 @@
...
+        context.get_secret("DATABASE_CONNECTION_STRING")
+    ) as engine:
+        tables = await _get_tables(engine, schema_name)
+        return tables
+
+
+@tool(requires_secrets=["DATABASE_CONNECTION_STRING"])
+async def get_table_schema(
+    context: ToolContext,
+    schema_name: Annotated[str, "The database schema to get the table schema of"],
```
EricGustin, line 44: possible future feature: I wonder if a discover_schemas/list_schemas tool would be useful? As someone who did not create the db, I'm not sure which schema names are available to me.

toolkits/postgres/arcade_postgres/tools/database_engine.py
```diff
@@ -88,3 +76,29 @@ async def cleanup(cls) -> None:
     def clear_cache(cls) -> None:
         """Clear the engine cache without disposing engines. Use with caution."""
         cls._engines.clear()
+
+    @classmethod
+    def sanitize_query(cls, query: str) -> str:
```
EricGustin: I understand this isn't going to be for 'production', but maybe at least additionally check for write keywords like DELETE, INSERT & error if they are in the str? Idk up to you

EricGustin (approved):
pre-approving in case I'm offline when you need to merge

PR #460, opened 2025-06-30
patching toolkit template generator for outside the main repo

torresmateo:
The current `arcade new` command fails to `make install` and `make check` if it's not manually patched with pre-commit files. It also fails due to local dependencies if it's not run inside `arcade-ai/toolkits`

This PR patches the toolkit template to only include the local dependencies if it detects is `arcade new` being run inside `arcade-ai/toolkits`, it also makes pre-commit commands optional.

### DIFF
--- a/libs/arcade-cli/arcade_cli/templates/{{ toolkit_name }}/Makefile
+++ b/libs/arcade-cli/arcade_cli/templates/{{ toolkit_name }}/Makefile
@@ -8,14 +8,14 @@ help:
 install: ## Install the uv environment and install all packages with dependencies
 	@echo "🚀 Creating virtual environment and installing all packages using uv"
 	@uv sync --active --all-extras --no-sources
-	@uv run pre-commit install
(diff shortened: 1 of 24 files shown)


libs/arcade-cli/arcade_cli/templates/{{ toolkit_name }}/Makefile
```diff
@@ -8,14 +8,14 @@ help:
 install: ## Install the uv environment and install all packages with dependencies
 	@echo "🚀 Creating virtual environment and installing all packages using uv"
 	@uv sync --active --all-extras --no-sources
-	@uv run pre-commit install
+	@if [ -f .pre-commit-config.yaml ]; then uv run pre-commit install; fi
```
EricGustin: I think you're missing the --no-sources flag? 

`@if [ -f .pre-commit-config.yaml ]; then uv run --no-sources pre-commit-install; fi`

PR #464, opened 2025-07-01
Filter out unneeded files/directories before deploying workers

EricGustin:
`arcade deploy` is failing for local packages that have large unneeded files such as `uv.lock`. It is failing because it is taking too long for the CLI to compress and PUT to the cloud.

### DIFF
--- a/libs/arcade-cli/arcade_cli/deployment.py
+++ b/libs/arcade-cli/arcade_cli/deployment.py
@@ -226,6 +226,24 @@ class Worker(BaseModel):
         if self.local_source is None:
             return None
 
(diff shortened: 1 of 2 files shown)


libs/arcade-cli/arcade_cli/deployment.py
```diff
@@ -226,6 +226,24 @@ def compress_local_packages(self) -> list[LocalPackage] | None:
         if self.local_source is None:
             return None
 
+        def exclude_filter(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo | None:
```
EricGustin, line 229: lmk if you have more files/folders in mind that we should exclude!

sdreyer (approved):

PR #465, opened 2025-07-01
Add Simplified Linear Toolkit

shubcodes:
This PR adds a simplified Linear toolkit focused on core functionality:

## What's Included

**Tools:**
- `get_issue` - Get detailed information about Linear issues
- `get_teams` - Get team information and details

**Models:**
- `DateRange` enum with comprehensive date range support (TODAY, YESTERDAY, THIS_WEEK, LAST_WEEK, THIS_MONTH, LAST_MONTH, THIS_YEAR, LAST_YEAR, LAST_7_DAYS, LAST_30_DAYS)
- Timezone-aware datetime handling following Google toolkit patterns

## What's Simplified

This toolkit has been streamlined by removing:
- Cycles management tools
- Projects management tools  
- Users management tools
- Workflows management tools
- Corresponding tests and evaluations for removed features

## Quality Assurance

- All linting and formatting checks pass
- Comprehensive test coverage for included functionality
- Follows established patterns from Google toolkit


docker/docker.engine.yaml
```diff
@@ -133,6 +133,14 @@ auth:
       client_id: ${env:LINKEDIN_CLIENT_ID}
       client_secret: ${env:LINKEDIN_CLIENT_SECRET}
 
+    - id: arcade-linear
+      description: "The Linear provider for Arcade"
+      enabled: true
```
EricGustin:
```suggestion
      enabled: false
```

toolkits/linear/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025, shubcodes
```
EricGustin:
```suggestion
Copyright (c) 2025 Arcade
```

toolkits/linear/.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+files: ^./
```
EricGustin:
```suggestion
files: ^.*/linear/.*
```

toolkits/linear/arcade_linear/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""Linear Toolkit for Arcade AI - Simplified Linear API Integration"""
```
EricGustin:
```suggestion
"""Linear Toolkit for Arcade AI"""
```

toolkits/linear/arcade_linear/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+    get_teams,  # Get team information
+)
+
+__all__ = [
+    "DateRange",
+    "get_issue",
+    "get_teams",
+]
+
+__version__ = "1.0.0"
```
EricGustin:
```suggestion
```

toolkits/linear/arcade_linear/constants.py
```diff
@@ -0,0 +1,16 @@
...
+except ValueError:
+    LINEAR_MAX_CONCURRENT_REQUESTS = 3
+
+try:
+    LINEAR_MAX_TIMEOUT_SECONDS = int(os.getenv("LINEAR_MAX_TIMEOUT_SECONDS", 30))
+except ValueError:
+    LINEAR_MAX_TIMEOUT_SECONDS = 30
+
+# Time range mappings have been moved to DateRange enum in models.py
+# This provides more structured and consistent datetime handling
```
EricGustin:
```suggestion
```

toolkits/linear/arcade_linear/models.py
```diff
@@ -0,0 +1,194 @@
...
+
+from datetime import datetime, timedelta, timezone
+from enum import Enum
+from typing import Optional
+
+
+class DateRange(Enum):
+    """Date range enum for Linear datetime queries
+
+    Provides consistent datetime range handling similar to Google toolkit.
```
EricGustin:
```suggestion
    Provides consistent datetime range handling.
```

toolkits/linear/arcade_linear/tools/teams.py
```diff
@@ -0,0 +1,121 @@
...
+        "Defaults to None (all teams).",
+    ] = None,
+    include_archived: Annotated[
+        bool, "Whether to include archived teams in results. Defaults to False."
+    ] = False,
+    created_after: Annotated[
+        str | None,
+        "Filter teams created after this date. Can be:\n"
+        "- Relative date string (e.g. 'last month', 'this week', 'yesterday')\n"
+        "- ISO date string (e.g. '2024-01-01')\n"
```
EricGustin: suggestion: provide the structure of the expected date string ('YYY-MM-DD') instead of an example.

The reasoning behind this is that 
1. The language model would be 'anchored' by this old date, and 
2. We don't want 2024-01-01 to be in the tool's embedding if someone decides to embed their tools.

EricGustin: I think we need to be more specific about what is not supported by this tool or implement the support that is currently missing. For example, does it support timezone offset or is it always treating the ISO 8601 string as UTC? Are hours, minutes, seconds supported or is the minimal time unit 'day'?

If, for example, 'hours' are not supported, then we should strip off hours from the string if they are in the provided string. I called this tool first with `"2025-06-26T12:00:00.000Z"` and then again with `"2025-06-26T06:00:00.000Z"` and it yielded different results. So we need to ensure we're sanitizing the string such that unsupported parts are not influencing what is in the response.

----

With all that being said, I think the simplest approach is to either only support `DateRange` enum as the input type, or remove the `created_after` parameter entirely as I'm not sure of the value that listing teams based on creation date provides.

toolkits/linear/arcade_linear/tools/teams.py
```diff
@@ -0,0 +1,121 @@
...
+        "Defaults to None (all time).",
+    ] = None,
+    limit: Annotated[
+        int, "Maximum number of teams to return. Min 1, max 100. Defaults to 50."
+    ] = 50,
+    after_cursor: Annotated[
```
EricGustin: suggestion: rename this input parameter to `cursor` and then alter the response such that it returns the key `cursor` instead of the key `end_cursor` and don't return the `start_cursor`. It helps the llm when the cursor's name is the same.

toolkits/linear/arcade_linear/tools/teams.py
```diff
@@ -0,0 +1,121 @@
...
+    - Team basic information (name, key, description)
+    - Team members and their roles
+    - Team settings and configuration
+    - Team creation and status information
+    - Team hierarchy and relationships
+
+    When NOT to use this tool:
+    - Finding individual users across teams - Use get_users
+    - Getting user-specific assigned issues - Use get_assigned_issues
+    - Searching for issues within teams - Use search_issues with team filter
```
EricGustin: this applies to all tools:

I think providing a list of when to use this tool and when to not use this tool could be discouraging the language model from calling the tool for an edge case that the docstring doesn't explicitly mention. Additionally, the examples are messing with the embedding

toolkits/linear/arcade_linear/tools/teams.py
```diff
@@ -0,0 +1,121 @@
...
+
+    This tool is the primary way to get team information and answer "who is in team X" questions.
```
EricGustin: Did you have trouble with the LLM choosing this tool for a non-read operation? If not I'd recommend saving everyone's tokens and omitting this line

toolkits/linear/arcade_linear/tools/teams.py
```diff
@@ -0,0 +1,121 @@
...
+        "total_count": len(cleaned_teams),
+        "filters": {
+            "team_name": team_name,
+            "include_archived": include_archived,
+            "created_after": created_after,
+        },
+    }
+
+    # Add pagination info
+    if "pageInfo" in teams_response:
```
EricGustin: The following ended up in the tool response:
```
  "pagination": {
    "end_cursor": null,
    "has_next_page": false,
    "has_previous_page": false,
    "start_cursor": null
  },
```

I think in addition to checking if "pageInfo" is in `teams_response` we should also be checking if `has_next_page` is true

toolkits/linear/arcade_linear/tools/teams.py
```diff
@@ -0,0 +1,121 @@
...
+    clean_team_data,
+    parse_date_string,
+    validate_date_format,
+)
+
+
+@tool(requires_auth=OAuth2(id="arcade-linear", scopes=["read"]))
+async def get_teams(
+    context: ToolContext,
+    team_name: Annotated[
```
EricGustin, line 18: I'm thinking out loud here, but supporting partial matches feels more like a feature for search_teams than a get_teams tools

toolkits/linear/conftest.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+@pytest.fixture
+def generate_random_email() -> Callable[[str | None, str | None], str]:
+    def random_email_generator(name: str | None = None, domain: str | None = None) -> str:
+        name = name or generate_random_str()
+        domain = domain or f"{generate_random_str()}.com"
+        return f"{name}@{domain}"
+
+    return random_email_generator
```
EricGustin: for unit tests you usually want repeatability, not cryptographic strength. This introduces non-deterministic test failures which can be really hard to reproduce. I would instead seed your random generator so that it is deterministic.

For example,
```python
random.seed(1)
return random.randint()
```

toolkits/linear/evals/eval_issues.py
```diff
@@ -0,0 +1,106 @@
...
+        name="Get Issue Evaluation",
+        system_message=(
+            "You are an AI assistant with access to Linear tools. "
+            "Use them to help the user get detailed information about Linear issues."
+        ),
+        catalog=catalog,
+        rubric=rubric,
+    )
+
+    # Eval Prompt: Show me complete details for issue API-789
```
EricGustin: nit: no need to have these comments, the prompt is already in the `user_message`

toolkits/linear/evals/eval_teams.py
```diff
@@ -0,0 +1,156 @@
...
+        expected_tool_calls=[
+            ExpectedToolCall(
+                func=get_teams,
+                args={
+                    "created_after": "last month",
+                },
+            ),
+        ],
+        critics=[
+            SimilarityCritic(critic_field="created_after", weight=1.0),
```
EricGustin: i think this should be BinaryCritic. For example, "lastmonth" would cause the tool to fail, but the eval would give it a great score. Applies to the cases below too

toolkits/linear/evals/eval_teams.py
EricGustin, line 1: I'd like to see at least one `add_case` that utilizes the `additional_messages` parameter. Cases that use this parameter are usually more adversarial and interesting

toolkits/linear/arcade_linear/tools/teams.py
shubcodes: after convo with @EricGustin -- thinks its okay because theirs such little content returned by this tool the llm will be able to parse and understand response

toolkits/linear/conftest.py
shubcodes: for emails have a hardcoded list of emails so that it does not break -- will do

shubcodes: write a generator that returns a list of hardcoded emails

shubcodes:
@EricGustin comments fixed

toolkits/linear/arcade_linear/tools/teams.py
```diff
@@ -0,0 +1,107 @@
...
+        "total_count": len(cleaned_teams),
+        "filters": {
+            "team_name": team_name,
+            "include_archived": include_archived,
+            "created_after": created_after,
+        },
+    }
+
+    # Add pagination info
+    if "pageInfo" in teams_response and teams_response["pageInfo"].get("has_next_page"):
```
EricGustin:
```suggestion
    if "pageInfo" in teams_response and teams_response["pageInfo"].get("hasNextPage"):
```

PR #471, opened 2025-07-01
Fix existing Slack test & bump TK versions

EricGustin:


toolkits/slack/arcade_slack/utils.py
```diff
@@ -371,7 +371,8 @@ async def paginate_loop() -> list[Any]:
 
     try:
         results = await asyncio.wait_for(paginate_loop(), timeout=max_pagination_timeout_seconds)
-    except TimeoutError:
+    # asyncio.TimeoutError for Python <= 3.10, TimeoutError for Python >= 3.11
+    except (TimeoutError, asyncio.TimeoutError):
```
EricGustin, line 375: This was the existing test that was failing because the action uses Python 3.10, but the exception handling was only for Python 3.11+

sdreyer (approved):

PR #475, opened 2025-07-04
adds a bunch of opentelemetry dependencies Mastra needs that aren't in the root package.json

rachelnabors:
Was livestreaming and @torresmateo came by and we weren't able to get this example working. Eventually I got  this working, no problem: https://docs.arcade.dev/home/mastra/use-arcade-tools After some comparison, it looks like Mastra has opentelemetry dependencies that aren't in the root package.json. I added them and resolve the problem, but it's still weird!

### DIFF
--- a/examples/mastra/package.json
+++ b/examples/mastra/package.json
@@ -18,6 +18,15 @@
     "@mastra/core": "latest",
     "@mastra/libsql": "latest",
     "@mastra/memory": "latest",
+    "@opentelemetry/auto-instrumentations-node": "^0.59.0",
+    "@opentelemetry/core": "^2.0.1",
+    "@opentelemetry/exporter-trace-otlp-grpc": "^0.201.0",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.201.0",
+    "@opentelemetry/instrumentation": "^0.202.0",
+    "@opentelemetry/resources": "^2.0.1",
+    "@opentelemetry/sdk-node": "^0.201.0",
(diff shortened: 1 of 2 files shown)


Spartee:
Are these required? @sdserranog is this not a dev dep?

evantahler:
I just gave our integration guide (https://docs.arcade.dev/home/mastra/use-arcade-tools) a try and did not run into any open telemetry problems.  

I have a *suspicion* that perhaps some of your dependencies might have been installed with `npm` and others with another package manager like `pnpm`, which could cause a problem like this to occur.  Please be sure to stick with a single pacakge manager, e.g. if you want pnmp, the intial setup of the project should also be with `pnpx create-mastra@latest my-arcade-agent`

I can confirm that `@mastra/core` properly lists the opentelemetry packages as dependencies, and you can confirm this with something like `npm list @opentelemetry/core`

```
❯ npm list @opentelemetry/core

my-arcade-agent@1.0.0 /private/tmp/mastra-demo/my-arcade-agent
├─┬ @mastra/core@0.10.10
│ ├─┬ @opentelemetry/auto-instrumentations-node@0.59.0
│ │ ├── @opentelemetry/core@2.0.1 deduped
```

Screenshot 2025-07-06 at 3 13 53 PM

nearestnabors:
Good point. Perhaps I was sloppy. Did a fresh clone, followed README again, problem didn't reoccur. Will reopen if it happens again. Cheers!

PR #477, opened 2025-07-08
added ruff toml and pre-comit files to template if community

torresmateo:
This reintroduces a question into `arcade new`, which adds the ruff and pre-commit files into new toolkits that are aimed to be contributed back to Arcade AI. 

I use it in the toolkit building tutorial

### DIFF
--- a/libs/arcade-cli/arcade_cli/new.py
+++ b/libs/arcade-cli/arcade_cli/new.py
@@ -97,11 +97,7 @@ def create_ignore_pattern(include_evals: bool, community_toolkit: bool) -> re.Pa
         patterns.append("evals")
 
     if not community_toolkit:
-        patterns.extend([
-            ".ruff.toml",
-            ".pre-commit-config.yaml",
-            "README.md"
-        ])
+        patterns.extend([".ruff.toml", ".pre-commit-config.yaml", "README.md"])
 
     return re.compile(f"({'|'.join(patterns)})$")
 


libs/arcade-cli/arcade_cli/new.py
```diff
@@ -174,6 +177,16 @@ def create_new_toolkit(output_directory: str, toolkit_name: str) -> None:
         "Do you want an evals directory created for you?", default=True
     )
 
+    cwd = Path.cwd()
+    # TODO: this detection mechanism works only for people that didn't change the
+    # name of the repo, a better detection method is required here
```
evantahler, line 182: I think this is fine for now - forks won't change the name of the repo by default

evantahler: An alternative approach would be to check in a file at the root of the repo with something very unique and specific you could check for, e.g. a string within a pyproject.toml

libs/arcade-cli/arcade_cli/new.py
```diff
@@ -174,6 +177,16 @@ def create_new_toolkit(output_directory: str, toolkit_name: str) -> None:
...
+    community_toolkit = False
+    if cwd.name == "toolkits" and cwd.parent.name == "arcade-ai":
+        community_toolkit = ask_yes_no_question(
+            "Is your toolkit a community contribution (to be merged into Arcade's repo)?",
```
evantahler:
```suggestion
            "Is your toolkit a community contribution (to be merged into Arcade's `arcade-ai` repo)?",
```

EricGustin: if their current working directory is .*/arcade-ai/toolkits then can't we already assume this is going to be a contribution to arcade-ai/toolkits?

libs/arcade-cli/arcade_cli/new.py
EricGustin: another alternative is to inspect the .git folder for the remote repo.

libs/arcade-cli/arcade_cli/new.py
```diff
@@ -96,6 +96,9 @@ def create_ignore_pattern(include_evals: bool) -> re.Pattern[str]:
     if not include_evals:
         patterns.append("evals")
 
+    if not community_toolkit:
+        patterns.extend([".ruff.toml", ".pre-commit-config.yaml"])
```
EricGustin, line 100: also exclude the readme.md

EricGustin (changes requested):
Thank you! Can you also exclude the readme.md if not community contribution

EricGustin (approved):

PR #478, opened 2025-07-09
contributing the Mastodon toolkit

torresmateo:
This adds the Mastodon toolkit to Arcade. It is intended to match the functionality of Arcade's existing X toolkit, offering similar tools

### DIFF
--- a/toolkits/mastodon/README.md
+++ b/toolkits/mastodon/README.md
@@ -1,0 +1,26 @@
+<div style="display: flex; justify-content: center; align-items: center;">
+  <img
+    src="https://docs.arcade.dev/images/logo/arcade-logo.png"
+    style="width: 250px;"
+  >
+</div>
+
+<div style="display: flex; justify-content: center; align-items: center; margin-bottom: 8px;">
+  <img src="https://img.shields.io/badge/python-3.10+-blue.svg" alt="Python version" style="margin: 0 2px;">
+  <img src="https://img.shields.io/badge/license-MIT-green.svg" alt="License" style="margin: 0 2px;">
+  <img src="https://img.shields.io/pypi/v/arcade_mastodon" alt="PyPI version" style="margin: 0 2px;">
+</div>
+
+
+<br>
+<br>
+
+# Arcade mastodon Toolkit
+Allow the agent to interact with a Mastodon server
+## Features
+
+- The mastodon toolkit does not have any features yet.
+
+## Development
+
+Read the docs on how to create a toolkit [here](https://docs.arcade.dev/home/build-tools/create-a-toolkit)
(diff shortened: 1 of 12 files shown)


evantahler (approved):
pre-approve

toolkits/mastodon/README.md
EricGustin, line 1: you can remove this file (also the PR you recently merged will make sure this won't be added to future toolkits, so thanks!)

toolkits/mastodon/pyproject.toml
```diff
@@ -0,0 +1,54 @@
+[build-system]
+requires = [ "hatchling",]
+build-backend = "hatchling.build"
+
+[project]
+name = "arcade_mastodon"
+version = "0.1.0"
+description = "Allow the agent to interact with a Mastodon server"
+requires-python = ">=3.10"
+dependencies = [
```
EricGustin, line 10: don't forget to add httpx as a dependency!

toolkits/mastodon/arcade_mastodon/tools/statuses.py
```diff
@@ -0,0 +1,173 @@
...
+@tool(
+    requires_auth=OAuth2(
+        id="mastodon",
+        scopes=["write:statuses"],
+    ),
+    requires_secrets=["MASTODON_SERVER_URL"],
+)
+async def post_status(
+    context: ToolContext,
+    status: Annotated[str, "The status to post"],
```
EricGustin, line 21: suggestion: might be worth mentioning that a 'status' is the text content of a user's post. Idk, worth a shot! Personally I thought status was 'online', 'offline', etc., so adding that context could be helpful.

toolkits/mastodon/arcade_mastodon/tools/statuses.py
```diff
@@ -0,0 +1,173 @@
...
+async def search_recent_statuses_by_keywords(
+    context: ToolContext,
+    keywords: Annotated[list[str] | None, "The keywords to search for."] = None,
+    phrases: Annotated[list[str] | None, "The phrases to search for."] = None,
+    limit: Annotated[
+        int, "The maximum number of statuses to return. Default is 20, maximum is 40."
+    ] = 20,
+) -> Annotated[dict, "The statuses from Mastodon"]:
+    """
+    Search for recent statuses by keywords and phrases.
```
EricGustin, line 140: this would be a good place to explain the relationship between keywords & phrases. Are they AND'd, OR'd? Does it match if only one keyword matches, or does it need all?

Also, you mentioned that this toolkit used X as inspiration. X is one of our oldest toolkits, but I recall having the keywords/phrases was something I regretted because language models were really bad at understanding the difference between the two (I think the X toolkit might even be older than the eval framework!). It'd be super exciting to see if you're able to come up with a new & improved way to search for statuses

EricGustin:
@torresmateo is this PR still relevant or should we close?

torresmateo:
I'm closing it, I use mastodon as an example toolkit to build and add to Arcade, so it's useful to teach MCP server writing in support sessions.

PR #481, opened 2025-07-10
custom `callback_host` for arcade login

evantahler:
Say you are @shubcodes and running arcade on replit or vs codespaces.  Today, `arcade login` assumes that the browser you've opened for the auth flow's "localhost" is the same that is running the `arcade login` command.  If you are running on one of these remote code execution environments, that won't be true.

Usage:
```
arcade login --callback-host "https://replit.com:9999/path/to/my/codespace"
```

### DIFF
--- a/libs/arcade-cli/arcade_cli/main.py
+++ b/libs/arcade-cli/arcade_cli/main.py
@@ -99,6 +99,11 @@ def login(
         "--port",
         help="The port of the Arcade Cloud host (if running locally).",
     ),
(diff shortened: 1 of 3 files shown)


EricGustin (approved):

evantahler:
@shubcodes - were you able to try this out... did it help?

shubcodes:
Need to retest and maybe also put in logic for device code auth for headless ssh sessions that don't support callback to localhost

PR #485, opened 2025-07-13
Remove arcade_ prefix requirement and add entry point toolkit discovery

EricGustin:
## Summary
This PR removes the requirement that all toolkits must have the arcade_ prefix and introduces a more flexible toolkit discovery system using Python entry points.

### 🏷️ Flexible Toolkit Naming
* Community toolkits: Only add arcade_ prefix when the user is in arcade-ai/toolkits/ directory and explicitly chooses to create a community contribution.
* External toolkits: No prefix requirement - developers can name their toolkits however they want
* Toolkit names are now determined by user choice rather than enforced automatically
### 🔍 Entry Point Discovery
* Added find_arcade_toolkits_from_entrypoints() method to discover toolkits via entry points
* Entry point group: arcade_toolkits with name: toolkit_name
* Updated pyproject.toml template to include entry point configuration
* Entry point discovery takes precedence over prefix-based discovery for deduplication
### 📦 Backward Compatibility
* Existing arcade_* prefixed toolkits continue to work via find_arcade_toolkits_from_prefix()
find_all_arcade_toolkits() now combines both discovery methods
* Deduplication logic prefers entry point toolkits over prefix-based ones when package names match
### 🛠️ `arcade new` Template Updates
* pyproject.toml template  for `arcade new` now includes entry point configuration: [project.entry-points.arcade_toolkits]
### 🔧 Minor Improvements
* Refactored _strip_arcade_prefix() into a separate method for reusability
* Updated variable naming for clarity (community_toolkit → is_community_toolkit)
### Benefits
* Developer Freedom: Toolkit developers are no longer forced to use the arcade_ prefix. They are also no longer forced to use the package name as the toolkit name.
* Cleaner Naming: External toolkits can use more natural names (e.g., my_company_toolkit instead of arcade_my_company_toolkit)
* Better Discovery: Entry points provide a more standard Python mechanism for plugin discovery
* Flexible Distribution: Toolkits can be distributed with any package name while still being discoverable
### Testing
* Added comprehensive tests for the new entry point functionality
* Tests cover edge cases like deduplication, error handling, and backward compatibility
### Version Bumps
arcade-core: 2.0.0 → 2.1.0
arcade-ai: 2.0.5 → 2.1.0

This change makes the Arcade toolkit ecosystem more flexible and developer-friendly while maintaining full backward compatibility with existing toolkits.

### DIFF
--- a/libs/arcade-cli/arcade_cli/new.py
+++ b/libs/arcade-cli/arcade_cli/new.py
@@ -147,12 +147,8 @@ def create_new_toolkit(output_directory: str, toolkit_name: str) -> None:
     """Create a new toolkit from a template with user input."""
     toolkit_directory = Path(output_directory)
 
-    package_name = toolkit_name if toolkit_name.startswith("arcade_") else f"arcade_{toolkit_name}"
-
     # Check for illegal characters in the toolkit name
-    if re.match(r"^[a-z0-9_]+$", package_name):
-        toolkit_name = package_name.replace("arcade_", "", 1)
-
+    if re.match(r"^[a-z0-9_]+$", toolkit_name):
         if (toolkit_directory / toolkit_name).exists():
             console.print(f"[red]Toolkit '{toolkit_name}' already exists.[/red]")
             exit(1)
@@ -180,15 +176,16 @@ def create_new_toolkit(output_directory: str, toolkit_name: str) -> None:
     cwd = Path.cwd()
     # TODO: this detection mechanism works only for people that didn't change the
     # name of the repo, a better detection method is required here
-    community_toolkit = False
+    is_community_toolkit = False
     if cwd.name == "toolkits" and cwd.parent.name == "arcade-ai":
-        community_toolkit = ask_yes_no_question(
-            "Is your toolkit a community contribution (to be merged into Arcade's `arcade-ai` repo)?",
-            default=False,  # False for now to keep the default behavior
+        prompt = (
+            "Is your toolkit a community contribution (to be merged into "
+            "\x1b]8;;https://github.com/ArcadeAI/arcade-ai\x1b\\ArcadeAI/arcade-ai\x1b]8;;\x1b\\ repo)?"
         )
+        is_community_toolkit = ask_yes_no_question(prompt, default=False)
 
     context = {
-        "package_name": package_name,
+        "package_name": "arcade_" + toolkit_name if is_community_toolkit else toolkit_name,
         "toolkit_name": toolkit_name,
         "toolkit_description": toolkit_description,
         "toolkit_author_name": toolkit_author_name,
@@ -200,7 +197,7 @@ def create_new_toolkit(output_directory: str, toolkit_name: str) -> None:
         "arcade_ai_min_version": ARCADE_AI_MIN_VERSION,
         "arcade_ai_max_version": ARCADE_AI_MAX_VERSION,
         "creation_year": datetime.now().year,
-        "community_toolkit": community_toolkit,
+        "community_toolkit": is_community_toolkit,
     }
     template_directory = Path(__file__).parent / "templates" / "{{ toolkit_name }}"
 
@@ -210,7 +207,7 @@ def create_new_toolkit(output_directory: str, toolkit_name: str) -> None:
     )
 
     # Create dynamic ignore pattern based on user preferences
-    ignore_pattern = create_ignore_pattern(include_evals, community_toolkit)
+    ignore_pattern = create_ignore_pattern(include_evals, is_community_toolkit)
 
     try:
         create_package(env, template_directory, toolkit_directory, context, ignore_pattern)
--- a/libs/arcade-cli/arcade_cli/templates/{{ toolkit_name }}/pyproject.toml
+++ b/libs/arcade-cli/arcade_cli/templates/{{ toolkit_name }}/pyproject.toml
@@ -36,6 +36,10 @@ dev = [
     "ruff>=0.7.4,<0.8.0",
 ]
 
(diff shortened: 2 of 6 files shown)


libs/tests/core/test_toolkit.py
```diff
@@ -0,0 +1,344 @@
...
+        toolkits = Toolkit.find_all_arcade_toolkits()
+        assert toolkits == []
+
+
+class TestEntryPointCompatibility:
+    """Test compatibility scenarios for entry point discovery."""
+
+    @patch("arcade_core.toolkit.importlib.metadata.entry_points")
+    @patch("arcade_core.toolkit.Toolkit.from_entrypoint")
+    def test_duplicate_toolkit_names_in_entrypoints(self, mock_from_ep, mock_entry_points):
```
EricGustin, line 243: question for the reviewer(s):

The behavior is that we now allow for more than 1 toolkit to have the same name as long as their package names are different. Do we want this behavior?
This means we could have something like:
```
`Github.ListStargazers`  # from the official arcade-github package
`Github.ListIssues` # from a community package named joe-shmoe-github
```

EricGustin: After thinking about this more, I believe this was always possible but required a multi-worker setup.

libs/arcade-cli/arcade_cli/new.py
```diff
@@ -200,7 +197,7 @@ def create_new_toolkit(output_directory: str, toolkit_name: str) -> None:
         "arcade_ai_min_version": ARCADE_AI_MIN_VERSION,
         "arcade_ai_max_version": ARCADE_AI_MAX_VERSION,
         "creation_year": datetime.now().year,
-        "community_toolkit": community_toolkit,
+        "community_toolkit": is_community_toolkit,
```
torresmateo:
```suggestion
        "is_community_toolkit": is_community_toolkit,
```

libs/arcade-cli/arcade_cli/templates/{{ toolkit_name }}/pyproject.toml
```diff
@@ -36,6 +36,10 @@ dev = [
     "ruff>=0.7.4,<0.8.0",
 ]
 
+# Tell Arcade.dev that this package is a toolkit
+[project.entry-points.arcade_toolkits]
+toolkit_name = "{{ package_name }}"
```
torresmateo, line 41: Can authors of non-community toolkits safely ignore this? If _not_, I'd love to include this condition:
 ```suggestion
{% if is_community_toolkit -%}
# Tell Arcade.dev that this package is a toolkit
[project.entry-points.arcade_toolkits]
toolkit_name = "{{ package_name }}"
{% endif %}
```

File: libs/arcade-core/arcade_core/toolkit.py [hunk 4]
```diff
@@ -110,31 +117,118 @@ def from_package(cls, package: str) -> "Toolkit":
...
         import sysconfig
 
-        # Get the site-packages directory of the current interpreter
+        toolkits = []
         site_packages_dir = sysconfig.get_paths()["purelib"]
+
         arcade_packages = [
             dist.metadata["Name"]
             for dist in importlib.metadata.distributions(path=[site_packages_dir])
             if dist.metadata["Name"].startswith("arcade_")
```
torresmateo, line 194: Is this likely to choke on things like "arcade-core"? it will the exception below handle it gracefully?

libs/tests/core/test_toolkit.py
torresmateo: Yes, I myself ran into some shadowing problems by running my local workers and registering to the cloud. I don't think there's an issue of having two toolkits with the same name. I think we should have some warnings or logs or notifications for the developer that there are overlapping _tools_ (toolkit.tool) in his tenant, and to either select which one has priority or only enable one. But I don't think that belong in this PR

libs/tests/core/test_toolkit.py
```diff
@@ -0,0 +1,344 @@
...
+        assert toolkit.name == "example"
+        assert toolkit.package_name == "arcade-example"
+
+
+class TestToolkitIntegration:
+    """Integration tests for toolkit discovery and loading."""
+
+    @patch("arcade_core.toolkit.Toolkit.find_arcade_toolkits_from_entrypoints")
+    @patch("arcade_core.toolkit.Toolkit.find_arcade_toolkits_from_prefix")
+    def test_mixed_toolkit_sources(self, mock_find_prefix, mock_find_ep):
```
torresmateo, line 307: Nit: could we add a mocked package with the correct prefix but that is not a toolkit?

libs/arcade-cli/arcade_cli/templates/{{ toolkit_name }}/pyproject.toml
EricGustin: The entrypoint is for everyone

libs/arcade-core/arcade_core/toolkit.py
EricGustin: This was existing logic. But the reason why it does not 'choke' on things like "arcade-core" is because the package starts with `arcade-`, not `arcade_`

torresmateo: In that case, does the exception below this code handle packages that are not toolkits but have the correct prefix? (I think it does just want to confirm)

EricGustin: Yeah, so that would be a `Toolkit` that has an empty internal tools list which is ok and is essentially ignored.

torresmateo (approved):

PR #495, opened 2025-07-16
Skip tests that require secrets in forks

evantahler:
We know that forks won't have our secrets, so we separate those toolkits out to new a new test group and skip them if you aren't in this main repo.  We know if a test suite has a secret via a magic comment

### DIFF
--- a/.github/workflows/test-toolkits.yml
+++ b/.github/workflows/test-toolkits.yml
@@ -30,6 +30,43 @@ jobs:
           echo "toolkits_without_secrets=$TOOLKITS_WITHOUT_SECRETS" >> $GITHUB_OUTPUT
           echo "toolkits_with_secrets=$TOOLKITS_WITH_SECRETS" >> $GITHUB_OUTPUT
 
+  test-toolkits:
+    needs: setup
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        toolkit: ${{ fromJson(needs.setup.outputs.toolkits_without_secrets) }}
+      fail-fast: true
+    steps:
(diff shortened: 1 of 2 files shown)


toolkits/postgres/tests/test_postgres.py
```diff
@@ -1,3 +1,5 @@
+# arcade-ci:requires-secret:TEST_POSTGRES_DATABASE_CONNECTION_STRING
+
```
evantahler: This is the magic comment that "toolkits that require secrets" greps for

.github/workflows/test-toolkits.yml
```diff
@@ -48,7 +48,31 @@ jobs:
...
+            echo "No tests found for toolkit ${{ matrix.toolkit }}, skipping..."
+            exit 0
+          elif [ "${EXIT_CODE:-0}" -ne 0 ]; then
+            exit ${EXIT_CODE}
+          fi
+
+      - name: Test toolkits that require secrets
+        if: |
+          contains(needs.toolkits_with_secrets.outputs.toolkits_with_secrets, matrix.toolkit) &&
+          github.repository == 'arcade-ai/arcade-ai'
```
EricGustin:
```suggestion
          github.repository == 'ArcadeAI/arcade-ai'
```

toolkits/postgres/tests/test_postgres.py
EricGustin: If we have to hardcode this comment then could the test-toolkits.yml just have a hard coded list of toolkit names instead? Feels like that would be one less step of indirection.

PR #499, opened 2025-07-17
[ENG-1537][Jira Toolkit] Adding List Sprints ArcTool + List Boards ArcTool

jottakka:
# 🎯 Jira Toolkit: Sprint Boards & Sprint Planning Tools

[ENG-1537](https://app.clickup.com/t/9014390315/ENG-1537)

## 📋 Overview

This PR introduces comprehensive sprint management capabilities to the Jira Toolkit, enabling users to interact with Jira boards and sprint planning functionality through new tools and enhanced API support.

### 🏃‍♂️ Sprint Planning Tools
- **Sprint Listing**: List all sprints for given boards with filtering and pagination capabilities

### 📊 Board Management Tools  
- **Board Listing**: List Jira boards with filtering and pagination capabilities

### 🔧 API Enhancements
- **Multi-version API Support**: Enhanced Jira client to support both Agile v1 and v3 REST APIs
- **Flexible API Client**: Optional enum type parameter with v3 as default

### 📁 New Files Added
- `toolkits/jira/arcade_jira/tools/boards.py` - Board management tools 
- `toolkits/jira/arcade_jira/tools/sprint_planning.py` - Sprint planning tools 
- Enhanced `toolkits/jira/arcade_jira/utils.py` - Utility functions 
- Updated `toolkits/jira/arcade_jira/client.py` - API client improvements
- New constants in `toolkits/jira/arcade_jira/constants.py`

### 🧪 Comprehensive Testing
- `test_boards.py` - 4board functionality tests
- `test_sprint_planning.py` - sprint planning tests  
- `test_utils.py` - new utility function tests

### 📊 Evaluation Framework
- `evals/boards/eval_get_boards.py` - Board listing evaluation suite 
- `evals/sprint_planning/eval_list_sprints.py` - Sprint Listing evaluation suite

## 🧪 Testing

All new functionality includes:
- ✅ Unit tests with comprehensive coverage
- ✅ Error handling validation


toolkits/jira/arcade_jira/tools/boards.py
```diff
@@ -0,0 +1,173 @@
...
+) -> Annotated[
+    dict[str, Any],
+    "A comprehensive dictionary containing successfully resolved boards in a 'boards' list, "
+    "with each board including metadata like ID, name, type, and location information. "
+    "Any boards that couldn't be found are listed in an 'errors' array with detailed "
+    "error messages and suggestions for available boards.",
+]:
+    """
+    Resolve multiple Jira boards by their names or IDs.
+    Returns successfully found boards and error details for any boards that couldn't be resolved.
```
EricGustin: opinion & nit: "resolve" seems more like a state change where as saying "Get multiple Jira..." falls more in line with what the tool is doing.

toolkits/jira/arcade_jira/tools/boards.py
```diff
@@ -0,0 +1,173 @@
...
+        "boards": [],
+        "errors": [],
+    }
+
+    # Get all available boards for name matching fallback
+    available_boards_response = None
+
+    for board_identifier in board_identifiers:
+        try:
+            # Try by ID first
```
EricGustin: per your annotation for `board_identifiers`, it seems that we can programmatically determine whether the current `board_identifier` is an ID or not by just checking if its numerical instead of calling the API to check. This would reduce the number of API calls in this tool by at most `len(board_identifier)` API calls.

Make sure you account for the rare case of a board name being numerical.

toolkits/jira/arcade_jira/tools/boards.py
```diff
@@ -0,0 +1,173 @@
...
+            except Exception:
+                # Board ID lookup failed, will try by name instead
+                board_found_by_id = False
+
+            if board_found_by_id:
+                continue
+
+            # Try by name (list all boards and match)
+            if available_boards_response is None:
+                available_boards_response = await _list_all_boards_simple(
```
EricGustin: Getting the list of all boards should be done outside of the for-loop. This would reduce the number of API calls by at most `len(board_identifiers)`

toolkits/jira/arcade_jira/tools/boards.py
```diff
@@ -0,0 +1,173 @@
...
+    create_board_error_message,
+    create_board_result_dict,
+    create_error_entry,
+    find_board_by_name,
+    validate_board_limit,
+)
+
+
+@tool(requires_auth=Atlassian(scopes=["read:board-scope:jira-software", "read:project:jira"]))
+async def get_boards_by_ids_or_names(
```
EricGustin: please add deduplication logic to this tool. For example, if I have a board named 'Engineering' with ID 123 and I call the tool as `get_boards_by_ids_or_name(board_identifiers=["Engineering", "123"])` then the tool shouldn't return the same board twice.

EricGustin: opinion: if there isn't a way to get a tool outside of its ID or Name, then I would suggest renaming to `get_boards`. Up to you though.

toolkits/jira/arcade_jira/tools/boards.py
```diff
@@ -0,0 +1,173 @@
...
+    context: ToolContext,
+    board_identifiers: Annotated[
+        list[str],
```
EricGustin: I'd suggest changing `board_identifiers`'s type to `list[str] | None`. If `board_identifiers` is _Falsey_, then return all boards up to a `limit`. `limit`  & `cursor` can be new input parameters. This means you'll be able to get rid of the 'list_all_boards' tool

toolkits/jira/arcade_jira/tools/boards.py
```diff
@@ -0,0 +1,173 @@
...
+                # Board ID lookup failed, will try by name instead
+                board_found_by_id = False
+
+            if board_found_by_id:
+                continue
+
+            # Try by name (list all boards and match)
+            if available_boards_response is None:
+                available_boards_response = await _list_all_boards_simple(
+                    context, client, limit=100
```
EricGustin: We should either implement pagination or add a TODO & a note in the tool description that this tool will sometimes not work correctly for workspaces that have more than 100 boards.

toolkits/jira/arcade_jira/tools/sprint_planning.py
```diff
@@ -0,0 +1,243 @@
...
+        "For example, offset=5 would skip the first 5 sprints on each board. "
+        "Must be 0 or greater. Defaults to 0 (start from the most recent sprints).",
+    ] = 0,
+    state: Annotated[
+        str | None,
+        "Filter sprints by their current state. Valid values are: "
+        "'future' (not started), 'active' (currently running), 'closed' (completed). "
+        "You can specify multiple states as a comma-separated list (e.g., 'active,future'). "
+        "If not provided, returns sprints in all states. Example: 'active' or 'active,future'",
+    ] = None,
```
EricGustin: This is the perfect opportunity to create a `SprintState` enum and use it for the input param's type. Leave 0 chance of an LLM hallucinating an invalid state and saves you from validating the input string inside the body of the function

toolkits/jira/arcade_jira/client.py
```diff
@@ -17,6 +17,10 @@ class JiraClient:
     base_url: str = JIRA_BASE_URL
     api_version: str = JIRA_API_VERSION
     max_concurrent_requests: int = JIRA_MAX_CONCURRENT_REQUESTS
+    # If True, use the Agile API (v1) instead of the REST API
+    # This is required for some operations, such as listing sprints for a board
+    # https://developer.atlassian.com/cloud/jira/software/rest/api-group-sprints/#api-agile-1-0-board-boardid-sprint-get
+    use_agile_api: bool = False
```
EricGustin: From what I can tell there are many more APIs beyond the two that are supported by this client so far. So I think having a toggle like this boolean might not be the most extensible.

Here are two other options:
1. Consider going the route of `JiraClient` having an `api_type` (Enum?) field and an `api_version` (Enum?) field. 
2. Have a base `JiraClient` that the two can inherit from.

toolkits/jira/arcade_jira/tools/sprint_planning.py
```diff
@@ -0,0 +1,410 @@
...
+        "Filters sprints that overlap with or occur after this date. "
+        "Can be used together with end_date to create a date range. "
+        "Example: '2024-01-01'",
+    ] = None,
+    end_date: Annotated[
+        str | None,
+        "Optional end date filter for sprints in YYYY-MM-DD format. "
+        "Filters sprints that overlap with or occur before this date. "
+        "Can be used together with start_date to create a date range. "
+        "Example: '2024-03-31'",
```
EricGustin: > This comment generally applies to all input parameters for tools

Try not to include the example date `2024-03-31`. The reasoning behind this is that the language model would be 'anchored' by this old date, and we don't want 2024-01-01 to be in the tool's embedding if someone decides to embed their tools.

toolkits/jira/arcade_jira/tools/sprint_planning.py
```diff
@@ -0,0 +1,410 @@
...
+    client = JiraClient(context.get_auth_token_or_empty(), use_agile_api=True)
+
+    results = {
+        "boards": [],
+        "sprints_by_board": {},
+        "errors": [],
+    }
+
+    # Process each board ID
+    for board_id in board_ids:
```
EricGustin, line 155: if we instead did

```py
boards = await get_boards(context, board_ids)
for board in boards:
    await _process_single_board(
            client,
            board,
            offset,
            sprints_per_board,
            state,
            start_date,
            end_date,
            specific_date,
            results,
        )
```

then we would only need to call the `get_boards` tool once instead of `len(board_ids)` times

toolkits/jira/arcade_jira/tools/sprint_planning.py
```diff
@@ -0,0 +1,410 @@
...
+    "A comprehensive dictionary containing sprint data for the specified boards. "
+    "Includes a 'boards' list with board metadata, 'sprints_by_board' dictionary mapping "
+    "board IDs to their sprint data, and an 'errors' list for any boards that "
+    "don't support sprints or couldn't be processed. Sprints are sorted with latest first.",
+]:
+    """
+    List sprints for Jira boards by name or ID with advanced filtering and pagination.
+    Requires a list of board identifiers and supports date-based filtering.
+    Returns valid responses for sprint boards and error messages for non-sprint boards.
+    """
```
EricGustin: Depending on how evals go, I'd suggest specifying the relationships between specific_date, start_date, and end_date since their relationship is complex.

toolkits/jira/arcade_jira/tools/sprint_planning.py
```diff
@@ -0,0 +1,410 @@
...
+        results["errors"].append(error_entry)
+
+
+async def _try_fetch_sprints_and_determine_type(
+    client: JiraClient, board_info: dict[str, Any], params: dict[str, Any]
+) -> tuple[bool, str, dict[str, Any] | None]:
+    """
+    Try to fetch sprints for a board and determine if it supports sprints.
+    Returns (supports_sprints, final_board_type, response).
+    """
```
EricGustin, line 403: typically we try to have 
```
"""
    Args:
        arg_name: arg_description
    Returns:
        return_description
    Raises:
        exception_type: when_this_is_raised_description
"""
```
in the docstring for all non-tools that are complex enough.

toolkits/jira/arcade_jira/tools/sprint_planning.py
```diff
@@ -0,0 +1,410 @@
...
+
+
+@tool(
+    requires_auth=Atlassian(
+        scopes=[
+            "read:board-scope:jira-software",
+            "read:project:jira",
+            "read:sprint:jira-software",
+            "read:issue-details:jira",
+            "read:board-scope.admin:jira-software",
```
EricGustin: (not a blocker)

Can you add comments on the right side of each scope specifying what the scope is needed for (which endpoints). This is especially useful when a tool calls other tools or when a tool calls multiple endpoints. We haven't done a very good job at documenting this type of thing in the past, but I think it would be great if we started doing this by default.

EricGustin (commented):
Looking good! I like the effort so far. I know the Jira APIs can be annoying at times, so thanks for taking this on.

I have one more suggestion regarding the structure of the toolkit. For easier navigation & future refactoring, consider structuring the toolkit like this:

arcade_jira/
  - rest_api/
  - agile_api/

toolkits/jira/arcade_jira/client.py
jottakka: I believe the better approach would be having the calls centralized in the client and the tools only have "access" to the methods itself without needing to build things like URL paths by themselves. Changes like this would only affect the client file and no need to changing tool codes, as the would be unware of the http interactions needed. If each time we find peculiarity we need to change a bunch of files it can get crazy.

 
What you think of a refactor for that (not necessary in this PR)? May take some time, but can save a lot of work when adding new tools that might need some other kinds of arguments in the API. Maybe Cursor can do it pretty fast. Having a JiraClient.GetJiraSprints({what i need for getting sprints}) will be more cleaner overall than JiraClient.Get(path, arguments, some internal info).

Besides that, for that PR I think a Inherited Jira Client make senses, the versioning enum would required the tool code to have this knowledge about the jira api itself.

toolkits/jira/arcade_jira/tools/sprint_planning.py
jottakka: Makes sense! Usally C# enums are integers, but if I remember correctly one can do as string in python. I will do it !

toolkits/jira/arcade_jira/tools/sprint_planning.py
jottakka: Nice to know.

toolkits/jira/arcade_jira/client.py
EricGustin: Thanks for the thoughtful response. I think we might be slightly talking past each other though. My original point wasn’t about pushing API knowledge into the tools. I fully agree tools should only interact with high-level methods like get_jira_sprints() and not be concerned with paths, params, etc.

The concern I raised is about how **the client itself** selects which API to use. Right now it’s done with a boolean (`use_agile_api`), which works for two APIs but doesn’t scale well if we ever need to support others that Jira has. At that point a boolean becomes ambiguous and fragile.

For example, today you are doing this inside your tools
```py
client = JiraClient(context.get_auth_token_or_empty(), use_agile_api=True)
```

And I am suggesting you do this instead inside your tools
```py
client = JiraClient(context.get_auth_token_or_empty(), client_type=APIType.AGILE)
```

So there is no additional "api writing in the tools" than what there was before. Just more extensible for the future.

jottakka: I see, makes sense. I my comment is more about what we have there also, with higher levels calls for the client we woundn't even need this parameter, but I think it would take a lot of time to rebuild the client, this enum approach seems faster, maybe latter a refactor to move all the api calls to higher level methods would also be nice.

Spartee:
Task linked: [ENG-1537 Jira Additions: Sprints](https://app.clickup.com/t/9014390315/ENG-1537)

jottakka:
> Looking good! I like the effort so far. I know the Jira APIs can be annoying at times, so thanks for taking this on.
> 
> I have one more suggestion regarding the structure of the toolkit. For easier navigation & future refactoring, consider structuring the toolkit like this:
> 
> arcade_jira/
> 
> * rest_api/
> * agile_api/

I also think that it is a good ideia, but doing it in other PR seems to be a better approach, also some higher level of care will be needed as some features are already be used by clients.

PR #500, opened 2025-07-18
bugfix: URL expansion in Twitter

torresmateo:
This solves an edge case where the `"url"` entity does not have an `"expanded_url"` case when expanding. The behavior for now is simply don't replace the URL as it seems the display_url is NOT what we want to keep as it's normally abbreviated

### DIFF
--- a/toolkits/x/pyproject.toml
+++ b/toolkits/x/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "arcade_x"
-version = "0.1.2"
+version = "0.1.16"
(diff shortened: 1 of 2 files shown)


toolkits/x/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "arcade_x"
-version = "0.1.15"
+version = "0.1.2"
```
EricGustin:
```suggestion
version = "0.1.16"
```

EricGustin (approved):
nice catch

PR #505, opened 2025-07-22
Add basic pagination to `GoogleSheets.GetSpreadsheet` tool

EricGustin:
A customer reported `tool execution failed: GoogleSheets.GetSpreadsheet@2.0.0: response body is too large to decode"` when calling the tool for very large spreadsheets. Large amounts of data in the tool response is not the desired paradigm for language model tools.

This PR adds **basic** pagination for the `get_spreadsheet` tool in the `GoogleSheets` toolkit. This update restricts the number of rows returned to up to 1000 and the number of columns to up to 100 for each sheet in the spreadsheet. `start_col` and `start_row` parameters have been added to support paginating through a large spreadsheet if one truly desires to do so. Also, `sheet_identifier` input parameter was added so that the tool only returns data for a single sheet. The sheet can be identified by its name, its ID, or its position within the spreadsheet (e.g., the 1st, 2nd, etc).

Perhaps a tool in the future can more intelligently select the range of columns & rows based on the structure of the spreadsheet. That is not this tool though.

### DIFF
--- a/toolkits/google_sheets/arcade_google_sheets/tools/read.py
+++ b/toolkits/google_sheets/arcade_google_sheets/tools/read.py
@@ -6,7 +6,8 @@ from arcade_tdk.auth import Google
 from arcade_google_sheets.decorators import with_filepicker_fallback
 from arcade_google_sheets.utils import (
     build_sheets_service,
-    parse_get_spreadsheet_response,
+    get_spreadsheet_simple,
+    get_spreadsheet_with_pagination,
 )
 
 
@@ -20,23 +21,53 @@ from arcade_google_sheets.utils import (
 async def get_spreadsheet(
     context: ToolContext,
     spreadsheet_id: Annotated[str, "The id of the spreadsheet to get"],
+    max_rows_per_request: Annotated[
+        int, "Maximum rows to fetch per request (defaults to 1000)"
+    ] = 1000,
+    max_cols_per_request: Annotated[int, "Maximum columns to fetch per request (default: 26)"] = 26,
+    start_row: Annotated[int, "Starting row number (1-indexed, defaults to 1)"] = 1,
+    start_col: Annotated[str, "Starting column letter (defaults to 'A')"] = "A",
+    max_rows: Annotated[
+        int | None, "Maximum total rows to fetch (optional, default: None for all rows)"
+    ] = None,
+    max_cols: Annotated[
+        int | None, "Maximum total columns to fetch (optional, default: None for all columns)"
+    ] = None,
 ) -> Annotated[
     dict,
     "The spreadsheet properties and data for all sheets in the spreadsheet",
 ]:
     """
     Get the user entered values and formatted values for all cells in all sheets in the spreadsheet
-    along with the spreadsheet's properties
+    along with the spreadsheet's properties. Supports pagination for large spreadsheets.
+
+    For large spreadsheets, use pagination parameters to control memory usage:
+    - max_rows_per_request: Chunk size for rows per API request
+    - max_cols_per_request: Chunk size for columns per API request
+    - start_row/start_col: Starting position
+    - max_rows/max_cols: Optional limits on total data to fetch
     """
     service = build_sheets_service(context.get_auth_token_or_empty())
 
-    response = (
-        service.spreadsheets()
-        .get(
-            spreadsheetId=spreadsheet_id,
-            includeGridData=True,
-            fields="spreadsheetId,spreadsheetUrl,properties/title,sheets/properties,sheets/data/rowData/values/userEnteredValue,sheets/data/rowData/values/formattedValue,sheets/data/rowData/values/effectiveValue",
-        )
-        .execute()
+    is_using_defaults = (
+        max_rows_per_request == 1000
+        and max_cols_per_request == 26
+        and start_row == 1
+        and start_col == "A"
+        and max_rows is None
+        and max_cols is None
     )
-    return parse_get_spreadsheet_response(response)
+
+    if is_using_defaults:
+        return get_spreadsheet_simple(service, spreadsheet_id)
+    else:
+        return get_spreadsheet_with_pagination(
+            service,
+            spreadsheet_id,
+            max_rows_per_request,
+            max_cols_per_request,
+            start_row,
+            start_col,
+            max_rows,
+            max_cols,
+        )
--- a/toolkits/google_sheets/arcade_google_sheets/tools/read.py
+++ b/toolkits/google_sheets/arcade_google_sheets/tools/read.py
@@ -6,6 +6,7 @@ from arcade_tdk.auth import Google
 from arcade_google_sheets.decorators import with_filepicker_fallback
 from arcade_google_sheets.utils import (
     build_sheets_service,
+    enforce_data_size_limit,
     get_spreadsheet_with_pagination,
     process_get_spreadsheet_params,
 )
@@ -21,6 +22,12 @@ from arcade_google_sheets.utils import (
 async def get_spreadsheet(
     context: ToolContext,
     spreadsheet_id: Annotated[str, "The id of the spreadsheet to get"],
+    sheet_identifier: Annotated[
+        str,
+        "The identifier of the sheet to get. This can be the order it appears "
+        "in the spreadsheet (1-based), the sheet name, or the sheet id. Defaults to '1', "
+        "which represents the first sheet in the spreadsheet.",
+    ] = "1",
     start_row: Annotated[int, "Starting row number (1-indexed, defaults to 1)"] = 1,
     start_col: Annotated[
         str, "Starting column letter(s) or 1-based column number (defaults to 'A')"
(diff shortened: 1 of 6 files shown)


toolkits/google_sheets/arcade_google_sheets/tools/read.py
```diff
@@ -20,23 +21,39 @@
...
+    ] = "A",
+    max_rows: Annotated[
+        int,
+        "Maximum number of rows to fetch for each sheet in the spreadsheet. "
+        "Must be between 1 and 1000. Defaults to 1000.",
+    ] = 1000,
+    max_cols: Annotated[
+        int,
+        "Maximum number of columns to fetch for each sheet in the spreadsheet. "
+        "Must be between 1 and 26. Defaults to 26.",
```
evantahler: What an odd default!

toolkits/google_sheets/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "arcade_google_sheets"
-version = "2.0.0"
+version = "2.1.0"
```
evantahler: This /might/ be a breaking change (for large sheets) that previously would have returned all the cols and rows it could have loaded

evantahler (commented):
I /think/ this should be a breaking change release

toolkits/google_sheets/arcade_google_sheets/tools/read.py
EricGustin: Since sheets are created with 1000 rows and 26 columns (A-Z) by default, it works well for the "typical" case. Should I be more clear about this?

evantahler: This is a non-blocking nit, but my mental model is that paginating rows is easy to reason about, but paginating columns is a little stranger because you won't be getting the whole row and it will take some work to notice.  But I also get you need to have an upper bound on size.  So, I'd prefer a column limit much larger than the default (100?) which would include any new columns for a long time

EricGustin: So I want I really wanted to do was remove the max_rows and max_cols input parameters, and then the tool itself determines the dimensions to fetch with preference for columns. Maybe the total number of cells can be configurable.

Examples: A 25row X 1000 column sheet would be supported by a single tool call. Same with a 10,000 row X 2 column sheet.

EricGustin:
Updated to default to 100 columns.
Updated to only get a single sheet per tool call.
Updated to explicitly raise a ToolExecutionError if the data retrieved is still somehow > 10MB

toolkits/google_sheets/arcade_google_sheets/tools/read.py
```diff
@@ -20,23 +22,46 @@
 async def get_spreadsheet(
     context: ToolContext,
     spreadsheet_id: Annotated[str, "The id of the spreadsheet to get"],
+    sheet_identifier: Annotated[
+        str,
+        "The identifier of the sheet to get. This can be the order it appears "
+        "in the spreadsheet (1-based), the sheet name, or the sheet id. Defaults to '1', "
+        "which represents the first sheet in the spreadsheet.",
+    ] = "1",
```
byrro: I think this could create conflicts. I can see not-so-rare cases of spreadsheets having this sequence of sheets:

1. "Year Report Summary"
2. "Monthly Reports Summary"
3. "1"
4. "2"
5. "3"
6. ...

If there's a single sheet named with an integer, this could become an issue...

I'd say we can accept a sheet ID or name. Not its position.

EricGustin: I believe I handled this case in utils.py `find_sheet_by_identifier` by first attempting to treat it as a sheet name.

byrro:
Alert: incoming nerdy ex-economist review

toolkits/google_sheets/arcade_google_sheets/utils.py
```diff
@@ -546,3 +547,212 @@ def parse_write_to_cell_response(response: dict) -> dict:
...
+        if not start_col.isdigit():
+            raise ToolExecutionError("Input 'start_col' must be alphabetic")
+        processed_start_col = index_to_col(int(start_col) - 1)
+    else:
+        processed_start_col = start_col.upper()
+
+    return processed_start_row, processed_start_col, processed_max_rows, processed_max_cols
+
+
+def enforce_data_size_limit(data: dict) -> None:
```
byrro: Since this only raises an exception if the data is too large, I'd rename it to something like `raise_exception_for_large_payload`.

The name suggested me it would going to trim the data to "enforce" it to comply with a certain size threshold - which would probably not be a good idea :)

toolkits/google_sheets/arcade_google_sheets/tools/read.py
```diff
@@ -20,23 +22,46 @@
...
+    start_row: Annotated[int, "Starting row number (1-indexed, defaults to 1)"] = 1,
```
byrro, line 36: Have u considered accepting a sheet range? Like `"A5:G200"`. This is one of the most common ways people will refer to sheet row/col ranges.

Having the `start_col` and `start_row` are great for sequential pagination. Adding a `sheet_range` argument would cover other use cases.

If someone refers to `"AZ4:CJ30"`, for instance, LLMs could have a hard time converting to positional integers.

toolkits/google_sheets/arcade_google_sheets/utils.py
```diff
@@ -546,3 +547,212 @@ def parse_write_to_cell_response(response: dict) -> dict:
...
+    """
+    processed_start_row = max(1, start_row)
+    processed_max_rows = max(1, min(max_rows, 1000))
+    processed_max_cols = max(1, min(max_cols, 26))
+    if not start_col.isalpha():
+        if not start_col.isdigit():
+            raise ToolExecutionError("Input 'start_col' must be alphabetic")
+        processed_start_col = index_to_col(int(start_col) - 1)
+    else:
+        processed_start_col = start_col.upper()
```
byrro, line 756: Need to validate that all characters are within `string.ascii_uppercase`. It may come as non-ascii alpha chars that will pass the `str.isalpha` test, but won't be valid sheet column references.

byrro: or, better, validate that all chars are within `string.ascii_uppercase + string.ascii_lowercase` and then use `upper`, like you did already

toolkits/google_sheets/arcade_google_sheets/utils.py
```diff
@@ -546,3 +547,212 @@ def parse_write_to_cell_response(response: dict) -> dict:
...
+    if not target_sheet:
+        raise ToolExecutionError(
+            message=f"Sheet with identifier '{sheet_identifier}' not found",
+            developer_message=(
+                "Sheet(s) in the spreadsheet: "
+                + ", ".join([sheet.model_dump_json() for sheet in spreadsheet.sheets])
+            ),
+        )
+
+    a1_ranges = []
```
byrro, line 677: nit: I see you use this "a1" reference many times. Why not call it "ranges" and the function `calculate_sheet_range` instead of `calculate_a1_sheet_range`? the "a1" gives a false impression that the range (or the function response) starts from the "A1" cell position.

toolkits/google_sheets/arcade_google_sheets/utils.py
```diff
@@ -546,3 +547,212 @@ def parse_write_to_cell_response(response: dict) -> dict:
...
+    Args:
+        data (dict): The data to enforce the size limit on.
+
+    Raises:
+        ToolExecutionError:
+            If the data size exceeds 10MB
+    """
+    num_bytes = len(str(data).encode("utf-8"))
+
+    if num_bytes >= (10 * 1024 * 1024):
```
byrro, line 780: Should we make the max size configurable?

Regarding the default, I think 10 MB is a lot already. In LLM context, this will take in the ballpark of ~2 million tokens, which is more than most (all?) models can manage.

I'd say 50K tokens looks more reasonable. Popular models cap at around ~128K, so this would sit around ~40% of the max capacity. Model performance may degrade quickly after this level, so it's better to keep far below its theoretical limit.

50K tokens would translate to roughly ~200K chars, which is around ~0.2 MB.

toolkits/google_sheets/arcade_google_sheets/tools/read.py
byrro: I noticed that. But what if the spreadsheet has a sheet named "1" and the user says: "get the first sheet". The LLM will pass in the value "1" and you'll return the sheet named "1", but not necessarily it's the first sheet.

byrro: If accepting a sheet position is important, I'd say we have to break down in two arguments:

1. `sheet_id_or_name`
2. `sheet_position`

Then instruct the LLM to provide one of the two in the tool call.

toolkits/google_sheets/arcade_google_sheets/utils.py
EricGustin: A1 is a specific notation for ranges. I'm explicit about this so that a dev won't confuse it with R1C1 notation. https://developers.google.com/workspace/sheets/api/guides/concepts#cell

toolkits/google_sheets/arcade_google_sheets/tools/read.py
EricGustin: I considered this, but I think that would be better for a "get range" tool.

toolkits/google_sheets/arcade_google_sheets/utils.py
EricGustin: addressed in slack

PR #487, opened 2025-07-15
Support Tool Output in ValueSchema of ToolDefinition

Spartee:
### Tool: ``GoogleNews.SearchNewsStories``

```python
@tool(requires_secrets=["SERP_API_KEY"])
async def search_news_stories(
    context: ToolContext,
    keywords: Annotated[
        str,
        "Keywords to search for news articles. E.g. 'Apple launches new iPhone'.",
    ],
    country_code: Annotated[
        CountryCode | None,
        "2-character country code to search for news articles. "
        "E.g. 'us' (United States). "
        f"Defaults to '{DEFAULT_GOOGLE_NEWS_COUNTRY}'.",
    ] = None,
    language_code: Annotated[
        LanguageCode,
        "2-character language code to search for news articles. E.g. 'en' (English). "
        f"Defaults to '{DEFAULT_GOOGLE_NEWS_LANGUAGE}'.",
    ] = DEFAULT_GOOGLE_NEWS_LANGUAGE,
    limit: Annotated[
        int | None,
        "Maximum number of news articles to return. Defaults to None "
        "(returns all results found by the API).",
    ] = None,
) -> Annotated[dict[str, Any]]:
    """Search for news articles related to a given query."""
    ...
```

### Tool Definition: ``GoogleNews.SearchNewsStories``
```
  {
    "name": "SearchNewsStories",
    "fully_qualified_name": "GoogleNews.SearchNewsStories",
    "description": "Search for news articles related to a given query.",
    "toolkit": {
      "name": "GoogleNews",
      "description": "Arcade.dev LLM tools for getting new via Google News",
      "version": "2.0.0"
    },
    "input": {
      "parameters": [
        {
          "name": "keywords",
          "required": true,
          "description": "Keywords to search for news articles. E.g. 'Apple launches new iPhone'.",
          "value_schema": {
            "val_type": "string",
            "inner_val_type": null,
            "enum": null,
          },
          "inferrable": true
        },
        {
          "name": "country_code",
          "required": false,
          "description": "2-character country code to search for news articles. E.g. 'us' (United States). Defaults to 'None'.",
          "value_schema": {
            "val_type": "string",
            "inner_val_type": null,
            "enum": null,
          },
          "inferrable": true
        },
        {
          "name": "language_code",
          "required": false,
          "description": "2-character language code to search for news articles. E.g. 'en' (English). Defaults to 'en'.",
          "value_schema": {
            "val_type": "string",
            "inner_val_type": null,
            "enum": null,
          },
[...]

PR #506, opened 2025-07-22
Support for multiple Atlassian Clouds in the Jira Toolkit

byrro:
Adds `Jira.GetAvailableAtlassianClouds` tool, which provides a list of clouds available (checking which Clouds were actually authorized by the current auth token).

Refactors the interface of every tool to accept an `atlassian_cloud_id` argument (when not provided, try to get a unique cloud ID - if multiple are available, raises a Retryable error with the list of Clouds available instructing to select one).

Gets rid of all caching. Now storing the global semaphore to the context object. The global semaphore is important because some tools depend on others, and each tool instantiates its own Jira HTTP client. Storing the semaphore in the context object ensures that all HTTP clients will respect the concurrency limit.

Removes from tool responses the Atlassian URLs linking to objects in the Jira GUI (users, projects, issues, etc. We do not keep track of the cloud name anymore, which is required to build the objects' URLs.

Extends/refactors unit tests accordingly.

Evals checking LLM behavior when:

- a cloud ID is explicitly mentioned in the prompt;
- no cloud ID is mentioned;
- a "multiple clouds available" error is raised and the user is prompted to pick one;
- user request triggers another tool call after having previously picked a cloud ID (in the same chat context);

### DIFF
--- a/toolkits/jira/pyproject.toml
+++ b/toolkits/jira/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "arcade_jira"
-version = "0.2.0"
+version = "1.0.0"
 description = "Arcade.dev LLM tools for interacting with Atlassian Jira"
 requires-python = ">=3.10"
 dependencies = [


toolkits/jira/arcade_jira/tools/projects.py
```diff
@@ -22,10 +22,22 @@ async def list_projects(
...
+    atlassian_cloud_id = await resolve_cloud_id(context, atlassian_cloud_id)
     return cast(
-        dict[str, Any], await search_projects(context, keywords=None, limit=limit, offset=offset)
+        dict[str, Any],
+        await search_projects(
+            context=context,
+            keywords=None,
+            limit=limit,
+            offset=offset,
+        ),
```
EricGustin: Don't forget to provide the `atlassian_cloud_id` to `search_projects`

byrro: good catch

toolkits/jira/arcade_jira/utils.py
```diff
@@ -1123,22 +1190,84 @@ def extract_id(field: Any) -> dict[str, str] | None:
...
+            developer_message=message,
+        )
 
-    return f"https://{cloud_name}.atlassian.net/jira/software/projects/{project_key}/summary"
+    if len(clouds) > 1:
+        message = (
+            "Multiple Atlassian Clouds are available. One Cloud ID has to be selected and provided "
+            "in the tool call using the `atlassian_cloud_id` argument."
+        )
+        raise RetryableToolError(
```
EricGustin, line 1290: This is great. I was in the chat playground and said "list all of my projects across my workspaces" without providing any context about cloud_ids, and the model ended up taking the context from the error and calling `list_projects` once for each cloud_id (3 times).

toolkits/jira/arcade_jira/tools/users.py
```diff
@@ -56,9 +66,15 @@ async def list_users(
 async def get_user_by_id(
     context: ToolContext,
     user_id: Annotated[str, "The the user's ID."],
+    atlassian_cloud_id: Annotated[
```
EricGustin, line 69: don't forget to provide this to the client call on line 82

byrro:
> When testing out the tools I noticed that language models would sometimes provide the "name" instead of the "id" for the `atlassian_cloud_id` input parameter. If we could detect that a `cloud_name` was provided instead of a `cloud_id`, then it would be great to resolve that `cloud_name` to an `cloud_id`! This would make the handling of clouds much less fragile

Interesting. That sounds like a good idea, will work on this.

toolkits/jira/arcade_jira/tools/users.py
byrro: It's provided on line 77. It's not necessary to provide to the `get` method. The URL is built using the internal `cloud_id` property.

EricGustin: i see. `JiraClient._build_url` adds the cloud_id to the URL

toolkits/jira/arcade_jira/tools/attachments.py
```diff
@@ -40,11 +39,17 @@ async def attach_file_to_issue(
         "If the filename is not recognized, it will attach the file without specifying a type. "
         "Defaults to None (infer from filename or attach without type).",
     ] = None,
+    atlassian_cloud_id: Annotated[
+        str | None,
+        "The ID of the Atlassian Cloud to use (defaults to None). If not provided and the user has "
+        "a single cloud authorized, the tool will use that. Otherwise, an error will be raised.",
```
jottakka, line 45: Question: For things like this, informing a regex, masking or giving a example on how it looks like would bring some benefits?

toolkits/jira/arcade_jira/tools/attachments.py
```diff
@@ -108,26 +122,42 @@ async def list_issue_attachments_metadata(
...
 
 
 @tool(requires_auth=Atlassian(scopes=["read:jira-work"]))
 async def download_attachment(
     context: ToolContext,
     attachment_id: Annotated[str, "The ID of the attachment to download"],
+    atlassian_cloud_id: Annotated[
+        str | None,
+        "The ID of the Atlassian Cloud to use (defaults to None). If not provided and the user has "
+        "a single cloud authorized, the tool will use that. Otherwise, an error will be raised.",
```
jottakka, line 149: Is there some way to reuse tool input paramets s like this one, that will be use everywhere and one chance should also be aplied to all other entries?

toolkits/jira/arcade_jira/tools/cloud.py
```diff
@@ -0,0 +1,46 @@
...
+from arcade_tdk.auth import Atlassian
+
+from arcade_jira.constants import JIRA_MAX_CONCURRENT_REQUESTS
+from arcade_jira.utils import check_if_cloud_is_authorized
+
+
+@tool(requires_auth=Atlassian(scopes=["read:jira-user"]))
+async def get_available_atlassian_clouds(
+    context: ToolContext,
+) -> Annotated[dict[str, list[dict[str, str]]], "Available Atlassian Clouds"]:
```
jottakka, line 15: As even atlassian does not follow the naming convention for clouds, should it be the case to inform the other possible names that are used like clients?

jottakka (approved):
I added some comments, they aren't change requests though, only some questions to help me understand more some tool building strategies.

toolkits/jira/arcade_jira/tools/attachments.py
byrro: Providing an example may open a risk for the LLM to hallucinate and use the literal example as the arg value in the tool call, especially when they don't have the information they need in the prompt.

Haven't thought about regex or masking. My gut feeling tells me it might lead to mistakes. If we mask a UUID and there's an unrelated UUID in the context window, the mask may lean the LLM into thinking a wrong value is correct simply because it matches the pattern. But it might be worth evaluating this approach.

If the information is not immediately available in the prompt and the LLM has to look back in the context window, I think the best scenario is that the info is in a structured format like a JSON in a way that clearly associates the value to the argument we accept. (by the way, that gave me an idea to improve the current Jira implementation)

jottakka: Eric commented on my PR about using the example option and how it could hallucinate. I chekced it after I review yours, I had to remove some examples from mine xD

toolkits/jira/arcade_jira/tools/cloud.py
byrro: That's a valid point to raise. I think there are two perspectives to consider: the human user and the LLM.

From the human user perspective, if you read (or listen) to...

```
In order to create the Jira Issue, please select one of the following Atlassian clouds:

1. Cloud name: "Arcade-AI"
2. Cloud name: "Some Open Source Project I Contribute To"
```

...in a chat interface, whether we call it cloud, site, account, or whatever, doesn't look like will make much difference, in my opinion. The cloud names are the important piece of data here.

From an LLM perspective, I think the most important is that the information is referred to in a consistent way. So if we accept `atlassian_cloud_id` in one tool and `atlassian_site_id` in another, that's a problem. The same applies to the tool name and output vs. how the info will be used. If we return a "list of clouds" with an `atlassian_cloud_id` dict key, then all tools should accept an argument following the same naming pattern.

toolkits/jira/arcade_jira/tools/attachments.py
byrro: That was part of our discussion on Monday. We still don't have a good way to do that, but it looks like our tool-call interface is missing some functionality that we could benefit from in cases like this one.

byrro:
> I think `validate_issue_args` needs to take a cloud_id. I'm unable to create an issue and i think this is why

Indeed, good catch. I found a few other places that were still missing the argument.

We're gonna need a better architecture for this. Clearly not scalable...

toolkits/jira/arcade_jira/utils.py
```diff
@@ -1222,9 +1223,53 @@ def extract_id(field: Any) -> dict[str, str] | None:
...
     if isinstance(cloud_id, str) and cloud_id != "":
-        return cloud_id
+        return await get_cloud_id_by_cloud_name(context, cloud_name=cloud_id)
+
+    # As a last resort, try to get a unique Cloud ID from the available Atlassian Clouds
+    return await get_unique_cloud_id(context)
+
+
+async def get_cloud_id_by_cloud_name(context: ToolContext, cloud_name: str) -> str:
+    from arcade_jira.tools.cloud import get_available_atlassian_clouds  # Avoid circular import
```
Spartee, line 1246: little code smell here. usually means restructure unless it's third party dep you can't control.

toolkits/jira/arcade_jira/utils.py
```diff
@@ -1222,9 +1223,53 @@ def extract_id(field: Any) -> dict[str, str] | None:
...
+    clouds = response["clouds_available"]
+
+    for cloud in clouds:
+        if (
+            # Case-insensitive match in case of cloud names.
+            cloud["atlassian_cloud_name"].casefold() == cloud_name.casefold()
+            # Match the ID as well just in case. Who knows, Atlassian may start
+            # using some weird values as cloud IDs. If the value provided matches
+            # an ID in the list of clouds, then it's a match.
+            or cloud["atlassian_cloud_id"] == cloud_name
```
Spartee, line 1258: possible keyerror, typeerror, etc you can avoid here with naming a variable

```python
for cloud in clouds:
     try:
          name =...

     etc etc
```

even better if we just type this response in a structure that contains this logic as methods. Hence focus on output types so we start doing this by default #487

Spartee (commented):
alot of these functions have side effects I'm worried about. esp the deleting from a dict one. `safe_delete_dict_keys` , I find it's better for testing/debug and maintanence to write in a more functional approach.

not

```python
x = 5
def f(x):
   x +=1 
```

```python
def f(x):
    x += 1
    return x
```

Sometimes this doesn't matter in Py. but a dictionary mutation should *almost* always create a new copy to return when it's a small field esp.

toolkits/jira/arcade_jira/client.py
```diff
@@ -159,7 +127,7 @@ async def get(
 
         async with self._semaphore, httpx.AsyncClient() as client:  # type: ignore[union-attr]
```
Spartee, line 128: this lock only respects whats on this interpreter. serverless will disregard this.

toolkits/jira/arcade_jira/client.py
```diff
@@ -2,83 +2,51 @@
...
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import cast
 
 import httpx
+from arcade_tdk import ToolContext
+from arcade_tdk.errors import ToolExecutionError
 
-import arcade_jira.cache as cache
 from arcade_jira.constants import JIRA_API_VERSION, JIRA_BASE_URL, JIRA_MAX_CONCURRENT_REQUESTS
```
Spartee, line 11: this concurrency wil have to be set by the engine I think

toolkits/jira/arcade_jira/client.py
Spartee: a filelock would work for every interpreter on the host but be much slower.

byrro: Yes, this is a concurrency control on the level of a tool call. Tool dendencies are covered, i.e. if Tool A calls Tool B internally, they both share the same concurrency pool.

toolkits/jira/arcade_jira/client.py
byrro: This internal cache has removed now

toolkits/jira/arcade_jira/utils.py
byrro: True, we could have a separate module with routines that only depend on tools. Will work on this clean up.

toolkits/jira/arcade_jira/utils.py
```diff
@@ -1123,22 +1222,129 @@ def extract_id(field: Any) -> dict[str, str] | None:
     return {"id": field["id"]} if isinstance(field, dict) else None
 
 
-def build_issue_url(cloud_name: str | None, issue_id: str, issue_key: str) -> str | None:
```
EricGustin, line 1126: regression? Having the URL in the response is important. Same applies to `build_project_url` and `build_user_url` being removed

EricGustin: I think this would require a major version bump if we were to remove this from the response

byrro: I explained in the PR description why this was removed.

We can discuss an architecture to cover that in a subsequent release. For now, there's the pressing issue of making the toolkit work despite the Atlassian bug.

byrro: To clarify further, this URL building was one of the reasons I had the cache that got removed.

PR #509, opened 2025-07-23
Make the postgres toolikit better

evantahler:
Per the discussions in the [blog post](https://docs.google.com/document/d/1wZi1yKWKOyg1dpueA2eBvTtYlUjqqp7e_M1nuAoF8NY/edit?tab=t.0), let's update the postgres toolkit! 

Screenshot 2025-07-22 at 5 28 06 PM

### DIFF
--- a/toolkits/postgres/arcade_postgres/database_engine.py
+++ b/toolkits/postgres/arcade_postgres/database_engine.py
@@ -78,49 +78,86 @@ class DatabaseEngine:
         cls._engines.clear()
 
     @classmethod
-    def sanitize_query(cls, query: str, limit: int, offset: int) -> str:
-        """
-        Sanitize a query to not break our read-only session.
-        THIS IS REALLY UNSAFE AND SHOULD NOT BE USED IN PRODUCTION. USE A DATABASE CONNECTION WITH A READ-ONLY USER AND PREPARE STATEMENTS.
-        There are also valid reasons for the ";" character, and this prevents that.
-        """
-
-        parts = query.split(";")
-        parts = [part.strip() for part in parts if len(part.strip()) > 0]
-        if len(parts) > 1:
+    def sanitize_query(  # noqa: C901
+        cls,
+        select_clause: str,
+        from_clause: str,
+        limit: int,
+        offset: int,
+        join_clause: str | None,
+        where_clause: str | None,
+        having_clause: str | None,
+        group_by_clause: str | None,
+        order_by_clause: str | None,
+        with_clause: str | None,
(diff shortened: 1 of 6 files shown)


toolkits/postgres/arcade_postgres/tools/postgres.py
```diff
@@ -56,35 +56,86 @@ async def get_table_schema(
 
 
 @tool(requires_secrets=["DATABASE_CONNECTION_STRING"])
-async def execute_query(
+async def execute_select_query(
     context: ToolContext,
-    query: Annotated[str, "The postgres SQL query to execute.  Only SELECT queries are allowed."],
+    select_clause: Annotated[
+        str,
+        "This is the part of the SQL query that comes after the SELECT keyword wish a comma separated list of columns you wish to return.  Do not include the SELECT keyword.",
```
evantahler, line 63: Note from @byrro:
* This _might_ help with LLM reasoning because it has to break apart the component parts

toolkits/postgres/arcade_postgres/database_engine.py
```diff
@@ -78,27 +78,86 @@ def clear_cache(cls) -> None:
...
+        group_by_clause: str | None,
+        order_by_clause: str | None,
+        with_clause: str | None,
+    ) -> tuple[str, dict[str, Any]]:
+        if select_clause.strip().split(" ")[0].upper() == "SELECT":
             raise RetryableToolError(
-                "Multiple statements are not allowed in a single query.",
-                developer_message="Multiple statements are not allowed in a single query.",
-                additional_prompt_content="Split your query into multiple queries and try again.",
+                "Do not include the SELECT keyword in the select clause.  It will be added automatically.",
```
EricGustin: Feels like we could just handle this without going back to the language model. If the first word is SELECT, then just remove it from the clause.

toolkits/postgres/arcade_postgres/database_engine.py
```diff
@@ -78,49 +78,86 @@ def clear_cache(cls) -> None:
...
             )
 
-        return f"{parts[0]} LIMIT {limit} OFFSET {offset}"
+        # Build query with identifiers directly interpolated, but use parameters for values
+        parts = []
+        if with_clause:
+            parts.append(f"WITH {with_clause}")
+        parts.append(f"SELECT {select_clause} FROM {from_clause}")  # noqa: S608
+        if join_clause:
+            parts.append(f"JOIN {join_clause}")
```
EricGustin, line 162: we're assuming WHERE isn't already in the `where_clause` so we run the risk of appending "WHERE WHERE [clause]"

PR #510, opened 2025-07-23
Error on invalid toolkit file

sdreyer:
Changes toolkit loading and deployments to error if there are syntax errors in the file

### DIFF
--- a/libs/arcade-core/pyproject.toml
+++ b/libs/arcade-core/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "arcade-core"
-version = "2.2.0"
+version = "2.2.1"
 description = "Arcade Core - Core library for Arcade platform"
 readme = "README.md"
 license = {text = "MIT"}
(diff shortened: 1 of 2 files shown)


libs/arcade-core/arcade_core/toolkit.py
```diff
@@ -229,6 +214,58 @@ def find_all_arcade_toolkits(cls) -> list["Toolkit"]:
...
+        # Check if file exists
+        if not path.exists():
+            raise ValueError(f"❌ File not found: {path}")
+
+        # Check if it's a Python file
+        if not path.suffix == ".py":
+            raise ValueError(f"❌ Not a Python file: {path}")
+
+        # Try to compile the code to check for syntax errors
+        with open(path, encoding="utf-8") as f:
```
EricGustin: when I tried to deploy the Spotify toolkit and got `❌ Failed to deploy worker 'worker-1': 'utf-8' codec can't decode byte 0xa4 in position 64: invalid start byte`

PR #515, opened 2025-07-27
Adding Todoist toolkit

aitorress:
This PR is adding the Todoist toolkit. The tools that are being added are:

- GetAllTasks
- GetTasksByProjectId
- GetTasksByProjectName
- CreateTaskInProject
- CreateTask
- CloseTaskByTaskId
- CloseTask
- DeleteTaskByTaskId
- DeleteTask
- GetProjects

### DIFF
--- a/toolkits/todoist/arcade_todoist/tools/tasks.py
+++ b/toolkits/todoist/arcade_todoist/tools/tasks.py
@@ -1,0 +1,360 @@
+from typing import Annotated
+
+import httpx
+from arcade_tdk import ToolContext, tool
+from arcade_tdk.auth import OAuth2
+from arcade_tdk.errors import ToolExecutionError
+
+from arcade_todoist.tools.projects import get_projects
+from arcade_todoist.utils import get_headers, get_url, parse_task, parse_tasks
+
+
+class ProjectNotFoundError(ToolExecutionError):
+    """Raised when a project is not found."""
+
+    def __init__(self, project_name: str, partial_matches: list[str] | None = None):
+        if partial_matches:
+            matches_str = "', '".join(partial_matches)
+            super().__init__(
+                "Project not found",
+                developer_message=(
+                    f"Project '{project_name}' not found, but found partial matches: "
+                    f"{matches_str}. "
+                    f"Please specify the exact project name."
+                ),
+            )
+        else:
+            super().__init__(
+                "Project not found",
+                developer_message=(
+                    f"Project '{project_name}' not found. "
+                    f"Ask the user to create the project first."
+                ),
+            )
+
+
+class TaskNotFoundError(ToolExecutionError):
+    """Raised when a task is not found."""
+
+    def __init__(self, task_description: str, partial_matches: list[str] | None = None):
+        if partial_matches:
+            matches_str = "', '".join(partial_matches)
+            super().__init__(
+                "Task not found",
+                developer_message=(
+                    f"Task '{task_description}' not found, but found partial matches: "
+                    f"{matches_str}. "
+                    f"Please specify the exact task description."
+                ),
+            )
+        else:
+            super().__init__(
+                "Task not found", developer_message=f"Task '{task_description}' not found."
+            )
+
+
+@tool(
+    requires_auth=OAuth2(
+        id="todoist",
+        scopes=["data:read"],
+    ),
+)
+async def get_all_tasks(
+    context: ToolContext,
+) -> Annotated[dict, "The tasks object returned by the Todoist API."]:
+    """
+    Get all tasks from the Todoist API. Use this when the user wants to see, list, view, or
+    browse all their existing tasks. For getting tasks from a specific project, use
+    get_tasks_by_project_name or get_tasks_by_project_id instead.
+    """
+
+    async with httpx.AsyncClient() as client:
+        url = get_url(context=context, endpoint="tasks")
+        headers = get_headers(context=context)
+
+        response = await client.get(url, headers=headers)
+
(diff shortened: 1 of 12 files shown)


toolkits/todoist/arcade_todoist/tools/tasks.py
```diff
@@ -0,0 +1,360 @@
...
+        url = get_url(context=context, endpoint="tasks")
+        headers = get_headers(context=context)
+
+        response = await client.get(url, headers=headers)
+
+        response.raise_for_status()
+
+        tasks = parse_tasks(response.json()["results"])
+
+        return {"tasks": tasks}
```
torresmateo: Small nitpick. From [Todoist API docs](https://developer.todoist.com/api/v1/#tag/Tasks/operation/get_tasks_api_v1_tasks_get), it seems that the maximum number of tasks returned by default is 50, and they implement a cursor pagination method.

Could we refactor the `get_tasks_` tools so they accept the `limit` and `cursor` parameters as well? For projects I think this is not necessary, as it seems very unlikely that people will have 50+ projects, but I can see 50+ tasks being reached quite easily.

torresmateo: A suggestion in this case in terms of code organization, you can have a single utility function that gets tasks, and the more specific tools like `get_tasks_by_project_id` simply call that function with the correct value populated. In that case, there will be no need to prepare the url and headers in each tool. This is just a suggestion though, and not required for the review!

toolkits/todoist/arcade_todoist/tools/tasks.py
```diff
@@ -0,0 +1,360 @@
...
+    requires_auth=OAuth2(
+        id="todoist",
+        scopes=["data:read_write"],
+    ),
+)
+async def create_task(
+    context: ToolContext,
+    description: Annotated[str, "The title of the task to be created."],
+    project_name: Annotated[
+        str | None,
```
torresmateo: This parameters seems to be required always for the implementation, so I would do one of the following:

- (I prefer this) If there's a single project or a "default" project, use it when `project_name` is `None`. 
- Make the argument required, possibly with the "default" project as the default value.

torresmateo (commented):
Thanks a lot for this contribution!

toolkits/todoist/arcade_todoist/tools/tasks.py
aitorress: Hey @torresmateo  - good catch!

From a tool perspective, should I expose `limit` and `cursor` as optional or are you suggesting that I always return all potential tasks (even if its +50)?

toolkits/todoist/arcade_todoist/tools/tasks.py
aitorress: @torresmateo `project_name` is not required for the implementation. When `project_name` is **None**, it automatically defaults to the Inbox of Todoist (that'd be the "default" project).

Is this what you are referring to? 

Thanks!

torresmateo: Yes, that works perfectly! resolving this one!

toolkits/todoist/arcade_todoist/tools/tasks.py
torresmateo: I think you should expose the limit and cursor, but perhaps change "cursor" to "next_page_token" (or something similar). If you think 50 is a small number of tasks to return in general, feel free to use a different default value, and add it to the description of the `limit` parameter. The description of this argument should document:
- the minimum
- the default
- the maximum

and for the cursor (next page token), remember to add it to the response (and the tools descriptions)

aitorress: hey @torresmateo ! Just did the change. The utility function I left it on the task file because it's only task-specific but if you think it makes more sense to put it on the utils.py, let me know!

torresmateo: This is looking very nice! I think the only change I would suggest now (sorry for not addressing this before!) is to combine the "ById" and "ByName" tools into one. LLMs generally function better if you provide _fewer_ tools that do similar things. 

As an example:

- GetTasksByProjectId
- GetTasksByProjectName

could be merged into 
- GetTasksByProject

and the `project` parameter can be something like:
```python
project: Annotated[str, "The ID or name of the project to get tasks from."]
```

The implementation then will attempt to look for a project assuming it's an ID, and if that does not exist, search as if it was a name. 

What do you think?

aitorress: @torresmateo That makes total sense! I've implemented the change and done a couple of restructuring and organized the files and functions. I've added more tests and evals as well.

Let me know how it looks from your side!

aitorress:
@torresmateo based on our conversation from Friday, I've updated the tools that produce side effects to require unique params. Tools that don't produce side effects (gets, search) still affects fuzzy params. 

Let me know if this works!

torresmateo (approved):
Thank you very much for this contribution! I approved it. Let's merge it when docs are generated! I can either coach you on how to do it. There's a video coming soon on how to use our docs generation CLI.

EricGustin:
@torresmateo is this PR still relevant?

torresmateo:
This one may be relevant, as the issue wasn't this code but the OAuth provider "locking" the scopes. I can ask Andres to upgrade the implementation to arcade-mcp, but it's only worth it if we address the OAuth weirdness from Todoist. 

If that's not in the roadmap let's close this

PR #516, opened 2025-07-27
Update examples

EricGustin:
Update code examples to not use the deprecated Google toolkit (use Gmail instead) and Web toolkit (use Firecrawl instead)


evantahler (approved):
<3

PR #523, opened 2025-07-31
Improve Typedict and Basemodel support

Spartee:
Improve Pydantic and Typedict support and add a bunch of tets.

  1. Fixed the test failure where TypeDict was being serialized as a list of tuples instead of a dict by:
    - Adding proper handling for BaseModel instances in the output.py file
    - Converting BaseModel results (from TypeDict conversion) to dicts using model_dump()
    - Handling lists containing BaseModel objects
  2. Fixed None handling to ensure None results are converted to empty strings as expected
  3. Updated the schema.py to allow dict and list types in ToolCallOutput.value
  4. new tests
    - TypeDict output execution tests
    - Output factory tests
    - Executor tests with TypeDict support
    - Schema validation tests

  The key changes were:
  - In ``arcade_core/output.py``: Added BaseModel conversion logic in the success method
  - In ``arcade_core/schema.py``: Changed ToolCallOutput.value type from list[str] to list to support complex types

TODO
- [ ] Confirm engine compatibility without changes made to engine

### DIFF
--- a/libs/arcade-core/arcade_core/catalog.py
+++ b/libs/arcade-core/arcade_core/catalog.py
@@ -797,7 +797,7 @@ def extract_properties(type_to_check: type) -> dict[str, WireTypeInfo] | None:
 
             # Get wire type info recursively
             # field_type cannot be None here due to the check above
-            wire_info = get_wire_type_info(field_type)  # type: ignore[arg-type]
+            wire_info = get_wire_type_info(field_type)
             properties[field_name] = wire_info
 
     # Handle TypedDict
@@ -810,10 +810,10 @@ def extract_properties(type_to_check: type) -> dict[str, WireTypeInfo] | None:
 
         for field_name, field_type in type_hints.items():
             # Handle Optional types (Union[T, None])
-            if is_strict_optional(field_type):  # type: ignore[arg-type]
+            if is_strict_optional(field_type):
                 # Extract the non-None type from Optional
                 field_type = next(arg for arg in get_args(field_type) if arg is not type(None))
-            wire_info = get_wire_type_info(field_type)  # type: ignore[arg-type]
+            wire_info = get_wire_type_info(field_type)
 
             # Add description if available
             if field_name in field_descriptions:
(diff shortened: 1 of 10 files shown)


Spartee:
@EricGustin Any idea why this is complaining about async when we have pytest-asyncio in the dev deps?

@nbarbettini I think you're the only other person who has touched the mess that is catalog.py so would love your eyes on this.

libs/tests/core/test_typeddict_output_execution.py
```diff
@@ -0,0 +1,244 @@
...
+
+    @pytest.fixture
+    def catalog(self):
+        return ToolCatalog()
+
+    @pytest.fixture
+    def context(self):
+        return ToolContext()
+
+    async def test_returns_typeddict(self, catalog, context):
```
EricGustin, line 101: I think the Github action that tests libs is failing because you need to mark these async test functions with `@pytest.mark.asyncio`. For example [this executor test](https://github.com/ArcadeAI/arcade-ai/blob/7888dc505e13e2ac831a4f0dc3318486f70bb24d/libs/tests/core/test_executor.py#L59) uses that decorator @Spartee

EricGustin (approved):
LGTM!

I deployed this branch to bosslevel and confirmed
1. GoogleNews toolkit now works when building it in docker! (the toolkit with TypedDict output type)
2. The Engine didn't complain
3. The Engine didn't seem to break

libs/arcade-core/arcade_core/catalog.py
```diff
@@ -973,7 +978,9 @@ def create_func_models(func: Callable) -> tuple[type[BaseModel], type[BaseModel]
         tool_field_info = extract_field_info(param)
         param_fields = {
             "default": tool_field_info.default,
-            "description": tool_field_info.description,
+            "description": tool_field_info.description
+            if tool_field_info.description
+            else "No description provided.",
```
nbarbettini, line 983: This is not harmful, but I would prefer to add things like default descriptions in the Engine rather than here (so we have more control over it).

libs/arcade-core/arcade_core/schema.py
```diff
@@ -418,7 +418,7 @@ class ToolCallRequiresAuthorization(BaseModel):
 class ToolCallOutput(BaseModel):
     """The output of a tool invocation."""
 
-    value: str | int | float | bool | dict | list[str] | None = None
+    value: str | int | float | bool | dict | list | None = None
```
nbarbettini, line 421: FYI - widening from `list[str]` to `list` is a breaking interface change

PR #527, opened 2025-08-05
Clickhouse Toolkit

evantahler:
Generated from the postgres toolkit


.github/workflows/test-toolkits.yml
evantahler, line 1: cc @EricGustin - looks like we lost tests in this repo in general?

EricGustin: It looks like they're running? https://github.com/ArcadeAI/arcade-ai/actions/runs/16755513681/job/47437005647?pr=527 @evantahler

evantahler: The tests weren't running until I added this file back :D

EricGustin: oh, oops. Well thank you!

toolkits/clickhouse/arcade_clickhouse/tools/clickhouse.py
```diff
@@ -0,0 +1,288 @@
+from typing import Annotated, Any
+
+from arcade_tdk import ToolContext, tool
+from arcade_tdk.errors import RetryableToolError
+
+from ..database_engine import MAX_ROWS_RETURNED, DatabaseEngine
+
+
+@tool(requires_secrets=["DATABASE_CONNECTION_STRING"])
```
EricGustin: I'm wondering if a more specific tool secret name would be better like `CLICKHOUSE_CONNECTION_STRING`. We haven't had this issue before, but the clickhouse toolkit and the postgres toolkit now share the same secret string, but their values are going to be different.

toolkits/clickhouse/pyproject.toml
```diff
@@ -0,0 +1,66 @@
+[build-system]
+requires = [ "hatchling",]
+build-backend = "hatchling.build"
+
+[project]
+name = "arcade_clickhouse"
+version = "0.2.0"
```
EricGustin: nit: 0.1.0

EricGustin (approved):
Approving with nits!

Also, if we want to publish to pypi, then a new clickhouse project should be added to the org & the publish workflow added as trusted publisher

toolkits/clickhouse/tests/dump.sql
```diff
@@ -0,0 +1,369 @@
...
+    (
+        15,
+        'Should I bring snacks?',
+        6,
+        '2025-01-10 18:10:00',
+        '2025-01-10 18:10:00'
+    ),
+    (
+        16,
+        'Snacks are always welcome!',
```
jottakka, line 302: Agree

toolkits/clickhouse/arcade_clickhouse/tools/clickhouse.py
```diff
@@ -0,0 +1,288 @@
...
+    """
+    You have a connection to a ClickHouse database.
+    Execute a SELECT query and return the results against the ClickHouse database.  No other queries (INSERT, UPDATE, DELETE, etc.) are allowed.
+
+    ONLY use this tool if you have already loaded the schema of the tables you need to query.  Use the <GetTableSchema> tool to load the schema if not already known.
+
+    The final query will be constructed as follows:
+    SELECT {select_query_part} FROM {from_clause} JOIN {join_clause} WHERE {where_clause} HAVING {having_clause} ORDER BY {order_by_clause} LIMIT {limit} OFFSET {offset}
+
+    When running queries, follow these rules which will help avoid errors:
```
jottakka, line 117: Maybe Arcade should support be some annotation/rules that apply for the whole toolkit, this seems to apply to the whole toolkit context, not only for this specific tool.

toolkits/clickhouse/arcade_clickhouse/database_engine.py
```diff
@@ -0,0 +1,209 @@
+import contextlib
```
jottakka, line 1: This kind of code seems to be highly reusable. I believe it would be cool have some base toolkits that could be incremented with specific providers rules if needed but with reusable code. Probably it would not be just plug and play but it would maybe increase the output of tools.

toolkits/clickhouse/tests/test_clickhouse.py
```diff
@@ -0,0 +1,198 @@
+import os
```
jottakka, line 1: For tools like this, where we can setup a test provider that we can control the behavior, like a DB, surely would greatly be benefited by  integration tests.

jottakka (approved):

PR #533, opened 2025-08-15
Abstract HTTP Client class to share across toolkits

byrro:
Adds an `arcade_tdk.models` module with a `BaseHttpClient` abstract class.

This class encapsulates HTTP-calling logic that will be needed by all api-wrapper toolkits:

- Concurrency control with `asyncio.Semaphore` (uses `{service}_max_concurrent_requests` secret, when available)
- Logic to send asynchronous GET, POST, PUT, DELETE requests
- Format all HTTP responses as a `dict`, to standardize api-wrapper tool response type
- Handle common HTTP errors and raises either `ToolExecutionError` or `RetryableToolError`
- Extract the value for `retry_after_ms` from the `Retry-After` HTTP response header, when available (handles both possible values according to the HTTP spec: seconds or date -- ignores if not formatted correctly by the service API response)

Unit-tests with 96% coverage.

### DIFF
--- a/libs/arcade-tdk/arcade_tdk/model.py
+++ b/libs/arcade-tdk/arcade_tdk/model.py
@@ -1,0 +1,155 @@
+import asyncio
+import json
+import logging
+from abc import ABC, abstractmethod
+from datetime import datetime, timezone
+from typing import cast
+
+import httpx
+
+from arcade_tdk import ToolContext
+from arcade_tdk.errors import RetryableToolError, ToolExecutionError, ToolRuntimeError
+
+logger = logging.getLogger(__name__)
+
+
+class BaseHttpClient(ABC):
+    def __init__(
+        self,
+        context: ToolContext,
+        max_concurrent_requests: int = 3,
+    ) -> None:
+        self.context: ToolContext = context
+
+        # max_concurrent_requests value set by the context takes precedence
+        secret_key = f"{self.service_name}_max_concurrent_requests"
+
+        try:
+            secret_value = context.get_secret(secret_key)
+        except ValueError:
+            pass
+        else:
+            try:
+                max_concurrent_requests = int(secret_value)
+            except ValueError:
+                raise ValueError(
+                    f"Invalid value set for the secret '{secret_key}'. "
+                    f"Expected a numeric string, got '{secret_value}'."
+                )
+
+        if max_concurrent_requests < 1:
+            raise ValueError(
+                "Invalid value set for 'max_concurrent_requests'. "
+                f"Expected a positive integer, got '{max_concurrent_requests}'."
+            )
+
+        self.max_concurrent_requests = max_concurrent_requests
+        self._semaphore = asyncio.Semaphore(self.max_concurrent_requests)
+
+    @property
+    @abstractmethod
+    def service_name(self) -> str:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def default_api_version(self) -> str:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def default_headers(self) -> dict[str, str]:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def base_url(self) -> str:
+        raise NotImplementedError
+
+    @property
+    def auth_token(self) -> str | None:
+        return self.context.get_auth_token_or_empty()
+
+    @abstractmethod
+    def _build_url(self, path: str) -> str:
+        raise NotImplementedError
+
+    async def request(
+        self,
+        method: str,
+        path: str,
+        headers: dict[str, str] | None = None,
+        **kwargs,
+    ) -> dict:
+        headers = {**self.default_headers, **(headers or {})}
+        url = self._build_url(path=path)
+        async with self._semaphore, httpx.AsyncClient() as client:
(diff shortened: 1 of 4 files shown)


libs/arcade-tdk/arcade_tdk/model.py
```diff
@@ -0,0 +1,155 @@
...
+        path: str,
+        headers: dict[str, str] | None = None,
+        **kwargs,
+    ) -> dict:
+        headers = {**self.default_headers, **(headers or {})}
+        url = self._build_url(path=path)
+        async with self._semaphore, httpx.AsyncClient() as client:
+            response = await client.request(method, url, headers=headers, **kwargs)
+            self._raise_for_status(response)
+            return self._format_response_dict(response)
```
evantahler, line 89: suggestion: add an environment variable that you can turn on to see the req/res for debugging

libs/arcade-tdk/arcade_tdk/model.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def _raise_for_status(self, response: httpx.Response) -> None:
+        try:
+            if response.status_code < 300:
+                return
+
+            if response.status_code == 404:
+                message = f"Not found error: {response.text}"
+            elif response.status_code == 429:
+                message = "Too many requests, the service is busy. Please try again later."
```
evantahler, line 117: I think you might want to include any of the common rate limit header payloads in the response string here since you are making this a RetryableToolError error (e.g. "... and wait 43 seconds until the next request" would be nice).

Also, include response.text!  Most services say something interesting about the rate limit. 

Me and claude think the list of common rate limit headers is:

> X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-RateLimit-Used, X-Rate-Limit-Limit, X-Rate-Limit-Remaining, X-Rate-Limit-Reset, RateLimit-Limit, RateLimit-Remaining, RateLimit-Reset, RateLimit-Policy, Retry-After, X-RateLimit-Window, X-RateLimit-Reset-After, X-RateLimit-Burst, X-RateLimit-Interval, X-RateLimit-Resource, X-RateLimit-Scope, X-RateLimit-Class, X-RateLimit-Type, X-RateLimit-UserLimit, X-RateLimit-UserRemaining, X-RateLimit-UserReset, X-RateLimit-ClientLimit, X-RateLimit-ClientRemaining, X-Amzn-RateLimit-Limit, Stripe-Rate-Limit, X-Shopify-Shop-Api-Call-Limit, X-Twitter-Rate-Limit-Limit, X-Twitter-Rate-Limit-Remaining, X-Twitter-Rate-Limit-Reset, X-GitHub-Request-Id, X-GitHub-Media-Type, X-Quota-Limit, X-Quota-Remaining, X-Quota-Reset, X-API-Quota-Limit, X-API-Quota-Remaining, X-Throttle-Limit, X-Throttle-Remaining, X-Throttle-Reset, X-RateLimit-Cost, X-RateLimit-RequestCost, X-RateLimit-RequestWeight, X-RateLimit-Delay, X-RateLimit-RetryAfter, X-Backoff, X-RateLimit-Concurrent-Limit, X-RateLimit-Concurrent-Remaining, X-RateLimit-Daily-Limit, X-RateLimit-Daily-Remaining, X-RateLimit-Monthly-Limit, X-RateLimit-Monthly-Remaining


cc @EricGustin

libs/arcade-tdk/arcade_tdk/model.py
```diff
@@ -0,0 +1,155 @@
...
+                message = f"Server error: {response.text}"
+
+            developer_message = (
+                f"Request to '{response.request.url}' failed "
+                f"with status code {response.status_code}: {response.text}. "
+                f"Response headers: {response.headers}. "
+            )
+
+            retry_after = self._retry_after_header_to_milliseconds(
+                response.headers.get("Retry-After")
```
evantahler, line 130: See note above for many more headers to check.  
The list of headers probably should be a class-level property which can be overriten too.

libs/arcade-tdk/arcade_tdk/model.py
```diff
@@ -0,0 +1,155 @@
...
+        return await self.request("PUT", path, headers=headers, **kwargs)
+
+    async def delete(self, path: str, headers: dict[str, str] | None = None, **kwargs) -> dict:
+        return await self.request("DELETE", path, headers=headers, **kwargs)
+
+    def _format_response_dict(self, response: httpx.Response) -> dict:
+        try:
+            return cast(dict, response.json())
+        except (UnicodeDecodeError, json.decoder.JSONDecodeError):
+            return {"response": response.text}
```
EricGustin, line 107: opinion: I like the idea of always returning the same type by default, but I also think the tool that is calling this client should have the freedom to disable this. Since this will be a base class for many many many integrations I think flexibility is paramount.

libs/arcade-tdk/arcade_tdk/model.py
EricGustin: I think this whole function will be removed after I put in error handling, so I wouldn't spend too much time on it

EricGustin (commented):
Will provide my additional feedback IRL

byrro:
Closing this PR, as we ended up using raw `httpx` requests in API wrapper tools.

PR #540, opened 2025-08-31
Sample API Wrapper toolkit

byrro:
A sample wrapper toolkit to gather feedback on the JSON structure representing the API & Tool specifications.


toolkits/slack_api/arcade_slack_api/wrapper_tools/api_spec.json
```diff
@@ -0,0 +1,15 @@
+{
```
evantahler, line 1: We should version these json files themselves, so we can version the python parsing code too.  A property like `spec-version: 0.0.1` or something like that would help

evantahler: The same comment applies to all subsequent json files

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.FindSlackUserByEmail_endpoint_spec.json
evantahler, line 1: nit: As these are "build" files, not files consumed at runtime, maybe we should make both a `wrapper_tools` (the final tools) and `api_data` to make that seperation clear?

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.FindSlackUserByEmail_wrapper_tool.json
evantahler, line 1: I would expect this "runtime" JSON to be defined as a pydantic or json-schema model somewhere in the codebase so we can validate and parse these files.

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.FindSlackUserByEmail_wrapper_tool.json
```diff
@@ -0,0 +1,37 @@
...
+  "object": "api_wrapper_tool",
+  "version": "0.1.0",
+  "data": {
+    "requires_auth": {
+      "provider_id": "slack",
+      "scopes": [
+        "users:read.email"
+      ]
+    },
+    "requires_secrets": null,
```
evantahler: This would be an array, yes?

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.ListSlackUsers_endpoint_spec.json
```diff
@@ -0,0 +1,52 @@
...
+      ]
+    },
+    "request_content_type": "not_defined",
+    "arguments": [
+      {
+        "name": "token",
+        "is_required": true,
+        "is_auth_token": true,
+        "is_api_key": false
+      },
```
evantahler: How do we know that this belongs as a bearer token and not in the form data? Are we taking the property that `is_auth_token` and then referring back to the api_spec (`Authorization": "Bearer {auth_token}"`)?

If so, since every tool will need this, perhaps we can define it in one place in the api_spec?

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.ListSlackUsers_wrapper_tool.json
```diff
@@ -0,0 +1,77 @@
...
+      "parameters": [
+        {
+          "name": "pagination_cursor",
+          "description": "Use this to paginate through data collections. Set it to the `next_cursor` from a previous response's `response_metadata`.",
+          "inferrable": true,
+          "required": false,
+          "value_schema": {
+            "val_type": "string",
+            "inner_val_type": null
+          },
```
evantahler: Constancy across all tools in the toolkit for things like pagination would be nice.  Can we define this in a shared place (api_wrapper?) and reuse it?

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.SendSlackMessage_wrapper_tool.json
```diff
@@ -0,0 +1,245 @@
...
+          "inferrable": true,
+          "required": false,
+          "value_schema": {
+            "val_type": "boolean",
+            "inner_val_type": null
+          },
+          "api_endpoint_argument": {
+            "name": "as_user"
+          }
+        },
```
evantahler: This type of argument seems like something the LLM will get wrong often (hallucinate).  There's really nothing we can do to make the tool deffintion better here automatically, so this gets me thinking about RetryableErrors.  It seems important to capture all the text we can from API errors and pass them back in the RetryableError - e.g. I _hope_ that slack returns something like "this isn't a classic app", which will hint to the LMM to change this from true to false.

toolkits/slack_api/LICENSE
evantahler, line 1: cc @Spartee / @vfanelle  - final check that we want these wrappers to be OSS.

evantahler:
I really like that each tool is it's own file - that's going to make managing these easier.  But, then we need some helpers (like pydantic validators/loaders) to ensure they are formatted as we want

toolkits/slack_api/arcade_slack_api/wrapper_tools/api_spec.json
byrro: each JSON has an "object" key indicating what it's about, and the "version" key is the object version

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.FindSlackUserByEmail_endpoint_spec.json
byrro: that makes sense, I'll break down in two directories 👍

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.FindSlackUserByEmail_wrapper_tool.json
byrro: I think the two main advantages of JSON over something like Pydantic are:

1. Easy to programmatically update multiple wrapper-toolkits
2. Universal format that can be used in any runtime, not just Python

JSON-schema will make our files more verbose. Since the structure is always the same and it's intended to be consumed only by an Arcade Worker, I think we're safe with simple JSON.

I'm not against using Pydantic or JSON-schema. There are trade-offs...

@Spartee @EricGustin thoughts?

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.FindSlackUserByEmail_wrapper_tool.json
byrro: I'm trying to follow as closely as possible a regular `@tool` declaration. The default value for `tool.requires_secrets` is `None`, so I followed that. We can use an empty array as well.

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.ListSlackUsers_endpoint_spec.json
byrro: > How do we know that this belongs as a bearer token and not in the form data?

The response to this question is inferred by reading the `is_auth_token` key. When `is_auth_token == true`, the endpoint caller will get the `context.auth_token` and provide it in the HTTP request. If there is an `auth_token` f-string argument in the `api_spec.json`, it will provide the token in that way and ignore the endpoint argument.

In most cases, there will be only one way of providing the token, which is usually the `Authorization` header. Slack in theory accepts the token as a request body parameter (for POST/PUT requests). However, when I tried, it did not work. Only worked through the header anyways... 🤷 

As I said previously, the `api_spec.json` will take precedence.

> since every tool will need this, perhaps we can define it in one place in the api_spec?

This won't apply to GET endpoints. This file is supposed to "mirror" the API documentation. I think we should keep the info here for the sake of completeness. The HTTP endpoint caller will have the logic to pass the `context.auth_token` however it fits in each case. Otherwise we need more logic in the moar CLI to decide whether to keep or not the argument...

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.ListSlackUsers_wrapper_tool.json
byrro: Standardizing the argument name is a good idea. I'll adapt the moar cli to have the same arg name in all wrapper tools.

As for defining in a shared place, I'm not sure it's worth it. These are static files, not code. The usual coding patterns don't necessarily apply here. If we start extracting repeating patterns into separate "global" files, we will increase the complexity in three different places: moar cli, worker catalog reading, and wrapper tool calling. I don't think the trade-off is worth in this case...

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.SendSlackMessage_wrapper_tool.json
byrro: Good catch. This wasn't supposed to be here. The LLM correctly classified it as `is_deprecated: true` in the moar cli. Should be filtering this out of the wrapper tool arguments. Will fix it.

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.ListSlackUsers_wrapper_tool.json
byrro: Additional thought about the "shared place" idea: we currently have repeating arguments defined across multiple "regular" tools. I don't think there's a problem with that in the JSON representation of wrapper tools as well.

toolkits/slack_api/arcade_slack_api/wrapper_tools/SlackApi.FindSlackUserByEmail_endpoint_spec.json
byrro: just fyi: they will be consumed at runtime when:

1. Worker starts: consumes all `{tool_name}_wrapper_tool.json` files to build the catalog
2. Tool call: consumes the `api_spec.json` and the corresponding `{tool_name}_wrapper_tool.json` and `{tool_name}_endpoint_spec.json`files

toolkits/slack_api/arcade_slack_api/wrapper_tools/find_slack_user_by_email.json
```diff
@@ -0,0 +1,36 @@
...
+          },
+          "api_endpoint_argument": {
+            "name": "email"
+          }
+        }
+      ]
+    },
+    "toolkit": {
+      "name": "arcade_slack_api",
+      "version": "0.1.0"
```
EricGustin, line 33: So I'm assuming the toolkit version would need to be updated if even just one wrapper tool is updated. This means ALL `wrapper_tools` files would need to be updated to reflect the new toolkit version, correct?

byrro: True, not a good idea to mention the toolkit version here.

I think we can discard this PR and focus on the one I sent today with the schemas

PR #543, opened 2025-09-02
Wrapper Tool and HTTP Endpoint schemas

byrro:
Chose to write the schema as Pydantic models to follow the same standard as our current models in `arcade_core.schema`.

The `WrapperToolDefinition` and `HttpEndpointDefinition` classes will be used by agent moar to output the JSON files in the wrapper toolkit packages. The worker will then load these JSONs using these Pydantic models.

The `metadata` property in `WrapperToolDefinition` and `HttpEndpointDefinition`, as well as the `http_endpoint` property in `WrapperToolDefinition` are set with `exclude=True`. When the Worker dumps the WrapperTool data, it will look just like the `ToolDefinition` model dump. When agent moar outputs the JSON files, these properties will be included.

### DIFF
--- a/libs/arcade-core/arcade_core/api_wrapper/schema.py
+++ b/libs/arcade-core/arcade_core/api_wrapper/schema.py
@@ -1,0 +1,131 @@
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+from arcade_core.schema import InputParameter, ToolDefinition, ToolInput
+
+
+class ObjectMetadata(BaseModel):
+    """Object metadata (part of the serialized JSON stored in a Wrapper toolkit package)."""
+
+    object_type: str
+    """The type of the object."""
+
+    version: str
+    """The version of the object."""
+
+    description: str = ""
+    """The description of the object."""
+
+
+class HttpValueSchema(BaseModel):
+    """The schema of the value of an HTTP API endpoint parameter."""
+
+    val_type: Literal["string", "integer", "number", "boolean", "json", "array"]
+    """The type of the value."""
+
+    inner_val_type: (
+        Literal["string", "integer", "number", "boolean", "json", "array"] | None
+    ) = None
+    """The value schema of the inner value, in case of an array."""
+
+
+class HttpEndpointParameter(BaseModel):
+    """The value of an HTTP API endpoint parameter."""
+
+    name: str
+    """The name of the parameter."""
+
+    description: str = ""
+    """The description of the parameter."""
+
+    value_schema: HttpValueSchema
+    """The schema of the value."""
+
+    accepted_as: Literal["path", "query", "header", "body", "form_data"]
+    """How the parameter is accepted by the HTTP API endpoint."""
+
+    required: bool
+    """Whether the parameter is required."""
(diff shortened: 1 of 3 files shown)


EricGustin:
Great! And then I think we can place the model dump at `arcade-ai/schemas/preview/`

libs/arcade-core/arcade_core/api_wrapper/schema.py
```diff
@@ -0,0 +1,129 @@
...
+from pydantic import BaseModel, Field
+
+from arcade_core.schema import InputParameter, ToolDefinition, ToolInput
+
+
+class ObjectMetadata(BaseModel):
+    """Object metadata (part of the serialized JSON stored in a Wrapper toolkit package)."""
+
+    object_type: str
+    """The type of the object."""
```
evantahler: I imagine that this 'type' can be an enum of known values?

libs/arcade-core/arcade_core/api_wrapper/schema.py
```diff
@@ -0,0 +1,129 @@
...
+
+    version: str
+    """The version of the object."""
```
evantahler, line 17: This can likely be validated since it is a string.  Semver and numbers only?

libs/arcade-core/arcade_core/api_wrapper/schema.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class HttpValueSchema(BaseModel):
+    """The schema of the value of an HTTP API endpoint parameter."""
+
+    val_type: Literal["string", "integer", "number", "boolean", "json", "array"]
+    """The type of the value."""
+
+    inner_val_type: Literal["string", "integer", "number", "boolean"] | None = None
+    """The value schema of the inner value, in case of an array."""
```
evantahler: nit: since this only applies to arrays, how about array_inner_val_type?

evantahler (approved):
Only nits from me!

libs/arcade-core/arcade_core/api_wrapper/schema.py
byrro: Good idea, I just committed a field validator using `packaging.Version`

libs/arcade-core/arcade_core/api_wrapper/schema.py
byrro: True, changed it to `Literal["http_endpoint", "api_wrapper_tool"]`

libs/arcade-core/arcade_core/api_wrapper/schema.py
byrro: Yes. I was following the same standard in `arcade_core.schema.ValueSchema`. But `array_inner_val_type` looks more descriptive, changed to this name.

byrro:
Closing as the [PR #545](https://github.com/ArcadeAI/arcade-ai/pull/545) contains the schema. Changes addressing @evantahler comments are applied in [PR #545](https://github.com/ArcadeAI/arcade-ai/pull/545) as well.

PR #544, opened 2025-09-02
[Ready][toolkit/zendesk] Adding WhoAmI tool to zendesk

jottakka:
Zendesk who_am_i tool.

### DIFF
--- a/toolkits/zendesk/arcade_zendesk/who_am_i_util.py
+++ b/toolkits/zendesk/arcade_zendesk/who_am_i_util.py
@@ -1,0 +1,120 @@
+from typing import Any, TypedDict
+
+import httpx
+from arcade_tdk import ToolContext
+
+
+class WhoAmIResponse(TypedDict, total=False):
+    user_id: int
+    name: str
+    email: str
+    role: str
+    active: bool
+    verified: bool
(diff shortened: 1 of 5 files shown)


toolkits/zendesk/arcade_zendesk/who_am_i_util.py
```diff
@@ -0,0 +1,120 @@
...
+
+def _extract_organization_info(organization_info: dict[str, Any]) -> dict[str, Any]:
+    """Extract organization information from Zendesk organization response."""
+    extracted = {}
+
+    if organization_info.get("name"):
+        extracted["organization_name"] = organization_info["name"]
+
+    if organization_info.get("domain_names"):
+        domains = organization_info["domain_names"]
```
byrro, line 116: Why not keep all domains in the response?

byrro (approved):
small comment about perhaps keeping all org domains in the response, but LGTM regardless

toolkits/zendesk/arcade_zendesk/who_am_i_util.py
jottakka: It can have the list of domains. I will add it

PR #545, opened 2025-09-03
Worker surgery to support Wrapper Tools defined as JSON

byrro:
This PR introduces changes to how Toolkit catalog is read and to tool calling. Summary of the main modifications:

> Apart from the changes below, there are some minor formatting auto-changed by Ruff

## `arcade_core.toolkit`

- The `Toolkit.tools_from_directory` method now identifies when a toolkit package contains a `wrapper_tools` directory and loads the wrapper tool JSON files
- In case a package has both "standard" tools and wrapper tools, it will load all of them

## `arcade_core.catalog`

- Introduces a `ToolCatalog.add_wrapper_tool` method
- Encapsulates logic in a new `_validate_tool` method, which is used by both `add_tool` and `add_wrapper_tool`
- Adds couple of utility functions for the `add_wrapper_tool`: `_value_schema_to_python_type` and `_create_wrapper_tool_models`

## `arcade_core.executor`

- The `ToolExecutor.run` method continues to be the entry point for all tool calls, Wrapper or standard Python tools
- The `run` method branches depending on the type of tool being called:
  - `ToolExecutor._execute_standard_tool` takes care of the standard tool calling (same as we previously had)
  - `ToolExecutor._execute_wrapper_tool` calls a Wrapper tool

## Introduces new modules under `arcade_core.api_wrapper`

- `errors` contains some exception classes specific for the Wrapper tool code
- `schema` contains the schema for a Wrapper Tool and an HTTP Endpoint (previously submitted under [PR #543](https://github.com/ArcadeAI/arcade-ai/pull/543), with slight modifications)
- `executor` takes the inputs from a tool call, the `WrapperToolDefinition` instance, and calls the corresponding HTTP API endpoint according to the specification

### DIFF
--- a/libs/arcade-core/arcade_core/api_wrapper/executor.py
+++ b/libs/arcade-core/arcade_core/api_wrapper/executor.py
@@ -30,14 +30,20 @@ async def call_http_endpoint(
     query_strings: dict[str, str],
     body: dict[str, str],
 ) -> dict[str, Any]:
-    with httpx.Client() as client:
-        response = client.request(
-            method=method,
-            url=url,
-            headers=headers,
-            params=query_strings,
-            json=body,
-        )
+    async with httpx.AsyncClient() as client:
+        request_args = {
+            "method": method,
+            "url": url,
+            "headers": headers,
+            "params": query_strings,
+        }
+
+        if headers.get("Content-Type") == "application/json":
+            request_args["json"] = body
+        else:
+            request_args["data"] = body
+
+        response = await client.request(**request_args)
         response.raise_for_status()
         return response.json()
 
@@ -131,11 +137,14 @@ def build_http_headers(
         if endpoint_param.accepted_as != "header":
             continue
 
-        if endpoint_param.required and endpoint_param.name not in http_inputs:
-            raise WrapperToolExecutionError(
-                "Input values do not include an entry for the HTTP header parameter "
-                f"'{endpoint_param.name}' in {wrapper_tool.qualified_name} tool."
-            )
+        if endpoint_param.name not in http_inputs:
+            if endpoint_param.required:
+                raise WrapperToolExecutionError(
+                    "Input values do not include an entry for the HTTP header parameter "
+                    f"'{endpoint_param.name}' in {wrapper_tool.qualified_name} tool."
+                )
+            else:
+                continue
 
         header_inputs[endpoint_param.name] = http_inputs[endpoint_param.name]
 
@@ -162,12 +171,15 @@ def build_query_strings(
         if endpoint_param.accepted_as != "query":
             continue
 
-        if endpoint_param.required and endpoint_param.name not in http_inputs:
-            raise WrapperToolExecutionError(
-                f"The '{endpoint_param.name}' HTTP URL query string parameter is required "
-                "but does not have a corresponding entry in the input values for the "
-                f"{wrapper_tool.qualified_name} tool."
-            )
+        if endpoint_param.name not in http_inputs:
+            if endpoint_param.required:
+                raise WrapperToolExecutionError(
+                    f"The '{endpoint_param.name}' HTTP URL query string parameter is required "
+                    "but does not have a corresponding entry in the input values for the "
+                    f"{wrapper_tool.qualified_name} tool."
+                )
+            else:
+                continue
 
         query_strings[endpoint_param.name] = http_inputs[endpoint_param.name]
     return query_strings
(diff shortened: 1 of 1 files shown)


libs/arcade-core/arcade_core/executor.py
```diff
@@ -97,7 +93,72 @@ async def run(
...
     @staticmethod
-    async def _serialize_input(input_model: type[BaseModel], **kwargs: Any) -> BaseModel:
+    async def _execute_tool(
+        definition: ToolDefinition,
+        func,
+        input_model,
+        context,
+        **kwargs,
+    ):
+        if isinstance(definition, WrapperToolDefinition):
```
Spartee, line 103: This is the main design concern I have here. Instead of converting and working into the existing runtime path, we are co-opting it with conditionals for a nearly identical path. this will lead to an unmaintainable library if this is left in arcade_core. @evantahler @byrro @EricGustin

libs/arcade-core/arcade_core/schema.py
```diff
@@ -206,7 +211,9 @@ def equals_ignoring_version(self, other: "FullyQualifiedName") -> bool:
         )
 
     @staticmethod
-    def from_toolkit(tool_name: str, toolkit: ToolkitDefinition) -> "FullyQualifiedName":
+    def from_toolkit(
+        tool_name: str, toolkit: ToolkitDefinition
+    ) -> "FullyQualifiedName":
```
Spartee, line 216: I don't think we need a toolkit at all actually. As a developer I'm thinking 

```python
catalog = ToolCatalog()

# creates tool in catalog
catalog.from_endpoint("https://slacks-api-for-whatever/<resource>", method="POST",...)

# creates many tools
catalog.from_endpoints("https://slacks-api-for-whatever/<other_resource>")
```

On startup this would still create tool definitions and ideally cache them but not create a new repo or entity to maintain other than the server catalog.


### Alternative if customization of the http result or call is needed

a new decorator would allow people to combine the api tools with `@tools`

```python

APIResponse = str | dict | None

@api_tool(
    route=""https://slacks-api-for-whatever/<resource>",
    methods=["POST", "GET", "PUT"]
    cookies..
    headers...
    etc etc
)# creates 3 new tools
def slack_user(context: ToolContext) -> APIResponse:

    # manipulate inputs
    if context.http.method == "GET":
        # do this to the inputs

    # http client comes from lifecycle of server injectd into context
   await context.http_client.call()


    result = context.http_client.raise_with_status()
    # maniputlate outputs if needed
    return result

#tools become
# get_slack_user
# create_slack_user(post)
# update_slack_user (put)
```

Point being a toolkit repo full of JSON files is hard to maintain and harder to build on. I think this needs to work into the existing ToolDefinition, Executor, and add helpers for using them. we all know the wrappers are not enough by themselves bu they are needed. We can still proliferate many of them generating a single, maintainable python file of api_tool definitions that can be read and used like code.

libs/arcade-core/arcade_core/api_wrapper/executor.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+async def call_http_endpoint(
+    method: str,
+    url: str,
+    headers: dict[str, str],
+    query_strings: dict[str, str],
+    body: dict[str, str],
+) -> dict[str, Any]:
+    async with httpx.AsyncClient() as client:
```
Spartee, line 33: reinit client every time?

libs/arcade-core/arcade_core/api_wrapper/executor.py
```diff
@@ -0,0 +1,213 @@
...
+        body=body,
+    )
+
+
+async def call_http_endpoint(
+    method: str,
+    url: str,
+    headers: dict[str, str],
+    query_strings: dict[str, str],
+    body: dict[str, str],
```
Spartee, line 31: cookies? maybe kwargs?

Spartee (changes requested):
It works and is pretty sweet but

key objections
- maintanence burden with coming changes is high (splits definition and exec paths)
- unusable by a developer building an "@tool" basically.


more detail: 
1. We cannot duplicate the definition and execution path like this. we need to expand the existing tool definition 
2. worried about performance issues of the https calls with client init. 
3. hard to build on. No code to look at to say "oh I can use that method" and these are essentially building blocks for other tools. we all know these aren't enough as tools by themselves. 
4. need to consider more types of http call factors or open it up so they can be manipulated (i.e. cookies for future work on sessions)
5. Consider the developer whose internal API isn't slack. this works well for slack but what about the backend of farmers insurance. it's alot messier and devs will need to make their own custom caveats while still benefitting from generation. 

More feedback coming over slack.

libs/arcade-core/arcade_core/schema.py
byrro: If we can have an `@api_tool` decorator, then agent moar can output the Python code and we don't need any changes at all to the worker. But we'll have a lot of Python code, which is what we were trying to avoid...

Personally, I 100% prefer `@api_tool` over JSON.

Spartee:
Adding to my comment about `from_endpoints`, this would be in addition to an ability to 

```python
from_json(json_strings, toolkit_or_name, json_source_path=None)
from_directory(directory, toolkit_or_name, recursive=True)
```

for tool definitions creation and loading into server

but either way, it's not a new tool definition class or executor, it's expanding the existing one to support new inroads.

byrro:
Closing this, since we decided to generate Python code for the wrapper tools, instead of modifying the Worker to read JSON definitions.

PR #546, opened 2025-09-05
upgraded langchain_arcade

torresmateo:
This bumps the requirement version of the Arcade client to the latest version (which supports the new function for the custom user verifier)
It also moves the tests to be compatible to our new tool names


evantahler (commented):
Assuming these tests pass (which we don't have wired up to CI), then 👍 from me!

evantahler (approved):

torresmateo:
They pass!

torresmateo:
Oops... they pass IFF you have poetry. Migrating this to uv on Monday!

PR #551, opened 2025-09-08
Mateo/lchainversion

torresmateo:


.github/workflows/test-toolkits.yml
```diff
@@ -43,6 +43,9 @@ jobs:
 
       - name: Set up the environment
         uses: ./.github/actions/setup-uv-env
+        with:
+          is-toolkit: "true"
```
EricGustin, line 47: do we need to provide is-lib and is-contrib as false?

.github/workflows/test-toolkits.yml
```diff
@@ -80,6 +83,9 @@ jobs:
...
```
EricGustin, line 87: do we need to provide is-lib and is-contrib as false?

.github/workflows/test-toolkits.yml
torresmateo: they all default to false

.github/workflows/test-toolkits.yml
torresmateo: they all default to false

PR #552, opened 2025-09-09
Slack Wrapper Toolkit

byrro:
Slack API endpoints wrapped as Arcade tools. To test the wrapper tools, first install the `arcade-core` package version from [this branch](https://github.com/ArcadeAI/arcade-ai/tree/renato/wrap-worker-surgery), then run the worker.

Wrapper tools should work just like any other Arcade tool, so you should be able to call them directly or use them in the chat interface.


toolkits/slack_api/arcade_slack_api/wrapper_tools/GetSlackUserProfile.json
```diff
@@ -0,0 +1,134 @@
+{
+  "metadata": {
+    "object_type": "api_wrapper_tool",
+    "version": "1.0.0"
+  },
+  "name": "GetSlackUserProfile",
+  "fully_qualified_name": "SlackApi.GetSlackUserProfile@0.1.0",
+  "description": "Retrieve Slack user profile information and custom status.\n\nThis tool fetches the profile information of a Slack user, including their custom status and email (if authorized). It requires appropriate read scopes (users.profile:read or users:read.email).",
```
evantahler, line 8: Can we prompt moar to not list scopes in the descriptions?

evantahler (approved):
Let's Go!!!

There are some rules to add to the tool description generation: 
* Do not include the scopes that are required (the platform handles that)
* Do not mention that this is part of the $service or $service_API (that's implied)

byrro:
Closing this, since we decided to go with the Python implementation, instead of pure-JSON.

PR #575, opened 2025-09-22
[READY][PROD-215][TDK] Adding MS error adapter

jottakka:
# [PROD-215](https://app.clickup.com/t/9014390315/PROD-215) 🎫 

Added:
- MicrosoftGraphErrorAdapter for tools using MS oauth provider.
- Updated the hhtpx error handler to consider 403 + rate limiting exceed headers as a UpstreamRateToolError.


jottakka:
@EricGustin 
image

libs/arcade-tdk/arcade_tdk/providers/microsoft/error_adapter.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # Not a Microsoft Graph SDK error
+        return None
+
+    def _sanitize_uri(self, uri: str) -> str:
+        """Strip query params and fragments from URI for privacy."""
+        parsed = urlparse(uri)
+        return f"{parsed.scheme}://{parsed.netloc.strip('/')}/{parsed.path.strip('/')}"
+
+    def _parse_retry_after(self, error: Any) -> int:
```
byrro: perhaps name this `_get_retry_after_milliseconds`?

libs/arcade-tdk/arcade_tdk/providers/microsoft/error_adapter.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        # Lazy import kiota abstractions to avoid import errors for toolkits that don't use msgraph-sdk
+        try:
+            from kiota_abstractions import api_error
+        except ImportError:
+            logger.info(
+                f"'kiota-abstractions' is not installed in the toolkit's environment, "
+                f"so the '{self.slug}' adapter was not used to handle the upstream error"
+            )
+            return None
```
byrro, line 32: Shouldn't we run the failsafe below in case the `api_error` module is not available?

libs/arcade-tdk/arcade_tdk/providers/microsoft/error_adapter.py
EricGustin, line 1: Is it possible to extract the endpoint and method (see http adapter for an example) from an msgraph error and place it in the error's `extra` dictionary? This can be valuable information. E.g., we would immediately be able to know which endpoint we are sending a bad request to.

EricGustin: nevermind I spoke to soon

PR #577, opened 2025-09-22
[READY][PROD-215][TDK] Adding Slack error adaptor

jottakka:
# [PROD-215](https://app.clickup.com/t/9014390315/PROD-215) 🎫 

Added:
- SlackErrorAdapter for tools using Slack oauth provider.

### DIFF
--- a/libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
+++ b/libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
@@ -41,6 +41,11 @@ class SlackErrorAdapter:
 
         # Failsafe for any unhandled Slack SDK errors that are not mapped above
         if hasattr(exc, "__module__") and exc.__module__ and "slack_sdk" in exc.__module__:
+            logger.warning(
+                "Unknown Slack SDK error encountered: %r. Falling back to generic UpstreamError.",
+                exc,
+                exc_info=True,
+            )
             return UpstreamError(
                 message=f"Upstream Slack SDK error: {exc}",
                 status_code=500,
--- a/libs/arcade-tdk/arcade_tdk/providers/google/error_adapter.py
+++ b/libs/arcade-tdk/arcade_tdk/providers/google/error_adapter.py
@@ -215,6 +215,12 @@ class GoogleErrorAdapter:
 
         # Failsafe for any unhandled Google API client errors that are not mapped above
         if hasattr(exc, "__module__") and exc.__module__ == "googleapiclient.errors":
+            logger.warning(
+                "Unknown Google API client error encountered: %r. "
+                "Falling back to generic UpstreamError.",
+                exc,
+                exc_info=True,
+            )
             return UpstreamError(
                 message=f"Upstream Google API error: {exc}",
                 status_code=500,
--- a/libs/arcade-tdk/arcade_tdk/providers/microsoft/error_adapter.py
+++ b/libs/arcade-tdk/arcade_tdk/providers/microsoft/error_adapter.py
@@ -42,6 +42,12 @@ class MicrosoftGraphErrorAdapter:
             and exc.__module__
             and ("msgraph" in exc.__module__ or "kiota" in exc.__module__)
         ):
+            logger.warning(
+                "Unknown Microsoft Graph SDK error encountered: %r. "
+                "Falling back to generic UpstreamError.",
+                exc,
+                exc_info=True,
+            )
             return UpstreamError(
                 message=f"Upstream Microsoft Graph error: {exc}",
                 status_code=500,


jottakka:
@EricGustin 
image

libs/arcade-tdk/arcade_tdk/error_adapters/utils.py
```diff
@@ -12,4 +12,7 @@ def get_adapter_for_auth_provider(auth_provider: ToolAuthorization | None) -> Er
     if isinstance(auth_provider, Google):
         return GoogleErrorAdapter()
 
+    if isinstance(auth_provider, Slack):
```
byrro, line 21: It's good that we're using class instance here instead of the `provider_id`. All wrapper tools will use the `OAuth2` class by default, even if there's a custom class for the provider. These error adapters should not be used for wrapper tools (because they use raw `httpx` requests, instead of SDKs), so as long as we're not using the `provider_id`, we're fine.

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+        """
+        # Lazy import the Slack SDK errors module to avoid import errors for toolkits that don't use slack-sdk
+        try:
+            from slack_sdk import errors
+        except ImportError:
+            logger.info(
+                f"'slack-sdk' is not installed in the toolkit's environment, "
+                f"so the '{self.slug}' adapter was not used to handle the upstream error"
+            )
+            return None
```
byrro, line 32: Same as in the other PR: shouldn't we use the failsafe below in case the `errors` module cannot be imported?

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+                },
+            )
+
+        # Not a Slack SDK error
+        return None
+
+    def _sanitize_uri(self, uri: str) -> str:
+        """Strip query params and fragments from URI for privacy."""
+
+        parsed = urlparse(uri)
```
byrro: We need a `try/except` block here catching any `Exception` and returning an empty string or `None`, whatever. Otherwise, if the URL parsing fails, the entire error handling crumbles...

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+            if retry_after:
+                try:
+                    # If it's a number, it's seconds
+                    if retry_after.isdigit():
+                        return int(retry_after) * 1000
+                    # Otherwise try to parse as date
+                    dt = datetime.strptime(retry_after, "%a, %d %b %Y %H:%M:%S %Z")
+                    return int((dt - datetime.now(timezone.utc)).total_seconds() * 1000)
+                except Exception:
+                    # TODO: Log?
```
byrro: We should be logging a Warning in cases like this indeed

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+        # Extract error code from Slack API response
+        error_code = "unknown_error"
+        if hasattr(error, "response") and error.response:
+            error_code = error.response.get("error", "unknown_error")
+
+        status_code = 500  # Default to server error
+        if (
+            hasattr(error, "response")
+            and hasattr(error.response, "status_code")
+            and isinstance(error.response.status_code, int)
```
byrro, line 113: could the value possibly be a numeric string as well?

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+        ):
+            status_code = error.response.status_code
+
+        reason = error_code if error_code != "unknown_error" else "Slack API error"
```
byrro: Perhaps use `"Unknown"` instead of `"Slack API error"`. Otherwise the `message` below will be repetitive: `"Upstream Slack API error: Slack API error"`.

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
byrro: Actually, in case of exception, simply returning the original `uri` value seems appropriate.

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+
+        message = f"Upstream Slack API error: {reason}"
+
+        # Build developer message with additional details
+        developer_message = self._build_developer_message(error, error_code)
+
+        # Build extra metadata
+        extra = {
+            "service": self.slug,
```
byrro, line 126: I think we should keep the dictionary keys always consistent. So it should initialize as:

```python
extra = {
    "service": self.slug,
    "endpoint": None,
    "error_code": error_code,
}
```

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+        if warning:
+            developer_details.append(f"warning: {warning}")
+
+        response_metadata = self._extract_response_field(error.response, "response_metadata")
+        if response_metadata and isinstance(response_metadata, dict):
+            warnings = response_metadata.get("warnings", [])
+            if warnings:
+                developer_details.append(f"warnings: {', '.join(warnings)}")
+
+        return " - ".join(developer_details) if len(developer_details) > 1 else developer_details[0]
```
byrro: nit: the expression `" - ".join(developer_details)` will already return `developer_details[0]` in case `len(developer_details) == 1`, so the `if/else` condition is redundant here.

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+            )
+
+        if isinstance(exc, errors_module.SlackTokenRotationError):
+            return UpstreamError(
+                message="Upstream Slack SDK error: Token rotation failed",
+                status_code=401,
+                developer_message=str(exc),
+                extra={
+                    "service": self.slug,
+                    "error_type": "SlackTokenRotationError",
```
byrro: ideally we should reference the class names dynamically with `errors_module.SlackTokenRotationError.__name__`

libs/arcade-tdk/arcade_tdk/error_adapters/utils.py
EricGustin: The httpx adapter is always added to the end of the adapter chain by default those wrappers would still fallback to the HTTP adapter. But still is a nice little performance optimization

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+        result = self._handle_api_errors(exc, errors)
+        if result:
+            return result
+
+        result = self._handle_other_errors(exc, errors)
+        if result:
+            return result
+
+        # Failsafe for any unhandled Slack SDK errors that are not mapped above
+        if hasattr(exc, "__module__") and exc.__module__ and "slack_sdk" in exc.__module__:
```
EricGustin, line 43: what is a scenario where this conditional is true

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
jottakka: it is alway a int here

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
jottakka: Extras is something relate to extra important field for the provider, sometimes the errors they return have not exact the same properties, so extras can be different.

libs/arcade-tdk/arcade_tdk/providers/slack/error_adapter.py
jottakka: If for some reason the SDK is updated with some new error type documented.

PR #586, opened 2025-09-26
Local MCP Fixes and Address General Feedback

EricGustin:
# Release Candidate 2
## This PR:
- [x] No more confusing 307 redirect logs when using `/mcp` instead of `/mcp/` (requested by @shubcodes)
- [x] Fix bug in `arcade configure` for Python < 3.12 (reported by @evantahler 
- [x] Fix bug where tools with unsatisfied secret requirements could still be executed (reported by @evantahler, @shubcodes)
- [x] Auth providers can now be imported via `from arcade_mcp_server.auth import Reddit` (requested by @shubcodes)
- [x] Add complete E2E oauth flow for tool calls with informational errors about how to log into arcade and where to go to authorize (requested by @evantahler, @shubcodes)
- [x] Add OAuth tool in `arcade new`'s generated server (requested by @shubcodes)
- [x] Standardize on defaulting to running servers on port 8000
- [x] Improve credentials.yaml reading logic
- [x] CLI user friendliness (requested by @Spartee)
- [x] Remove `arcade serve` CLI command
- [x] Fix race condition in `arcade logout`
- [x] Update docs for desired developer onboarding flow

## Next PRs:
- Get `arcade deploy` working for MCP servers. (Command is hidden for now)
- Rename  all occurrences of `toolkit` to `server`/`tools` and rename all occurrences of `worker` to `server`


evantahler (commented):
🚀

evantahler (approved):

shubcodes:
@EricGustin thanks for merging the fixes were great! 

I had an idea/worry/thoughts for future as I was testing: in the example that `requires_secret` adding the secret value to the Arcade engine, after adding the mcp server to the arcade engine, will not do anything. This is because, in that setup, arcade acts as an mcp client that simply proxies requests to the mcp server directly - hence the secret is pulled from the `.env ` at the server. My worry is two-fold:

1. for developers, like myself, who have already used `arcade serve` when building a toolkit or read our build a toolkit with secrets doc will think of this as a bug rather than a feature. I think in the future when we support `serve` and `deploy` for mcp-servers we should update the example to use a secret through the arcade engine

2. the secret example not working with the arcade engine, is confusing to understand especially when the `requires_auth` example right after it works just fine with tokens being managed at the arcade engine. 

would love your thoughts on this.

PR #591, opened 2025-09-30
[MOAR][TRELLO] Adding trello tools

jottakka:
No Oauth, uses Oauth1 and needs a Token and a Api key as secrets.


byrro (approved):
@jottakka's first Starter Toolkit! 🎉

PR #593, opened 2025-10-02
CLI Usage

EricGustin:
TLDR; 

The philosophy of CLI usage is "fire and forget" and "best effort". You can opt out by setting `ARCADE_USAGE_TRACKING=0`.

We are capturing two events: `CLI execution succeeded` and `CLI execution failed`. Reporting to PostHog is a short lived (maximum 10 seconds) subprocess that does not block the main CLI execution process.

`~/.arcade/usage.json` persists two values `anon_id` and `linked_principal_id`. The logged in status of the CLI user determines which ID is used. Upon `arcade login`, the `anon_id` is aliased with `linked_principal_id`. Upon `arcade logout` the `linked_principal_id` is removed and the `anon_id` is rotated.

## CLI Usage Tracking - How It Works

The usage tracking system implements an identity management and event tracking pipeline. Here's how the pieces work together:

### **Identity State Management (`usage.json`)**

The system maintains a persistent identity file at `~/.arcade/usage.json` with this structure:
```json
{
  "anon_id": "uuid",
  "linked_principal_id": "uuid" | null
}
```

**Key mechanics:**
- **`anon_id`**: Generated once on first CLI use and persists across sessions. This UUID tracks all anonymous activity.
- **`linked_principal_id`**: Initially `null`. Once the user logs in and we successfully alias their identity, this field stores their `principal_id` to indicate this `anon_id` has been linked.
- **Atomic writes**: All updates use a temp file + atomic rename pattern to prevent corruption from concurrent CLI processes
- **File locking**: Uses `fcntl` (Unix) to coordinate reads/writes across multiple simultaneous CLI invocations
- **In-memory cache**: The `UsageIdentity` class caches the loaded data to avoid repeated file I/O within a single CLI invocation

### **Identity Resolution Flow**

When tracking an event, the system determines the `distinct_id` (who to attribute the event to) via this waterfall:

1. **Check `linked_principal_id`** in `usage.json`
   - If present → use it (user was previously aliased)
   - This is the fastest path and avoids API calls

2. **Fetch `principal_id` from Arcade Cloud API**
   - Makes HTTP request to `/api/v1/auth/validate` with the user's API key from `~/.arcade/credentials.yaml`
   - If authenticated → returns `principal_id`
   - Has 2s timeout for responsiveness

3. **Fall back to `anon_id`**
   - If not authenticated or API call fails → use anonymous ID
   - Marks event with `is_anon=True` flag

### **The Aliasing Lifecycle**

PostHog aliasing links anonymous activity to authenticated users. Here's the state machine:

#### **Stage 1: Anonymous User**
```
usage.json: { "anon_id": "abc-123", "linked_principal_id": null }
All events → sent with distinct_id="abc-123" and is_anon=True
```

#### **Stage 2: Login Event**
1. User runs `arcade login`
2. Command completes successfully (auth token saved)
3. `CommandTracker` detects successful login
4. Fetches `principal_id` from API
5. Checks `should_alias()` → returns `True` because `linked_principal_id` is `null`
6. **Calls `alias()` synchronously** (blocking):
   ```python
   posthog.alias(previous_id="abc-123", distinct_id="zyx-321")
   ```
7. Updates `usage.json`:
   ```json
   { "anon_id": "abc-123", "linked_principal_id": "zyx-321" }
   ```
8. PostHog backend merges all events with `distinct_id="abc-123"` into the user profile for `"zyx-321"`

#### **Stage 3: Authenticated User**
```
usage.json: { "anon_id": "abc-123", "linked_principal_id": "zyx-321" }
All events → sent with distinct_id="zyx-321" and is_anon=False
```
- Events are directly attributed to the authenticated user
- No more API calls needed (uses cached `linked_principal_id`)

#### **Stage 4: Logout Event**
1. User runs `arcade logout`
2. Logout event is sent with the authenticated `distinct_id`
3. `CommandTracker` detects successful logout
4. **Rotates identity** by calling `reset_to_anonymous()`:
   ```json
   { "anon_id": "xyz-789", "linked_principal_id": null }
   ```
5. New `anon_id` prevents cross-contamination if another user logs in

### **Critical Constraint: Alias Timing**

PostHog requires that `alias()` is called **BEFORE** any events are sent with the new `distinct_id`. This is why:
- **`alias()` is synchronous (blocking)**: Guarantees it completes before the login success event is sent
- **Subsequent events use `linked_principal_id`**: Once aliased, all future events use the authenticated ID
- **Lazy aliasing**: If a user authenticates via another mechanism (not through `arcade login`), the system detects this on the next command and performs aliasing before sending that command's event

### **Event Capture Pipeline**

When `CommandTracker.track_command_execution()` is called:

1. **Resolve identity** → determines `distinct_id` and `is_anon` flag
2. **Build event properties**:
   ```python
   {
     "command_name": "toolkit.run",
     "cli_version": "1.2.3",
     "python_version": "3.11.0",
     "os_type": "Darwin",
     "os_release": "23.4.0",
     "duration": 1250.42,  # milliseconds
     "error_message": "..."  # if failed
   }
   ```
3. **Call `UsageService.capture()`**:
   - Serializes event data to JSON
   - Spawns detached subprocess: `python -m arcade_cli.usage`
   - Passes data via `ARCADE_USAGE_EVENT_DATA` env var
   - **Returns immediately** (non-blocking)

4. **Detached subprocess (`__main__.py`)**:
   - Runs independently, survives parent CLI exit
   - Deserializes event data
   - If `is_anon=True`, sets `$process_person_profile=False` (tells PostHog not to create a full profile)
   - Sends event to PostHog with 5s timeout
   - Exits (hard exit after 10s max via timeout thread)

### **Concurrency Handling**

Multiple CLI processes can run simultaneously. The system handles this via:
- **File locking** on `usage.json` (shared lock for reads, exclusive for writes)
- **Atomic writes** via temp files ensure incomplete writes never corrupt the file
- **Idempotent aliasing**: `should_alias()` prevents redundant alias calls

### **Edge Cases Handled**

1. **Side-channel authentication**: User authenticates outside of `arcade login` (e.g., manually editing credentials)
   - Detected via "lazy aliasing" check on every command
   - Performs alias if `linked_principal_id` doesn't match current `principal_id`

2. **API failures during identity fetch**: Falls back to anonymous tracking
   - 2s timeout prevents hanging
   - Silent failure doesn't disrupt CLI

3. **PostHog merge restrictions**: Can't alias returning users who already have a profile
   - System stores `linked_principal_id` to avoid retrying impossible aliases
   - New users (never logged in before) get full history stitched

4. **Multiple accounts on same machine**: Logout rotates `anon_id`
   - User A's anonymous activity won't leak into User B's profile

### **Privacy & Performance**

- **Opt-out**: `ARCADE_USAGE_TRACKING=0` disables all tracking
- **Non-blocking**: Events never slow down CLI (detached subprocess)
- **Anonymous profiles**: `$process_person_profile=False` for `anon_id` events minimizes data collection
- **Silent failures**: Network issues or PostHog errors never surface to users

### DIFF
--- a/libs/arcade-cli/arcade_cli/usage/command_tracker.py
+++ b/libs/arcade-cli/arcade_cli/usage/command_tracker.py
@@ -1,4 +1,5 @@
 import functools
+import os
 import platform
 import sys
 import time
@@ -6,6 +7,7 @@ from importlib import metadata
 from typing import Any
 
 import typer
+from arcade_cli.constants import ARCADE_CONFIG_PATH
 from arcade_cli.usage.constants import (
     EVENT_CLI_COMMAND_EXECUTED,
     EVENT_CLI_COMMAND_FAILED,
@@ -23,8 +25,11 @@ from arcade_cli.usage.constants import (
 from arcade_cli.usage.identity import UsageIdentity
 from arcade_cli.usage.usage_service import UsageService
 from arcade_cli.usage.utils import is_tracking_enabled
+from rich.console import Console
 from typer.core import TyperCommand, TyperGroup
 from typer.models import Context
+
+console = Console()
 
 
 class CommandTracker:
@@ -217,6 +222,13 @@ class TrackedTyperCommand(TyperCommand):
 
     def invoke(self, ctx: Any) -> Any:
         """Override invoke to track command execution."""
+        if not os.path.exists(ARCADE_CONFIG_PATH):
+            console.print(
+                "[yellow]Arcade collects CLI usage data to help us debug and improve the service. "
+                "By continuing to use the Arcade CLI, you agree to the terms of our Privacy Policy. "
+                "To opt out, set the ARCADE_USAGE_TRACKING environment variable to 0.[/yellow]"
+            )
+
         command_name = ctx.command.name
         is_login = command_name == "login"
         is_logout = command_name == "logout"


libs/arcade-cli/arcade_cli/main.py
```diff
@@ -271,16 +271,11 @@ def mcp(
             handle_cli_error("Failed to run MCP server")
 
     except KeyboardInterrupt:
-        console.print("\n[yellow]MCP server stopped[/yellow]")
-        raise typer.Exit(0)
+        console.print("\n[yellow]MCP server gracefully shutdown[/yellow]")
     except FileNotFoundError:
-        console.print(
-            "[red]arcade_mcp_server module not found. Make sure arcade-mcp-server is installed.[/red]"
+        handle_cli_error(
```
EricGustin, line 276: As part of this PR I improved how our CLI commands are handling errors to ensure the correct data is sent to posthog

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -271,16 +271,11 @@ def mcp(
             handle_cli_error("Failed to run MCP server")
 
     except KeyboardInterrupt:
-        console.print("\n[yellow]MCP server stopped[/yellow]")
-        raise typer.Exit(0)
+        console.print("\n[yellow]MCP server gracefully shutdown[/yellow]")
```
EricGustin, line 274: Since `arcade mcp` is a long running process and keyboard interrupt is the only way to gracefully shutdown, we don't want to raise a typer exit. This ensures that the event is correctly sent to posthog as a 'success'

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -44,8 +44,8 @@
     version_callback,
 )
 
-cli = typer.Typer(
-    cls=OrderCommands,
+cli = TrackedTyper(
```
EricGustin, line 47: our CLI now uses `TrackedTyper` and `TrackedTyperGroup`. These are wrappers around `typer.Typer` and `OrderCommands` such that we can include usage tracking after successful/failed invocation of the CLI command. All tracking logic is outside of the commands themselves.

libs/arcade-cli/arcade_cli/usage/__main__.py
```diff
@@ -0,0 +1,62 @@
+"""Entry point for detached usage tracking subprocess.
```
EricGustin, line 1: This file is the detached subprocess that actually sends the events to posthog. Gets its payload via env var.

libs/arcade-cli/arcade_cli/usage/command_tracker.py
```diff
@@ -0,0 +1,305 @@
+import functools
```
EricGustin, line 1: This file is what wraps all CLI commands to track their execution without needing to explicitly track inside of command implementations

libs/arcade-cli/arcade_cli/usage/identity.py
```diff
@@ -0,0 +1,207 @@
+"""
```
EricGustin, line 1: This is the file that handles the user's identity state throughout authentication lifecycle

libs/arcade-cli/arcade_cli/usage/usage_service.py
```diff
@@ -0,0 +1,104 @@
+import json
```
EricGustin, line 1: This is the file that **manages** the communication with Posthog

libs/arcade-cli/arcade_cli/utils.py
```diff
@@ -80,6 +79,23 @@ class Provider(str, Enum):
     OPENAI = "openai"
 
 
+class CLIError(Exception):
```
EricGustin, line 82: Triggers a 'failed' event for the command tracker. Will eventually resolve to an exit code == 1 after the error message is extracted for tracking.

evantahler (approved):
📈

PR #597, opened 2025-10-02
Update arcade-mcp-server docs

EricGustin:
1. Updates docs to prefer `uv run server.py` instead of `arcade mcp` or `python -m arcade_mcp_server`
2. Found a bug with running stdio servers while updating the docs, so i snuck that in this PR


libs/arcade-mcp-server/arcade_mcp_server/mcp_app.py
```diff
@@ -226,7 +226,6 @@ def run(
             asyncio.run(
                 run_stdio_server(
                     catalog=self._catalog,
-                    host=host,
```
EricGustin, line 229: the bug fix I'm sneaking in.

evantahler (approved):

evantahler:
Note: I think we are missing a "use tools with auth" example

EricGustin:
@evantahler good point. Adding...

PR #603, opened 2025-10-06
Xero API Starter MCP server

byrro:
Combines tools auto-generated from the OpenAPI specs for the [Accounting](https://github.com/XeroAPI/Xero-OpenAPI/blob/master/xero_accounting.yaml) & [Identity](https://github.com/XeroAPI/Xero-OpenAPI/blob/master/xero-identity.yaml) APIs.

The Identity API is necessary to retrieve the currently available tenants, since many endpoints in the Accounting API require a tenant ID.


toolkits/xero_api/pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+build-backend = "hatchling.build"
+
+[project]
+name = "arcade_xero_api"
+version = "0.1.0"
+description = "Tools that enable LLMs to interact directly with the Xero API."
+requires-python = ">=3.10"
+dependencies = [
+    "arcade-tdk>=2.6.0rc2,<3.0.0",
+    "httpx>=0.27.2,<1.0.0",
```
jottakka:
```suggestion
   "httpx[http2]>=0.27.2,<1.0.0",
```
Otherwise it wont work when using "arcade mcp"

jottakka (commented):
Just the project.toml needs to be updated to use http 2.0

PR #614, opened 2025-10-10
Fix server info bug

EricGustin:
Name, title, version, etc. for an `MCPApp` were being overwritten by its internal `MCPServer`.

⚠️ this is blocking `arcade deploy` from working

### DIFF
--- a/libs/arcade-mcp-server/arcade_mcp_server/server.py
+++ b/libs/arcade-mcp-server/arcade_mcp_server/server.py
@@ -156,9 +156,7 @@ class MCPServer:
             self.title = self.name
 
         self.instructions = (
-            instructions
-            if instructions
-            else (self.settings.server.instructions or self._default_instructions())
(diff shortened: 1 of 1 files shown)


libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -136,17 +136,31 @@ def __init__(
...
+        elif (
+            self.settings.server.title
+            and self.settings.server.title != ServerSettings.model_fields["title"].default
+        ):
+            self.title = self.settings.server.title
+        else:
+            self.title = self.name
+
+        self.instructions = (
+            instructions
```
byrro: nit: I think this could be simplified to `instructions or self.settings.server.instructions or self._default_instructions()`

EricGustin: Thanks, updated!

PR #618, opened 2025-10-13
`arcade deploy` for MCP Servers

EricGustin:
Blocked by https://github.com/ArcadeAI/arcade-mcp/pull/614 (and the reason for failing tests)

# PR Description
`arcade deploy` will deploy your local MCP server to Arcade. `arcade deploy` should be executed at the root of your MCP Server package. Before deploying, the command runs your server locally to ensure your project is setup correctly and the server runs properly. `arcade deploy` assumes your entrypoint file will execute `MCPApp.run` when the file is invoked directly. This means you must either have an `if __name__ == "__main__" block that contains `MCPApp.run`, or `MCPApp.run` should be top-level code (unindented living directly in the body of the file).

image

image

### DIFF
--- a/libs/arcade-cli/arcade_cli/deploy.py
+++ b/libs/arcade-cli/arcade_cli/deploy.py
@@ -14,6 +14,7 @@ from dotenv import load_dotenv
 from pydantic import BaseModel, Field
 from rich.console import Console
 
+from arcade_cli.secret import load_env_file
 from arcade_cli.utils import compute_base_url, validate_and_get_config
 
 console = Console()
@@ -465,6 +466,10 @@ def deploy_server_to_engine(
 
 def deploy_server_logic(
     entrypoint: str,
+    skip_validate: bool,
+    server_name: str | None,
+    server_version: str | None,
+    secrets: str,
     host: str,
     port: int | None,
     force_tls: bool,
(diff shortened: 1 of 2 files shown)


libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -0,0 +1,582 @@
...
+    """
+    Start the server, verify it's healthy, and extract metadata.
+
+    This function orchestrates:
+    1. Starting the server on a random port
+    2. Waiting for the server to become healthy
+    3. Extracting server name and version via POST /mcp (initialize method)
+    4. Extracting required secrets via GET /worker/tools
+    5. Stopping the server
+    6. Returning the metadata
```
evantahler, line 328: I think you should log every step as this is happening out:
```
starting server on port 123 to test... OK
extracting tools... OK
```

This way if something fails it will be very clear what went wrong.

evantahler: Please add a `--skip-server-check` or similar arg.  I might not be able to fully boot the server locally if there is some dependency missing (a dataabase_connection_url ?), but i still might want to `sudo deploy`

libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -0,0 +1,582 @@
...
+
+    Args:
+        entrypoint: Path to the entrypoint file that runs the MCPApp instance
```
evantahler, line 331:
```suggestion
        entrypoint: Path to the entrypoint file that runs the MCPApp instance.  This file must contain a __main__ method that runs MCPApp()
```

evantahler: Give the users hints as to what an entrypoint is

libs/arcade-cli/arcade_cli/deploy.py
EricGustin: Hmm
1. the `GET /worker/tools` request is how I determine which secrets need to be upserted before deploying
2. the `POST /mcp` with `initialize` method is how I determine the server's name and version

I guess if someone wants to "sudo deploy", then they'll need to explicitly provide the `server_name` and `server_version`. For the secrets I could either skip the upserting step (assume they've already upserted on their own) or just always upsert every secret in their `.env` regardless of whether the server need them

EricGustin: updated

evantahler (approved):
nice.

PR #621, opened 2025-10-13
[MOAR][FIGMA]  Figma Starter Toolkit

jottakka:


EricGustin (changes requested):
moving forward we will use 

Deps:
"arcade-tdk>=3.0.0,<4.0.0",

Dev deps:
"arcade-mcp[all]>=1.2.0,<2.0.0",
"arcade-serve>=3.0.0,<4.0.0",

PR #627, opened 2025-10-16
Update toolkit deps

EricGustin:


evantahler:
Looks like Brightdata needs to be bumped as well

EricGustin:
@evantahler good catch, updated

evantahler (approved):

PR #628, opened 2025-10-16
Updating Brightdata community pkg

jottakka:
Updating BrightData 
- Updating project.toml
- Fix linting issues (related to the repo configs)s
- Rename package from brightdata -> arcade-brightdata ( also will be used by PyPI)
- Added to toolkits.txt so it can be deployed

Extra:
- Arcade new templates did not have the extra line at the end, so it has been added.

### DIFF
--- a/toolkits/brightdata/tests/test_brightdata.py
+++ b/toolkits/brightdata/tests/test_brightdata.py
@@ -2,6 +2,7 @@ from os import environ
 from unittest.mock import Mock, patch
 
 import pytest
+import requests
 from arcade_tdk import ToolContext, ToolSecretItem
 from arcade_tdk.errors import ToolExecutionError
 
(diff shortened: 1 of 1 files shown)


libs/arcade-cli/arcade_cli/templates/full/{{ toolkit_name }}/evals/eval_{{ toolkit_name }}.py
jottakka, line 1: All files in this folder just got a extra line at the end due ruff rules.

docker/toolkits.txt
```diff
@@ -14,3 +15,4 @@ arcade-xero-api
 arcade-zendesk
 arcade-zoho-books-api
 arcade-zoho-creator-api
+brightdata
```
EricGustin: This will need to go away

toolkits/brightdata/arcade_brightdata/bright_data_client.py
```diff
@@ -54,9 +54,11 @@ def make_request(self, payload: dict) -> str:
         )
 
         if response.status_code != 200:
-            raise Exception(f"Failed to scrape: {response.status_code} - {response.text}")  # noqa: TRY002
+            msg = f"Failed to scrape: {response.status_code} - {response.text}"
```
EricGustin: should use `raise_for_status` so that it can be adapted

PR #632, opened 2025-10-17
arcade docs command fixes and concurrency improvement

byrro:


EricGustin:
closing since this command was moved to docs

PR #634, opened 2025-10-18
Improve `arcade deploy` CLI Command

EricGustin:
Also fleshed out `arcade server` commands and MCPApp.name validation.

Example of output of `arcade deploy`:
image


libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -63,15 +71,221 @@ class DeploymentToolkits(BaseModel):
...
+    """
+    Get the status of a deployment.
+
+    Args:
+        engine_url: The base URL of the Arcade Engine
+        server_name: The name of the server to get the status of
+
+    Returns:
+        The status of the deployment.
+        Possible values are: "pending", "unknown", "running", "failed".
```
sdreyer: and updating

sdreyer (approved):

PR #641, opened 2025-10-21
Disallow executing auth/secret tools for unauthenticated servers using HTTP transport

EricGustin:
## PR Description
This PR tackles 3 things:
1. At tool execution runtime, blocks local HTTP servers from executing tools that have `requires_auth` or `requires_secrets`
2. Make `stdio` the default transport in various locations
3. Improve the `arcade configure` CLI command

image
image

### DIFF
--- a/libs/arcade-cli/arcade_cli/configure.py
+++ b/libs/arcade-cli/arcade_cli/configure.py
@@ -245,7 +245,7 @@ def configure_vscode_local(
 ) -> None:
     """Configure VS Code to add a local MCP server to the configuration."""
 
-    def http_config(server_name: str, port: int = 8000) -> dict:
+    def http_config(port: int = 8000) -> dict:
         return {
             "type": "http",
             "url": f"http://localhost:{port}/mcp",
@@ -278,7 +278,7 @@ def configure_vscode_local(
     config["servers"][server_name] = (
         get_stdio_config(entrypoint_file, server_name)
         if transport == "stdio"
-        else http_config(server_name, port)
+        else http_config(port)
     )
 
     # Write updated config


libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -754,8 +754,28 @@ async def _check_tool_requirements(
...
     ) -> JSONRPCResponse[CallToolResult] | None:
         """Check tool requirements before executing the tool"""
+        # Check transport restrictions for tools requiring auth or secrets
+        if session and session.init_options:
+            transport_type = session.init_options.get("transport_type")
+            if transport_type != "stdio":
+                requirements = tool.definition.requirements
+                if requirements and (requirements.authorization or requirements.secrets):
+                    tool_response = {
+                        "message": (
```
EricGustin, line 768: I'm open to suggestions on how we want to describe this intentional restriction of http servers to the client. For the local case, the end-user is typically the server developer too, so I feel comfortable using server specific language

evantahler: I think you should say less here, and link folks to a page in the docs about this where we can expound more, include your chart, etc.

libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -754,8 +754,28 @@ async def _check_tool_requirements(
...
+                            f"Tool '{tool_name}' cannot be executed over HTTP transport for security reasons. "
+                            "This tool requires end-user authorization or access to sensitive secrets, and therefore "
+                            "must not run over HTTP. This restriction is permanent - retrying the call will not resolve the issue.\n\n"
```
evantahler:
```suggestion
                            "must not run over unauthenticated HTTP. This restriction is permanent - retrying the call will not resolve the issue.\n\n"
```

libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -754,8 +754,28 @@ async def _check_tool_requirements(
...
+                            "To use this tool, the server developer must either:\n"
+                            "1) Use the stdio transport (recommended for local development and desktop clients),\n"
+                            "2) Deploy the server using 'arcade deploy', or\n"
```
evantahler:
```suggestion
                            "2) Deploy the server using 'arcade deploy' which supports per-user OAuth authentication, or\n"
```

libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -754,8 +754,28 @@ async def _check_tool_requirements(
...
+                            "3) Register the server in the Arcade Developer Dashboard with type 'Arcade'."
```
evantahler:
```suggestion
                            "3) Register the server in the Arcade Developer Dashboard with type 'Arcade' (rather than MCP), which also supports per-user OAuth."
```

EricGustin:
@torresmateo for eyes on changes to `arcade configure` and `arcade new`

libs/arcade-cli/arcade_cli/configure.py
```diff
@@ -51,14 +54,21 @@ def get_vscode_config_path() -> Path:
         return Path.home() / ".config" / "Code" / "User" / "mcp.json"
 
 
-def configure_claude_local(server_name: str, port: int = 8000, path: Path | None = None) -> None:
-    """Configure Claude Desktop to add a local MCP server to the configuration."""
-    config_path = path or get_claude_config_path()
-    config_path.parent.mkdir(parents=True, exist_ok=True)
+def is_uv_installed() -> bool:
```
byrro, line 57: I was gonna say we should cache the response here but `shutil.which` is surprisingly fast

libs/arcade-cli/arcade_cli/configure.py
```diff
@@ -145,18 +219,44 @@ def configure_cursor_local(server_name: str, port: int = 8000, path: Path | None
...
+    entrypoint_file: str,
+    server_name: str,
+    transport: str,
+    port: int = 8000,
+    config_path: Path | None = None,
+) -> None:
     """Configure VS Code to add a local MCP server to the configuration."""
-    config_path = path or get_vscode_config_path()
+
+    def http_config(server_name: str, port: int = 8000) -> dict:
```
byrro: nit: looks like the `server_name` arg is not being used inside the function

byrro (approved):

libs/arcade-cli/arcade_cli/configure.py
EricGustin: addressed

PR #657, opened 2025-10-28
Fix MCP capabilities, examples, tests, and more

EricGustin:
# PR Description
Consider this PR the result of a full pass through of this repository.
## Add helper for adding tools to an `MCPApp`
You can now add all of the tools in a module to an `MCPApp` via `app.add_tools_from_module(...)`
## Edit what `arcade new` generates
First, I updated the backend to use hatchling.

Second, the structure generated before this PR was simple, but did not create a proper Python module.
This hindered developers in the following ways:
1. Difficult to add the tools in your server to an evaluation suite
2. Difficult to add more than one tool to an MCPApp at a time
3. All other niceties that come with being able to import modules
```
# Before
server/
├── .env.example
├── server.py
└── pyproject.toml
```
This PR updates the structure generated such that a valid Python module is generated:
```
# After 
server/
├── pyproject.toml
└── src/
    └── server/
        ├── __init__.py
        ├── .env.example
        └── server.py
```
## Fix Tool Chaining
`self._ctx.server.executor.run(...)` was being called, but `MCPServer` does not have an instance of `ToolExecutor` (and it's not intended to be an instance anyways). I updated `Tool.call_raw` to pass the programmatic tool call through the `MCPServer._handle_call_tool`. This means that the programmatic tool calls now go through the same steps that a typical tool call (initiated by the MCP client) would.

This means that **toolA**, which specifies **requirementsA**, is permitted to call **toolB**, which specifies **requirementsB**, without needing to explicitly declare or satisfy **requirementsB**. I believe this is acceptable because the secrets and/or auth token associated with **toolB's** `Context` are not exposed to **toolA**, and the secrets and/or auth token associated with **toolA's** `Context` are not exposed to **toolB**.

## Fix User Elicitation
1. The read & write streams were created with a maximum queue size of 0. I increased this to 100.
2.  I updated `ServerSession`'s run loop to both read messages from the stream & process them concurrently. This enables server initiated requests (like user elicitation and progress reporting) to be handled while tools are being executed. Otherwise, the server initiated requests would wait for the tool to finish executing and the tool execution would wait for the server initiated request to finish.
3. 
## Fix Progress Reporting
Progress tokens sent by the client were not being stored. Therefore there was no way to notify a client with progress updates. I am now storing the `progressToken`, along with other `_meta` sent from the client, in the `ServerSession`'s `_request_meta`. I am setting `_request_meta` whenever the `MCPServer` is handling an incoming message from a client. 

## Fix handling of server names with spaces
Before: 
Server name: "The simple server name"
Tool name: whisper_secret
Name seen by client: "The_simple_server_name_WhisperSecret"

After
Server name: "The simple server name"
Tool name: whisper_secret
Name seen by client: "TheSimpleServerName_WhisperSecret"

## Add Integration Tests
The stdio integration test is much more comprehensive than the http integration test. These tests will let me sleep a bit more at night

## Add Example MCP Servers
Example servers for sampling, user-elicitation, progress reporting, logging, tool chaining, combining prebuilt tools with custom tools, tool secrets, tool auth, evaluations, and more!

## Add Docker template
Added a Docker template for running an MCP server in Docker (and removed the old docker stuff)


examples/docker-template/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```bash
+cd your-server-name
+
+# Build and run with docker-compose
+docker-compose -f docker/docker-compose.yml up --build
+
+# Or build manually
+docker build -f docker/Dockerfile -t your-server .
+docker run -p 8001:8001 -e PACKAGE_NAME=your-package-name your-server
```
evantahler: what does `PACKAGE_NAME` do here, and why do I need this if I am running my server.py?

examples/docker-template/docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ghcr.io/astral-sh/uv:python3.11-bookworm-slim
+
```
evantahler, line 2: Should we run as a non-root user?
cc @sdreyer

examples/docker-template/docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Define build arguments
+ARG PACKAGE_NAME=logging
+
+# Set environment variables
+ENV PACKAGE_NAME=${PACKAGE_NAME}
```
evantahler: This `PACKAGE_NAME` thing seems brittle (case sensitivity, spaces, etc).  Can we assume that there's only one pkg within src and run it?  Can we fail if there are more than one  pkg in src?

examples/mcp_servers/logging/docker/Dockerfile
evantahler, line 1: If these files will be duplicated, maybe a sumlink vs a copy?

examples/mcp_servers/logging/src/logging/server.py
```diff
@@ -0,0 +1,22 @@
...
+    await context.log.info(f"Info via log.info: {message}")
+    await context.log.warning(f"Warning via log.warning: {message}")
+    await context.log.error(f"Error via log.error: {message}")
+    await context.log.log("info", f"Info via log.log: {message}")
+
+    return message
+
+
+if __name__ == "__main__":
+    app.run(transport="http")
```
evantahler, line 24: How should we log outside of a tool call (e.g. lifecycle messages like "server booted on port X")?

examples/docker-template/README.md
EricGustin: removed the need for PACKAGE_NAME

examples/mcp_servers/logging/src/logging/server.py
EricGustin: added in the example

examples/mcp_servers/logging/docker/Dockerfile
EricGustin: will stick to copy so they can be customized

examples/docker-template/docker/Dockerfile
EricGustin: made it non root

PR #661, opened 2025-10-30
Track whether a tool call event happened

EricGustin:
We are now tracking whether a tool call event happens. We track generic "failure reasons" if the tool call fails. We DO NOT track names of tools, tool parameters, or any PII.

Event name: 
- MCP tool called

Properties:
- is_execution_success
- failure_reason - one of "missing requirements", "transport restriction", "error during tool execution", "unknown tool", "internal error calling tool" or doesn't exist in the case of successful tool execution.
- arcade_mcp_server_version
- runtime_language
- os_type
- os_release
- device_timestamp

As always you can opt out via setting the `ARCADE_USAGE_TRACKING` environment variable to 0.

### DIFF
--- a/libs/arcade-mcp-server/pyproject.toml
+++ b/libs/arcade-mcp-server/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "arcade-mcp-server"
-version = "1.5.0"
+version = "1.6.0"
 description = "Model Context Protocol (MCP) server framework for Arcade.dev"
 readme = "README.md"
 authors = [{ name = "Arcade.dev" }]


libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -659,10 +663,18 @@ async def _handle_call_tool(
             # Create tool context
             tool_context = await self._create_tool_context(tool, session)
 
+            # Check restrictions for unauthenticated HTTP transport
```
EricGustin, line 673: checking restrictions for unauthenticated HTTP transport used to be lobbed in together with checking tool requirements. In this PR I move it into its own function so that we can distinguish between failure reasons "missing requirements" and "transport restriction"

libs/arcade-mcp-server/arcade_mcp_server/usage/constants.py
```diff
@@ -1,9 +1,12 @@
 # MCP Server Specific Event Names
 EVENT_MCP_SERVER_STARTED = "MCP server started"
+EVENT_MCP_TOOL_CALLED = "MCP tool called"
```
evantahler, line 3: The ship has already sailed, but event names with spaces seem odd to me

libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -659,10 +663,18 @@ async def _handle_call_tool(
...
+                tool, tool_context, message, tool_name, session
+            ):
+                self._tracker.track_tool_call(False, "transport restriction")
+                return transport_restriction_response
+
             # Handle authorization and secrets requirements if required
             if missing_requirements_response := await self._check_tool_requirements(
                 tool, tool_context, message, tool_name, session
             ):
+                self._tracker.track_tool_call(False, "missing requirements")
```
evantahler, line 684: nit: there are a lot of place to call track_tool_call... we might forget one. Fine for now

evantahler (approved):

PR #679, opened 2025-11-05
Better Handling of MCP-specific `Context` usage for managed servers

EricGustin:
Since servers managed by Arcade use the `/worker` routes under the hood, tools that use MCP-specific properties of `Context` will fail.

This PR helps reduce the 'blast radius' of the above fact. For properties that were deemed 'non-critical' to the execution of a deployed tool, we simply no-op. For properties that were deemed 'critical' to the execution of a deployed tool, we raise an error that informs the caller that the feature is not supported for Arcade managed servers.

- Non-critical property: A context property that returns None
- Critical property: A context property that may return something that could be necessary for a tool execution to succeed.

### DIFF
--- a/libs/arcade-core/pyproject.toml
+++ b/libs/arcade-core/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "arcade-core"
-version = "3.3.3"
+version = "3.3.4"
 description = "Arcade Core - Core library for Arcade platform"
 readme = "README.md"
 license = {text = "MIT"}


Spartee:
This PR introduces clear degradation strategies for MCP-specific features in managed (Arcade) server contexts, distinguishing features that should silently "no-op" (logs, progress) from those that must raise errors (resources, tools, prompts, etc.). The approach is well-tested and aligns with the intended user experience for most deployments. However, I have identified three areas where improvements could further enhance developer experience, robustness, and future extensibility.

---

## 1. No User/Developer Feedback on No-Op Features

**Issue:**
Currently, when running in a managed server environment, calls to `context.log` and `context.progress` are silently ignored via no-op implementations. While this reduces user confusion, it can make it very difficult for developers to debug issues where expected logs or progress updates are not appearing.

**Evidence:**
- Both `_NoOpLogs` and `_NoOpProgress` swallow all messages without any feedback.
- There is no mechanism to alert developers (even once, or in debug mode) that they are operating in a degraded environment.

**Suggestion:**
- Add an optional warning or info message (emitted once per process, or configurable via an environment variable) when a no-op logger/progress is first used.
- Alternatively, provide a debug flag (`ARCADE_DEBUG_NOOP=1`) that enables these warnings for development/testing environments.
- Document this behavior in the developer docs.

---

## 2. Lack of Programmatic Feature Detection/Introspection

**Issue:**
There is no way for tool or agent code to check which MCP features are available in the current context. Developers must rely on catching RuntimeErrors, which is less robust and can lead to brittle code or poor user experience.

**Evidence:**
- All feature access is binary: either the property works, is a no-op, or throws.
- No `context.has_feature("tools")`/`context.available_features` API exists.

**Suggestion:**
- Add a method or property to `ToolContext`, e.g., `def has_feature(self, feature_name: str) -> bool` or `context.available_features`, listing which features are supported in the current runtime.
- This allows tool code to adapt gracefully or provide more informative messages to end users.

---

## 3. Error Signaling Uses Only Generic RuntimeError

**Issue:**
All unsupported critical features raise a generic `RuntimeError` with a string message. This limits the ability of developers or frameworks to catch and handle specific feature failures programmatically.

**Evidence:**
- Each property (e.g., `resources`, `tools`, `prompts`, etc.) raises a plain `RuntimeError` without a unique exception type or error code.

**Suggestion:**
- Define a custom exception class, e.g., `class ToolContextFeatureUnavailableError(RuntimeError):`, which accepts the feature name and a message.
- Raise this exception in each unsupported property with an identifying feature name.
- This makes it easier for callers to distinguish between different kinds of unsupported feature errors and to implement fallback or error handling logic as needed.

---

## Summary Table

| Issue                 | Impact         | Suggestion                                       |
|-----------------------|---------------|--------------------------------------------------|
| No-op is silent       | Med/DevX      | Add warning/debug flag for no-ops                |
| No introspection      | Med/DevX      | Add `has_feature`/`available_features` to context|
| Generic RuntimeError  | Low-Med/Code  | Use custom exception per feature                 |

---

**Overall:**
This PR is a strong step toward robust, explicit handling of MCP context features in managed environments. Addressing these three issues would make the system both more developer-friendly and future-proof, especially as more features or runtimes are added.

Posted by Cade Automation

PR #680, opened 2025-11-06
Remove tools from the arcade wrapper toolkit that should not pass data through an LLM

evantahler:


evantahler:
Thanks! Although I can't approve my own PR ;p

jottakka (approved):

PR #683, opened 2025-11-07
Write catalog to disk (off by default)

EricGustin:
Writes catalog to `/tmp/arcade_worker_cache/catalog_<hash>.json` on startup and then returns the contents of that file whenever `/worker/tools` is called. Off by default. Set `CACHE_WORKER_CATALOG=true` to enable.

### DIFF
--- a/libs/arcade-serve/arcade_serve/core/base.py
+++ b/libs/arcade-serve/arcade_serve/core/base.py
@@ -108,10 +108,8 @@ class BaseWorker(Worker):
             except OSError as e:
                 logger.warning(f"Failed to remove old cache file {self._catalog_cache_path}: {e}")
 
-        # Compute hash of catalog contents
         catalog_hash = self._compute_catalog_hash()
 
-        # Create cache directory if it doesn't exist
         os.makedirs(self._cache_dir, exist_ok=True)
 
         # Write cache file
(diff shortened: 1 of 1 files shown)


libs/arcade-serve/arcade_serve/core/base.py
```diff
@@ -32,6 +35,7 @@ class BaseWorker(Worker):
     """
 
     base_path = "/worker"  # By default, prefix all our routes with /worker
+    _cache_dir = "/tmp/arcade_worker_cache"  # noqa: S108
```
sdreyer, line 38: Should make this a random tmp path so if we have multiple servers running (like we can in deploy) then they don't overwrite each other

sdreyer: actually maybe the hash solves that

libs/arcade-serve/arcade_serve/core/base.py
```diff
@@ -82,11 +88,85 @@ def _set_secret(self, secret: str | None, disable_auth: bool) -> str:
...
-        return [tool.definition for tool in self.catalog]
+        if (
+            self._cache_enabled
+            and self._catalog_cache_path
+            and os.path.exists(self._catalog_cache_path)
+        ):
+            try:
+                with open(self._catalog_cache_path, encoding="utf-8") as f:
+                    catalog_data = json.load(f)
+                return [ToolDefinition.model_validate(td) for td in catalog_data]
```
sdreyer, line 162: Is this actually more efficient? Seems like having to load and decode then reencode would be slower than writing from memory

libs/arcade-serve/arcade_serve/core/base.py
EricGustin: yeah that was the intention of the hash in the file name. is hash good enough for you?

libs/arcade-serve/arcade_serve/core/base.py
evantahler: +1 - the goal here is to skip JSON parsing - can we stream the result?

PR #685, opened 2025-11-10
Fixing 403 error always being RateLimiting

jottakka:
Reponse 403 was returning RateLimiting all the time, but it was due only checking if rate limiting header exists, but it should be checked if it is 0 also.

### DIFF
--- a/libs/arcade-tdk/arcade_tdk/providers/http/error_adapter.py
+++ b/libs/arcade-tdk/arcade_tdk/providers/http/error_adapter.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from datetime import datetime, timezone
 from typing import Any
 from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


libs/arcade-tdk/arcade_tdk/providers/http/error_adapter.py
```diff
@@ -109,8 +110,10 @@ def _is_rate_limit_403(self, headers: dict[str, str], msg: str) -> bool:
         Returns:
             True if this 403 should be treated as rate limiting
         """
-        # Check if any rate limiting headers are present
-        return any(header.lower() in [h.lower() for h in headers] for header in RATE_HEADERS)
+        # Check if rate limit is actually exhausted (remaining requests is 0)
+        headers_lower = {k.lower(): v for k, v in headers.items()}
+        remaining = headers_lower.get("x-ratelimit-remaining")
```
EricGustin: `x-ratelimit-remaining` is probably not universal (?). Quick search for equivalent headers could help us make this more robust `ratelimit-remaining`, `x-rate-limit-remaining`, `x-app-rate-limit-remaining`, etc.

libs/arcade-tdk/arcade_tdk/providers/http/error_adapter.py
```diff
@@ -109,8 +110,10 @@ def _is_rate_limit_403(self, headers: dict[str, str], msg: str) -> bool:
...
+        return remaining is not None and remaining.isdigit() and int(remaining) == 0
```
EricGustin: If the value isn't a digit then are there other common formats that we could handle here?

EricGustin (approved):

PR #688, opened 2025-11-18
Worker Stability

EricGustin:
This PR does three things:
1. Executes synchronous tool calls in thread pool allowing for up to 4 + # of CPUs executions in parallel.
2. Makes force quitting via double SIGINT/SIGTERM possible and via single SIGINT/SIGTERM + graceful shutdown timeout expiry possible, even if there are active connections.
3. Sets `timeout_graceful_shutdown` to `ARCADE_UVICORN_TIMEOUT_GRACEFUL_SHUTDOWN` env var if set, else defaults to 15.
4. Disable the worker health check span to reduce noise

Tradeoffs:
Since this PR introduces executing synchronous tools via `await asyncio.to_thread(func, **func_args)`, this means that there is no way for the thread to be killed until it finishes. The ramifications of this is that the force quitting logic that is also implemented in this PR has to be very harsh `os._exit(1)` just in case there is a sync tool actively executing. This means that `MCPApp` teardown logic will not execute when force quitting is required. Although this was already the case because we weren't previously able to force quit! This tradeoff is justified for now since "parallel" tool executions will relieve us of many worker timeouts that we are seeing in prod.

Future work:
Minimize/eliminate the need for `os._exit(1)` such that `MCPApp` teardown logic will always execute, even when force quitting. The solution will likely be moving away from `await asyncio.to_thread(func, **func_args)` (while maintaining "parallelism" and then utilize the `TaskTrackerMiddleware` introduced in this PR to cancel all of the active HTTP requests.

Resolves PLT-713

### DIFF
--- a/libs/arcade-serve/arcade_serve/core/components.py
+++ b/libs/arcade-serve/arcade_serve/core/components.py
@@ -96,6 +96,4 @@ class HealthCheckComponent(WorkerComponent):
         """
         Handle the request to check the health of the worker.
         """
-        tracer = trace.get_tracer(__name__)
-        with tracer.start_as_current_span("HealthCheck"):
-            return self.worker.health_check()
+        return self.worker.health_check()


libs/arcade-core/arcade_core/executor.py
```diff
@@ -59,7 +59,7 @@ async def run(
             if asyncio.iscoroutinefunction(func):
                 results = await func(**func_args)
             else:
-                results = func(**func_args)
+                results = await asyncio.to_thread(func, **func_args)
```
EricGustin, line 62: (16 month old code) Sync tools were running directly on the uvicorn event loop preventing other tool calls from being processed.

I could see `asyncio.to_thread` causing some issues with the GIL if there ends up being multiple synchronous tools being executed this way simultaneously. But I think it's still better than completely blocking all other tool calls? @evantahler @sdreyer

sdreyer: Something else you could try, is set uvicorn works to more than 1. I believe that makes them not share a single GIL

EricGustin: I don't _think_ `uvicorn.run(..., workers=4)` will work for MCP sessions (because MCP sessions would then be process specific) and then `MCPApp`'s runtime capabilities would break since they would only affect one of the N processes so we would end up with uvicorn workers that have different tool catalogues, prompts, and resources

evantahler: That makes sense - MCP isn't stateless. 
Can the parent process manage the sessions for the children workers?

EricGustin: Definitely possible, but also adds **a lot** of complexity. Depends on what our goals are for this ticket

evantahler: Good point!  This is still better than what we have now, so we can merge this now and see if it is _enough_ of an improvement.

I think that this will:
* allow for multiple HTTP calls to happen in parallel or other things that "await" an external resource 
* (probably?) move some tool-exec CPU-bound work to the sub-threads as well (e.g. json ser/deser) 

I think that this will not:
* help the load from "front door" work (dealing with incoming MCP connections) or dealing with CPU-bound work (e.g. json ser/deser) when responding to the MCP client

Of note, I think we will want some control over how many threads exist in the threadpool for `asyncio.to_thread` to use.  I think the default is the # of CPUs on the host (or so Gemini tells me), so on your machine that's 8, and on prod, that's 1?  Since most tools are not CPU intensive, we can probably get away with more parallelism because most of the time will be spent waiting for a remote service

```py
num_threads = os.environ['TOOL_EXEC_THREAD_COUNT'] or os.cpu_count() * 2 
executor = concurrent.futures.ThreadPoolExecutor(max_workers=4)
loop = asyncio.get_running_loop()
loop.set_default_executor(executor)
```

EricGustin: Your understanding of what this will and will not help us with aligns with mine!

EricGustin: So I looked into what the default number of 'workers' for Python's `ThreadPoolExecutor`:

- Python >=3.8,<3.13: `min(32, os.cpu_count() + 4)` or in other words, 5 in production.
- Python >=3.13: `min(32, (os.process_cpu_count() or 1) + 4`, or in other words, 5 in production.

[Source](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)

pyproject.toml
```diff
@@ -136,6 +136,8 @@ omit = [
     "*/test_*",
     "*/__pycache__/*",
 ]
+parallel = true
+patch = ["subprocess"]
```
EricGustin, line 140: This is for including `test_end_to_end.py` in the coverage report.

libs/arcade-core/arcade_core/executor.py
EricGustin: See PR description for more details on why we should eventually move away from `await asyncio.to_thread(func, **func_args)`, and why it is okay to use for now.

libs/arcade-mcp-server/arcade_mcp_server/mcp_app.py
```diff
@@ -410,13 +409,14 @@ def _create_and_run_server(self, host: str, port: int) -> None:
...
         )
-        uvicorn.run(
-            app,
-            host=host,
-            port=port,
-            log_level=log_level,
-            reload=False,  # MCPApp handles its own reload via parent/child process pattern
-            lifespan="on",
+
+        asyncio.run(
```
EricGustin, line 413: moved away from `uvicorn.run` in favor of creating our own `uvicorn.Server` so that we can customize shutdown logic.

libs/arcade-serve/arcade_serve/fastapi/task_tracker.py
```diff
@@ -0,0 +1,57 @@
...
+import threading
+from collections.abc import Awaitable, Callable
+
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.types import ASGIApp
+
+
+class TaskTrackerMiddleware(BaseHTTPMiddleware):
```
EricGustin, line 11: although not absolutely required to be in this PR, it will set us up nicely for when we implement the 'future work' section of this PR description

PR #691, opened 2025-11-24
Only serve worker endpoints if secret is set

EricGustin:
Default to `ARCADE_WORKER_SECRET` being unset. This env var must be explicitly set now. Once it is set, the `worker/` endpoints will be served.

Closes https://github.com/ArcadeAI/arcade-mcp/security/advisories/GHSA-g2jx-37x6-6438

### DIFF
--- a/libs/arcade-mcp-server/pyproject.toml
+++ b/libs/arcade-mcp-server/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "arcade-mcp-server"
-version = "1.8.0"
+version = "1.9.0"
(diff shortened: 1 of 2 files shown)


libs/arcade-mcp-server/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "arcade-mcp-server"
-version = "1.8.0"
+version = "1.9.0"
```
EricGustin, line 7: Given that the `worker/` routes are only used internally, I am going to consider this a minor version bump

evantahler (approved):

PR #692, opened 2025-11-24
[TOO-145]GQL Error Adaptor

jottakka:
# 🎫  [TOO-145](https://linear.app/arcadedev/issue/TOO-145/adding-gql-error-interceptor-in-tdk)
- add GraphQL-specific error adapter, expose it, and slot it ahead of HTTP fallback  
- map gql query/transport errors (incl. rate-limit handoff) without logging sensitive data  
- cover adapter + tool chain behavior with new unit tests (pathload, dedupe, edge cases)

closes TOO-145

### DIFF
--- a/libs/arcade-tdk/pyproject.toml
+++ b/libs/arcade-tdk/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "arcade-tdk"
-version = "3.0.1"
+version = "3.1.0"
 description = "Arcade TDK - Toolkit Development Kit for building Arcade tools"
 readme = "README.md"
 license = {text = "MIT"}


libs/arcade-tdk/arcade_tdk/providers/graphql/error_adapter.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Handle GraphQL Query Errors (HTTP 200 OK but contains 'errors' list)
+        if isinstance(exc, TransportQueryError):
+            return self._handle_query_error(exc)
+
+        # Handle Transport Server Errors (HTTP 500s, 400s from the transport layer)
+        if isinstance(exc, TransportServerError):
+            return self._handle_server_error(exc)
+
+        return None
```
EricGustin, line 97: Consider adding a catch all for the base `gql.transport.exceptions.TransportError` before returning `None`

libs/arcade-tdk/arcade_tdk/providers/graphql/error_adapter.py
```diff
@@ -0,0 +1,164 @@
...
+
+        developer_message = "GraphQL error"
+        if error_codes:
+            developer_message = f"GraphQL error codes: {', '.join(sorted(set(error_codes)))}"
+
+        return UpstreamError(
+            message=f"Upstream GraphQL error: {joined_message}",
+            status_code=status_code,
+            developer_message=developer_message,
+            extra={
```
EricGustin, line 128: Please add the request URL as a minimum to the `extra` if possible.

EricGustin (commented):
Mostly looks good to me. Package also needs a version bump

libs/arcade-tdk/arcade_tdk/providers/graphql/error_adapter.py
jottakka: In graphql it always hit the same endpoint for all APIs f `{base_url}/graphql `, so it isn't so directly availabe in the error, so it needed some workaround was needed as it can use different transport libs (aiohttp, httpx, and requests)

libs/arcade-tdk/arcade_tdk/providers/graphql/error_adapter.py
jottakka: Ok!

libs/arcade-tdk/arcade_tdk/providers/graphql/error_adapter.py
EricGustin: understood

EricGustin (approved):

PR #697, opened 2025-11-25
Exclude health span from telemetry

sdreyer:


EricGustin:
Make sure to bump version in pyproject.toml

PR #699, opened 2025-11-26
Bumps critical JS dependency versions

evantahler:
Fixes a number of security issues with JS dependencies in our examples. Does not upgrade to the latest versions of frameworks.

### DIFF
--- a/contrib/examples/mastra/src/mastra/tools/flightSearchTools.ts
+++ b/contrib/examples/mastra/src/mastra/tools/flightSearchTools.ts
@@ -25,9 +25,20 @@ const flightToolkit = await arcade.tools.list({ toolkit: "GoogleFlights", limit:
  *
  * Learn more: https://docs.arcade.dev/home/use-tools/get-tool-definitions#get-zod-tool-definitions
  */
-export const flightTools = toZodToolSet({
+const baseFlightTools = toZodToolSet({
     tools: flightToolkit.items,
     client: arcade,
     userId: "<YOUR_USER_ID>", // Your app's internal ID for the user (an email, UUID, etc). It's used internally to identify your user in Arcade
     executeFactory: executeOrAuthorizeZodTool, // Checks if tool is authorized and executes it, or returns authorization URL if needed
 })
+
+// Add id property to each tool that copies the name
+export const flightTools = Object.fromEntries(
+    Object.entries(baseFlightTools).map(([key, tool]) => [
+        key,
+        {
+            ...tool,
+            id: tool.name,
+        },
+    ])
+)
--- a/contrib/examples/mastra/src/mastra/tools/hotelSearchTools.ts
+++ b/contrib/examples/mastra/src/mastra/tools/hotelSearchTools.ts
@@ -25,9 +25,20 @@ const hotelToolkit = await arcade.tools.list({ toolkit: "GoogleHotels", limit: 3
  *
  * Learn more: https://docs.arcade.dev/home/use-tools/get-tool-definitions#get-zod-tool-definitions
  */
-export const hotelTools = toZodToolSet({
+const baseHotelTools = toZodToolSet({
     tools: hotelToolkit.items,
     client: arcade,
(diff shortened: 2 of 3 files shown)


contrib/examples/langchain-ts/package-lock.json
evantahler, line 1: we both an npm and pnpm lockfile.  that was silly

contrib/examples/mastra/src/mastra/tools/flightSearchTools.ts
```diff
@@ -25,9 +25,20 @@ const flightToolkit = await arcade.tools.list({ toolkit: "GoogleFlights", limit:
...
+// Add id property to each tool that copies the name
+export const flightTools = Object.fromEntries(
+    Object.entries(baseFlightTools).map(([key, tool]) => [
+        key,
+        {
+            ...tool,
+            id: tool.name,
+        },
+    ])
+)
```
evantahler, line 44: Mastra tools were previously assigned random IDs - now we have names that make sense!
Screenshot 2025-11-26 at 2 31 58 PM

sdreyer (approved):
lgtm but Ill let @sdserranog have final say

contrib/examples/langchain-ts/package.json
```diff
@@ -13,15 +13,15 @@
...
   "license": "ISC",
-  "packageManager": "pnpm@10.6.5",
+  "packageManager": "pnpm@10.23.0",
   "dependencies": {
     "@arcadeai/arcadejs": "latest",
-    "@langchain/community": "^0.3.42",
-    "@langchain/core": "^0.3.55",
-    "@langchain/langgraph": "^0.2.72",
-    "@langchain/openai": "^0.5.10"
+    "@langchain/community": "^1.0.5",
```
sdserranog, line 19: These are breaking the langchain example for me.

contrib/examples/mastra/src/mastra/tools/hotelSearchTools.ts
```diff
@@ -25,9 +25,20 @@ const hotelToolkit = await arcade.tools.list({ toolkit: "GoogleHotels", limit: 3
...
-export const hotelTools = toZodToolSet({
+const baseHotelTools = toZodToolSet({
     tools: hotelToolkit.items,
     client: arcade,
     userId: "<YOUR_USER_ID>", // Your app's internal ID for the user (an email, UUID, etc). It's used internally to identify your user in Arcade
     executeFactory: executeOrAuthorizeZodTool, // Checks if tool is authorized and executes it, or returns authorization URL if needed
 })
+
+// Add id property to each tool that copies the name
+export const hotelTools = Object.fromEntries(
```
sdserranog, line 36: We should probably fix this in the createZodTool utility function, but I think it's fine for now until we update these examples with the latest versions and practices.

contrib/examples/langchain-ts/package.json
evantahler: weird. It works for me... can you do the thing where you `rm -rf node_modules` and try again?  There are so many transitive deps :/

evantahler:
Yeah, I'm really just trying to close out the security issues with the deps right now... a lot of these examples need to be updated

contrib/examples/langchain-ts/package.json
sdserranog: It was a clean install, but testing again

evantahler: Any luck?

sdserranog: Screenshot 2025-12-01 at 10 12 25 PM
Getting that message when I run `pnpm studio` and the dev-auth command is also failing because in this line:
`const githubToolkit = await arcade.tools.list({ toolkit: "github", limit: 30 })
We need to increase the limit from 30 to at least 50, or simply remove it. These issues are not really related to this PR, so I will approve, but we really need to address these examples.

PR #708, opened 2025-12-03
fix: Lock-less TaskTrackerMiddleware

nbarbettini:
Opening this PR to start a discussion on whether a **lock-free** mechanism makes sense here. I'm investigating this hot path as a potential GIL contention area.


EricGustin:
@nbarbettini Did we determine whether this hot path was a GIL contention area? i.e., can we close this PR or does it need to be revived

PR #717, opened 2025-12-05
feat: Support multiple orgs & projects in Arcade CLI

nbarbettini:
Fixes [PLT-720: Refactor CLI to support multiple orgs + projects](https://linear.app/arcadedev/issue/PLT-720/refactor-cli-to-support-multiple-orgs-projects)

This PR removes the legacy login flow (login to get an API key) from Arcade CLI. Believe it or not, this flow predates the ability to get an API key from the Dashboard, or even the Dashboard itself!

Notable changes:

**Legacy handling** - When a user with an existing `credentials.yaml` updates the CLI, they will get instructions on fixing their old credentials:
Screenshot 2025-12-08 at 10 10 37

Any commands that require login (non-public commands) will be blocked with the above message until `arcade logout / arcade login` is performed again.

**New login flow**

```sh
arcade login
Opening a browser to log you in...

✅ Logged in as nate@arcade.dev.

Active project: Nate Barbettini's organization / Default project
Run 'arcade org list' or 'arcade project list' to see available options.
```

**List and set the active organization**
```sh
arcade org list
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━┓
┃ Name                           ┃ ID                                   ┃ Default ┃ Active ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━┩
│ Nate Barbettini's organization │ 1c64968e-fdc5-4c55-8612-2ce46cd7881b │ ✓       │ ✓      │
│ Sergio 743                     │ 1f1f6184-58dc-4bac-bdde-b9184e43fdf3 │         │        │
└────────────────────────────────┴──────────────────────────────────────┴─────────┴────────┘

Use 'arcade org set <org_id>' to switch organizations.
```
```sh
arcade org set 1c64968e-fdc5-4c55-8612-2ce46cd7881b 

✓ Switched to organization: Nate Barbettini's organization
  Active project: Default project
```

**List and set the active project**
```sh
arcade project list

Active organization: Nate Barbettini's organization
Use 'arcade org list' and 'arcade org set <org_id>' to switch organizations.

┏━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━┓
┃ Name            ┃ ID                                   ┃ Default ┃ Active ┃
┡━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━┩
│ Default project │ 35166bf3-6e68-481e-bf16-f747fadc6c22 │ ✓       │ ✓      │
│ Second project  │ 62963205-31ea-4fda-9fc4-af10db89c06f │         │        │
└─────────────────┴──────────────────────────────────────┴─────────┴────────┘

Use 'arcade project set <project_id>' to switch projects.
```
```sh
arcade project set 35166bf3-6e68-481e-bf16-f747fadc6c22
✓ Switched to project: Default project
```

---

> [!NOTE]
> Migrates CLI to OAuth2 (PKCE) with saved org/project context, adds org/project commands, rewrites Engine calls to org-scoped endpoints, and bumps core packages.
> 
> - **Auth & Config**
>   - Implement OAuth2 Authorization Code + PKCE (`arcade_cli/authn.py`) with local callback server and Jinja templates.
>   - Persist tokens and active `context` (org/project) in `credentials.yaml` via updated config models (`arcade_core/config_model.py`).
>   - Add token refresh and CLI config fetch utilities (`arcade_core/auth_tokens.py`).
>   - Detect legacy API-key credentials and block protected commands until re-login; add `whoami` command.
> - **Org/Project Management**
>   - New subcommands: `arcade org list|set`, `arcade project list|set` (fetch via Coordinator).
> - **Engine API usage (org-scoped)**
>   - Introduce org/project URL rewriting transports (`arcade_core/network/org_transport.py`) and helpers (`get_org_scoped_url`, `get_arcade_client`, `get_auth_headers`).
>   - Update `deploy`, `server`, and `secret` commands to use Bearer tokens and org-scoped paths; adjust log streaming/status, secrets CRUD, and deployment workflows.
> - **CLI UX**
>   - Replace legacy login URLs/constants; add success/failure HTML templates for browser callback.
>   - Tweak `dashboard` to health-check without credentials.
>   - Usage tracking now includes `org_id`/`project_id` properties.
> - **Tests**
>   - Update tests for dashboard, secrets, utils, and usage identity (OAuth `/whoami`).
> - **Dependencies & Versions**
>   - Bump packages: `arcade-core@4.0.0`, `arcade-mcp-server@1.12.0`, `arcade-serve@3.2.0`, `arcade-tdk@3.3.0`; add `authlib`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 49702c2f74b9db15bb286d3ec71179b4e74a9134. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


libs/arcade-cli/arcade_cli/authn.py
```diff
@@ -1,35 +1,747 @@
...
+
+# Set up Jinja2 templates
+_TEMPLATES_DIR = Path(__file__).parent / "templates"
+_jinja_env = Environment(loader=FileSystemLoader(_TEMPLATES_DIR), autoescape=True)
+
+
+def _render_template(template_name: str, **context: Any) -> bytes:
+    """Render a Jinja2 template and return as bytes."""
+    template = _jinja_env.get_template(template_name)
+    return template.render(**context).encode("utf-8")
```
nbarbettini, line 45: We previously returned an old and hardcoded HTML page on login success/failure. I brought our newer and nicer jinja templates over from Coordinator.

libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -89,7 +94,7 @@ class UpdateDeploymentRequest(BaseModel):
 # Deployment Status Functions
 
 
-def get_deployment_status(engine_url: str, api_key: str, server_name: str) -> str:
+def _get_deployment_status(engine_url: str, server_name: str) -> str:
```
nbarbettini, line 97: Across this PR, I modified functions to stop passing around `api_key` as a param, and instead make it a concern of either `get_auth_headers` or a concern of getting a client. That way, these action functions only care about calling an endpoint and just assume that the underlying client or HTTP request is configured correctly.

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "typer==0.10.0",
     "rich==13.9.4",
     "Jinja2==3.1.6",
+    "authlib>=1.3.0",
```
EricGustin: Any way we can make this dependency stricter? The CLI is intended to be installed as a standalone instead of being added as a dependency of someone's project (e.g., uv tool install arcade-mcp), so we should be able to use strict dependency version here

libs/arcade-cli/arcade_cli/authn.py
```diff
@@ -1,35 +1,747 @@
...
+
+    Picks the org/project marked as default, or falls back to the first one
+    in the list if none are marked as default.
+
+    Args:
+        tokens: OAuth tokens
+        whoami: Response from /whoami endpoint with user and orgs/projects
+    """
+    # Ensure config directory exists
+    if not os.path.exists(ARCADE_CONFIG_PATH):
```
EricGustin: nit; you can remove if-statement & still get same behavior since `exist_ok=True` is used here

libs/arcade-cli/arcade_cli/utils.py
```diff
@@ -422,25 +429,227 @@ def validate_and_get_config(
...
+    if config.is_legacy_format():
+        # Legacy API key auth
+        if not config.api or not config.api.key:
+            handle_cli_error("API key not found. Please run `arcade login`.")
+            raise AssertionError("unreachable")  # handle_cli_error raises CLIError
+        console.print(
+            "⚠️  Your credentials use an older format. "
+            "Run 'arcade logout' then 'arcade login' to update.",
+            style="bold yellow",
+        )
```
EricGustin: I'm concerned about the way in which legacy configuration warnings are sprinkled throughout, because it's leading to various unhandled errors later in the execution of CLI commands. 

During my testing I ran into a few scenarios where I would be warned about a legacy configuration, and then the CLI commend would later run into an unhandled error.

I think we have two options here:
1. Handle every legacy configuration case & every up-to-date configuration case and audit all of them to ensure actual graceful degradation.
2. Perform an upront legacy configuration check & force the CLI user to go through the logout/re-login flow if they are using a legacy configuration.

I'm advocating for option 2 because (1) it's faster to implement, (2) it doesn't risk missing an edge-case & appearing fragile, and (3) we can get rid of all of the old code for the legacy config today

--------------

Here is an example of an unhandled error that can be avoided if we fail fast for legacy configs:

image

libs/arcade-cli/arcade_cli/org.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+@app.command("list", help="List organizations you belong to")
+def org_list(
+    host: str = typer.Option(
+        PROD_COORDINATOR_HOST,
+        "--host",
+        "-h",
+        help="The Arcade Coordinator host.",
+    ),
```
EricGustin, line 46: For commands that have subcommands like `arcade org`, the convention (at least in this repo) is to support common configuration options at the top-level command instead of at the sub-level command. `arcade secret` is an example where the top-level command handles configuration options like --host.

Example:

- `arcade secret --host api.bosslevel.dev list` instead of `arcade secret list --host api.bosslevel.dev`
- `arcade org --host cloud.bosslevel.dev list` instead of `arcade org list --host cloud.bosslevel.dev`

libs/arcade-cli/arcade_cli/usage/command_tracker.py
EricGustin, line 1: **Degradation**: We lost the ability to 'alias' an anonymous user to a known user upon `arcade login`, so now users remain anonymous even after logging in.

Previously we would use the `principal_id` received from the coordinator and store that in `~/.arcade/usage.json`. We'll probably need some updates to `_handle_successful_login()`

libs/arcade-cli/arcade_cli/project.py
```diff
@@ -0,0 +1,162 @@
...
+        table.add_column("Default", style="green")
+        table.add_column("Active", style="bold yellow")
+
+        for project in projects:
+            is_active = "✓" if project.project_id == active_project_id else ""
+            is_default = "✓" if project.is_default else ""
+            table.add_row(project.name, project.project_id, is_default, is_active)
+
+        console.print(table)
+        console.print("\nUse 'arcade project set <project_id>' to switch projects.\n")
```
EricGustin, line 109: loving these next step hints

EricGustin (commented):
Excited for this. Some comments below

libs/arcade-cli/arcade_cli/org.py
nbarbettini: Ah, I see. I assumed that was unintentional and "fixed" it in this PR.
I can revert my change.

nbarbettini: Can be out of scope for this PR: Do we like that pattern? I find it very unintuitive myself 😛

EricGustin: I see. I'm also fine if we change this convention! (as long as there's consistency across all of the commands that have subcommands)

libs/arcade-cli/arcade_cli/utils.py
nbarbettini: Good reasoning - agreed.
Updated to prevent any commands (that require login) from running if legacy credentials are detected. I like the fact that we can delete more code this way too 🙂

libs/arcade-cli/arcade_cli/org.py
nbarbettini: 👍 For this PR, I updated the new `org` and `project` commands to follow the existing pattern. The PR is already big enough!

libs/arcade-cli/arcade_cli/utils.py
nbarbettini: Updated the PR description with a screenshot of what this looks like now.

libs/arcade-cli/arcade_cli/usage/command_tracker.py
nbarbettini: This should work now, but I'd appreciate a double-check since you are more familiar with it than I! 🙏

libs/arcade-core/arcade_core/usage/identity.py
EricGustin, line 1: since we're making changes to arcade-core, we need to 
1. Bump arcade-core's version,
2. Increase the minimum arcade-core version in the CLI's dependencies

libs/arcade-core/arcade_core/usage/identity.py
```diff
@@ -124,34 +124,63 @@ def get_distinct_id(self) -> str:
...
+                config = yaml.safe_load(f) or {}
+
+            cloud_config = config.get("cloud", {}) if isinstance(config, dict) else {}
+
+            # OAuth credentials: use access_token to call /whoami
+            auth_config = cloud_config.get("auth", {}) if isinstance(cloud_config, dict) else {}
+            access_token = auth_config.get("access_token")
+            if access_token:
+                response = httpx.get(
+                    "https://api.arcade.dev/api/v1/auth/whoami",
```
EricGustin: I think this needs to be `https://cloud...`, not `https://api...` otherwise we'll get a 404 causing users to remain anonymous. 

Also, should we pass the coordinator_url to this function so that we can switch between prod & bosslevel?

libs/arcade-core/arcade_core/usage/identity.py
nbarbettini: Good catch, thanks!

libs/arcade-cli/arcade_cli/utils.py
```diff
@@ -422,25 +396,180 @@ def validate_and_get_config(
...
+    Example:
+        client = get_arcade_client("https://api.arcade.dev")
+        servers = client.workers.list()  # Automatically uses org-scoped URLs
+    """
+    config = validate_and_get_config()
+
+    # OAuth mode: need to rewrite URLs to include org/project scope
+    from arcade_cli.authn import get_valid_access_token
+
+    access_token = get_valid_access_token(base_url)
```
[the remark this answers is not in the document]
nbarbettini: Good bot

libs/arcade-core/arcade_core/usage/identity.py
nbarbettini: Excellent catch 👀 

In one of my other fixes (a bugbot report above) I started storing `coordinator_url` in the `credentials.yaml` file, so we don't have to guess this anymore. 👍

EricGustin: Nice! I like that idea

libs/arcade-mcp-server/arcade_mcp_server/auth_tokens.py
```diff
@@ -0,0 +1,109 @@
+"""
+Shared access-token utilities used by both the CLI and MCP server.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timedelta
+
+import httpx
+from arcade_cli.constants import PROD_COORDINATOR_HOST
```
EricGustin: `arcade_cli` isn't a dependency of `arcade_mcp_server` (instead, `arcade_mcp_server` is a dependency of `arcade_cli`) so I think this will cause either module not found, or circular import. Since this file is shared maybe we could place it in arcade-core? Or maybe be less DRY and duplicate code? This comment somewhat applies to org_transport.py as well.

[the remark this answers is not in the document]
EricGustin: you beat me to it Mr. Bot

PR #742, opened 2026-01-16
Support Ed25519 Algorithm

EricGustin:
Ed25519 is needed for Arcade AS. This required migrating from `python-jose` to `joserfc`, because `python-jose` didn't seem to support Ed25519

### DIFF
--- a/libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
+++ b/libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
@@ -5,12 +5,11 @@ Implements OAuth 2.1 Resource Server token validation using JWT with JWKS.
 """
 
 import binascii
-import json
 import time
 from typing import Any, cast
 
 import httpx
-from joserfc import jws
+from joserfc import jws, jwt
 from joserfc.errors import JoseError
 from joserfc.jwk import KeySet, KeySetSerialization
 from joserfc.jws import JWSRegistry
@@ -187,7 +186,7 @@ class JWKSTokenValidator(ResourceServerValidator):
         else:
             return self._jwks_cache
 
-    def _get_unverified_header(self, token: str) -> dict[str, Any]:
+    def _get_headers_without_verification(self, token: str) -> dict[str, Any]:
         """Extract header from JWT without verification.
 
         Args:
@@ -202,7 +201,7 @@ class JWKSTokenValidator(ResourceServerValidator):
         try:
             # Use joserfc's extract_compact to parse JWT without verification
             obj = jws.extract_compact(token.encode())
-            return dict(obj.protected)
+            return obj.headers()
         except (JoseError, ValueError, binascii.Error) as e:
             raise InvalidTokenError(f"Invalid JWT format: {e}") from e
 
@@ -219,7 +218,7 @@ class JWKSTokenValidator(ResourceServerValidator):
         Raises:
             InvalidTokenError: If no matching key found or algorithm mismatch
         """
-        header = self._get_unverified_header(token)
+        header = self._get_headers_without_verification(token)
         kid = header.get("kid")
         token_alg = header.get("alg")
 
@@ -251,9 +250,8 @@ class JWKSTokenValidator(ResourceServerValidator):
     def _decode_token(self, token: str, signing_key: Any) -> dict[str, Any]:
         """Decode and verify the provided JWT token.
 
-        Uses jws.deserialize_compact for signature verification only, then
-        performs all claims validation manually. This gives us full control
-        over leeway handling for time-based claims (exp, iat, nbf).
+        Uses jwt.decode for signature verification and payload parsing, then
+        performs time-based claims validation manually for leeway handling
 
         Args:
             token: JWT token
@@ -271,19 +269,15 @@ class JWKSTokenValidator(ResourceServerValidator):
         else:
             algorithms = [self.algorithm]
 
-        # First, verify signature without any claims validation,
+        # First, verify signature & decode payload
         try:
-            result = jws.deserialize_compact(
-                token.encode(), signing_key, algorithms=algorithms, registry=_ACCESS_TOKEN_REGISTRY
+            result = jwt.decode(
+                token, signing_key, algorithms=algorithms, registry=_ACCESS_TOKEN_REGISTRY
             )
         except JoseError as e:
             raise InvalidTokenError(f"Token signature verification failed: {e}") from e
 
-        # Parse the payload as JSON claims
-        try:
-            decoded = cast(dict[str, Any], json.loads(result.payload))
-        except (json.JSONDecodeError, ValueError) as e:
-            raise InvalidTokenError(f"Invalid token payload: {e}") from e
+        decoded = result.claims
 
         # Validate time based claims with leeway support
         current_time = int(time.time())
--- a/libs/arcade-mcp-server/pyproject.toml
+++ b/libs/arcade-mcp-server/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
     "anyio>=4.0.0",
     "python-dotenv>=1.0.0",
     "pydantic-settings>=2.10.1",
-    "joserfc>=1.0.0",
+    "joserfc>=1.5.0",
     "httpx>=0.27.0,<1.0.0",
 ]
 
--- a/libs/tests/arcade_mcp_server/test_resource_server_auth.py
+++ b/libs/tests/arcade_mcp_server/test_resource_server_auth.py
@@ -520,13 +520,10 @@ class TestJWKSTokenValidator:
             mock_response.raise_for_status = Mock()
             mock_get.return_value = mock_response
 
-            # Patch jws.deserialize_compact to count calls
-            from joserfc import jws
-
             with patch(
-                "arcade_mcp_server.resource_server.validators.jwks.jws.deserialize_compact",
-                wraps=jws.deserialize_compact,
-            ) as mock_deserialize:
+                "arcade_mcp_server.resource_server.validators.jwks.jwt.decode",
+                wraps=jwt.decode,
+            ) as mock_decode:
                 validator = JWKSTokenValidator(
                     jwks_uri="https://auth.example.com/.well-known/jwks.json",
                     issuer=[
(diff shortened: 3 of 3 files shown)


libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
```diff
@@ -139,6 +195,25 @@ async def _fetch_jwks(self) -> dict[str, Any]:
         else:
             return self._jwks_cache
 
+    def _get_unverified_header(self, token: str) -> dict[str, Any]:
```
EricGustin: python-jose had this built-in

libs/arcade-mcp-server/pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
     "anyio>=4.0.0",
     "python-dotenv>=1.0.0",
     "pydantic-settings>=2.10.1",
-    "python-jose[cryptography]>=3.3.0,<4.0.0",
+    "joserfc>=1.0.0",
```
wdawson: should this be >=1.5.0 since that's where Ed25519 was added?

libs/tests/arcade_mcp_server/test_resource_server_auth.py
```diff
@@ -448,34 +500,33 @@ async def test_validate_single_audience_with_list_token_aud(
...
 
         with patch("httpx.AsyncClient.get") as mock_get:
             mock_response = Mock()
             mock_response.json.return_value = jwks_data
             mock_response.raise_for_status = Mock()
             mock_get.return_value = mock_response
 
-            # Patch jwt.decode to count calls
+            # Patch jws.deserialize_compact to count calls
+            from joserfc import jws
```
wdawson: not sure if a norm in this repo or not, but do we want these lazy imports or should this be up at the top?

libs/tests/arcade_mcp_server/test_resource_server_auth.py
```diff
@@ -600,6 +639,209 @@ async def test_validate_with_leeway(self, serialized_private_key, jwks_data):
...
+            jwks_uri="https://example.com/jwks",
+            issuer="https://example.com",
+            audience="https://example.com",
+            algorithm="EdDSA",
+        )
+        assert validator.algorithm == "EdDSA"
+
+
+class TestIntermediateASConfiguration:
+    """Tests for Arcade Intermediate AS configuration."""
```
wdawson: "Intermediate AS" is our internal name, but should probably just be "Arcade AS" here.

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
```diff
@@ -92,6 +115,14 @@ def __init__(
                 audience="https://mcp.example.com/mcp",
                 algorithm="ES256",
             )
+
+            # Ed25519 algorithm (for Arcade Intermediate AS)
```
wdawson: I would call this "Arcade AS" here. "Intermediate" is more of an internal thing.

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
```diff
@@ -114,6 +145,31 @@ def __init__(
         self._jwks_cache: dict[str, Any] | None = None
         self._cache_timestamp: float = 0
 
+    def _normalize_algorithm(self, alg: str) -> str:
```
wdawson, line 139: For my own understanding, it looks like these are providing some early circuit breaking if we get a token with mismatched `kid` and `alg` values based on the configuration of the server. Is that right?

I think we don't _need_ these checks. If we simply looked up the key in a dict by `kid`, and passed it to `jws.deserialize_compact()` I think joserfc would do all that checking for us. But maybe there's good benefit to doing it ourselves? Providing better errors, etc?

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
```diff
@@ -92,6 +115,14 @@ def __init__(
...
                 algorithm="ES256",
             )
+
+            # Ed25519 algorithm (for Arcade Intermediate AS)
+            validator = JWKSTokenValidator(
+                jwks_uri="https://cloud.arcade.dev/.well-known/jwks/oauth2",
+                issuer="https://cloud.arcade.dev/oauth2",
+                audience="urn:arcade:mcp",
+                algorithm="Ed25519",
+            )
```
nbarbettini: Not sure if we need an example of this in the public docs; we are the only ones who will use it.

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
```diff
@@ -139,6 +195,25 @@ async def _fetch_jwks(self) -> dict[str, Any]:
...
+        Returns:
+            Header dictionary containing 'alg', 'kid', etc.
+
+        Raises:
+            InvalidTokenError: If token format is invalid
+        """
+        try:
+            # Use joserfc's extract_compact to parse JWT without verification
+            obj = jws.extract_compact(token.encode())
+            return dict(obj.protected)
```
nbarbettini: Double-checking: does this return just the header, or the whole object?

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
EricGustin: Oh, I was under the assumption that others could as well. I'll remove for now though

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
```diff
@@ -147,74 +222,97 @@ def _find_signing_key(self, jwks: dict[str, Any], token: str) -> Any:
...
 
         raise InvalidTokenError("No matching key found in JWKS")
 
-    def _decode_token(self, token: str, signing_key: str) -> dict[str, Any]:
+    def _decode_token(self, token: str, signing_key: Any) -> dict[str, Any]:
         """Decode and verify the provided JWT token.
 
+        Uses jws.deserialize_compact for signature verification only, then
+        performs all claims validation manually. This gives us full control
+        over leeway handling for time-based claims (exp, iat, nbf).
```
nbarbettini: Is this necessary because joserfc doesn't validate the way we want? (usually I'd like to let the lib do this stuff)

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
wdawson: I think you can:


```suggestion
            return obj.headers()
```

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
wdawson: is the name of this function misleading? Looks like you're extracting _all_ of the headers


```suggestion
    def _get_unverified_headers(self, token: str) -> dict[str, Any]:
```

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
EricGustin: just header params. e.g., typ, iss, etc.

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
```diff
@@ -147,74 +214,97 @@ def _find_signing_key(self, jwks: dict[str, Any], token: str) -> Any:
...
+                token.encode(), signing_key, algorithms=algorithms, registry=_ACCESS_TOKEN_REGISTRY
+            )
+        except JoseError as e:
+            raise InvalidTokenError(f"Token signature verification failed: {e}") from e
+
+        # Parse the payload as JSON claims
+        try:
+            decoded = cast(dict[str, Any], json.loads(result.payload))
+        except (json.JSONDecodeError, ValueError) as e:
+            raise InvalidTokenError(f"Invalid token payload: {e}") from e
```
wdawson: I think `result.protected` gives you a `dict[str, Any]` and avoids the need to do this json thing again

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
wdawson: authlib will do this, but joserfc is just the jose bits. We could use authlib instead

libs/tests/arcade_mcp_server/test_resource_server_auth.py
nbarbettini: FWIW I prefer imports at top!

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
nbarbettini: @wdawson I understood it to be more like `_get_headers_without_verification`

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
EricGustin: As you mentioned, it does allow us to provide very clear error messages like `Token algorithm 'RS256' doesn't match configured algorithm 'Ed25519'`. I _think_ this justifies the extra code

libs/arcade-mcp-server/pyproject.toml
EricGustin: good catch!

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
EricGustin: updated

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
EricGustin: Yes, `_get_headers_without_verification` is much better than what i have now. Changing that now

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
EricGustin: switched to jwt.decode so the json thing is avoided now

libs/arcade-mcp-server/arcade_mcp_server/resource_server/validators/jwks.py
EricGustin: updated to obj.headers() to be more obvious

PR #754, opened 2026-01-27
Four bug fixes

EricGustin:
1. Resolves [TOO-363](https://linear.app/arcadedev/issue/TOO-363/arcade-deploy-fails-when-additional-deps-are-added-to-the-server).
2. Resolves [TOO-364](https://linear.app/arcadedev/issue/TOO-364/arcade-cores-tool-skip-logic-is-missing-case-for-direct-execution).
3. Resolves [TOO-358](https://linear.app/arcadedev/issue/TOO-358/missing-evals-error-message-shows-wrong-command).
4. Resolves [TOO-365](https://linear.app/arcadedev/issue/TOO-365/arcade-evals-unit-tests-are-hanging).

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how `arcade deploy` spawns the server process and adjusts toolkit discovery skip logic, which can affect deployments and tool discovery; however, the changes are small and covered by new unit/integration tests.
> 
> **Overview**
> `arcade deploy` now starts the validation server using the project’s `.venv` interpreter (via `find_python_interpreter`) instead of the CLI’s own `sys.executable`, preventing missing dependency failures when the CLI is installed in an isolated env.
> 
> `arcade-core`’s `Toolkit.tools_from_directory` skip logic is hardened to also skip the currently executing entrypoint by module name (`__main__.__spec__.name`) when file paths don’t match (e.g., bundled execution). CLI error printing now escapes plain messages to avoid rich markup issues, and `arcade-evals` lock acquisition accepts an optional timeout default.
> 
> Adds unit tests for the new toolkit skip behavior and an integration test that boots the MCP server via direct Python invocation to mirror deployment behavior, and bumps `arcade-core`, `arcade-mcp-server`, and root dependency versions accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e7785634c231c059f2e0bd1bc73a56bd7470a494. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libs/arcade-cli/arcade_cli/configure.py
+++ b/libs/arcade-cli/arcade_cli/configure.py
@@ -146,7 +146,16 @@ def get_tool_secrets() -> dict:
 
 
 def find_python_interpreter() -> Path:
-    # Find the Python interpreter in the virtual environment
+    """
+    Find the Python interpreter in the virtual environment.
+
+    NOTE: This function assumes it is called from the project root directory (where .venv lives).
+    Currently, callers like `arcade deploy` enforce this by requiring pyproject.toml in cwd.
+    If this requirement is relaxed in the future, this function should be updated to:
+      1. Accept a project_root parameter, OR
+      2. Honor VIRTUAL_ENV / UV_PROJECT_ENVIRONMENT env vars, OR
+      3. Search upward from cwd to find pyproject.toml and resolve .venv relative to that
+    """
     venv_python = None
     # Check for .venv first (uv default)
     if (Path.cwd() / ".venv").exists():


libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -388,7 +388,13 @@ def start_server_process(entrypoint: str, debug: bool = False) -> tuple[subproce
         "ARCADE_WORKER_SECRET": "temp-validation-secret",
     }
 
-    cmd = [sys.executable, entrypoint]
+    # Use the project's Python environment, not the CLI's isolated environment.
+    # find_python_interpreter() looks for .venv/bin/python in cwd, falling back to sys.executable.
+    # This ensures the server runs in the project's environment even when the CLI is installed
+    # in an isolated environment (e.g., via 'uv tool install arcade-mcp').
+    project_python = find_python_interpreter()
+    cmd = [str(project_python), entrypoint]
```
EricGustin, line 396: This is the fix for TOO-363

libs/arcade-cli/arcade_cli/utils.py
```diff
@@ -394,7 +394,7 @@ def handle_cli_error(
     elif error:
         console.print(f"❌ {message}: {escape(str(error))}", style="bold red")
     else:
-        console.print(f"❌ {message}", style="bold red")
+        console.print(f"❌ {escape(message)}", style="bold red")
```
EricGustin, line 397: This is the fix for TOO-358

libs/arcade-evals/arcade_evals/loaders.py
```diff
@@ -98,10 +98,10 @@ async def _get_cache_lock(cache_key: str) -> asyncio.Lock:
     return _cache_locks[cache_key]
 
 
-async def _acquire_lock_with_timeout(
-    lock: asyncio.Lock, timeout: float = LOCK_TIMEOUT_SECONDS
-) -> bool:
+async def _acquire_lock_with_timeout(lock: asyncio.Lock, timeout: float | None = None) -> bool:
     """Acquire a lock with timeout. Returns True if acquired, False on timeout."""
+    if timeout is None:
+        timeout = LOCK_TIMEOUT_SECONDS
```
EricGustin, line 104: this is the fix for TOO-365. Reduced unit test runtime from 155 seconds to 37 seconds.

libs/arcade-core/arcade_core/toolkit.py
EricGustin, line 1: fix for TOO-364

libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -388,7 +388,13 @@ def start_server_process(entrypoint: str, debug: bool = False) -> tuple[subproce
         "ARCADE_WORKER_SECRET": "temp-validation-secret",
     }
 
-    cmd = [sys.executable, entrypoint]
+    # Use the project's Python environment, not the CLI's isolated environment.
+    # find_python_interpreter() looks for .venv/bin/python in cwd, falling back to sys.executable.
```
byrro, line 392: Could this be executed from a different place, a subdirectory perhaps? In that case, I believe it won't find the `.venv`, ignore it, and fallback to the system executable. Perhaps the path should be resolved relative to the project root (where `pyproject.toml` is) or relative to the entrypoint’s directory, not just `cwd`.

I think it would also be interesting to extend the implementation and honor the [VIRTUAL_ENV](https://docs.astral.sh/uv/reference/environment/#virtual_env) and/or [UV_PROJECT_ENVIRONMENT](https://docs.astral.sh/uv/reference/environment/#uv_project_environment) env vars as well, if set.

libs/arcade-core/arcade_core/toolkit.py
```diff
@@ -299,32 +299,50 @@ def tools_from_directory(cls, package_dir: Path, package_name: str) -> dict[str,
...
-        if main_module and hasattr(main_module, "__file__") and main_module.__file__:
-            with contextlib.suppress(Exception):
-                current_file = Path(main_module.__file__).resolve()
+        if main_module:
+            if hasattr(main_module, "__file__") and main_module.__file__:
+                with contextlib.suppress(Exception):
+                    current_file = Path(main_module.__file__).resolve()
+            # Get module name from __spec__ if available (used when paths don't match,
+            # e.g., script runs from bundle but package is in site-packages)
+            main_spec = getattr(main_module, "__spec__", None)
```
byrro, line 310: I believe in a direct script execution `__spec__` won't be set and the skip logic will continue to fail. If this is really the case, perhaps we need to add a fallback comparing the relative path within the package to the suffix of the current_file. If the suffix matches, we skip.

EricGustin: The `test_skips_file_when_path_matches`  unit test I added explicitly sets `__spec__` to `None` and it still successfully skips when it should. So I think we're good? Unless I misunderstood your comment.

libs/arcade-cli/arcade_cli/deploy.py
EricGustin: Fortunately `arcade deploy` requires you to be at the root of the project, so this case will never happen. But it's definitely possible that in the future `arcade deploy` will relax this requirement, so I just added a note about this in `find_python_interpretor` for the future developer who may decide to relax this requirement.

PR #766, opened 2026-02-05
Add Tool Metadata

EricGustin:
This is the first part of a larger project defined here: https://linear.app/arcadedev/project/tool-search-tool-tags-and-categories-static-bef3871e3f45/overview

### The Taxonomy
The current form is to support concrete use-cases that we know of today. There is nothing stopping us from added more dimensions to `Classification`, for example. But let's keep it simple for now! For example, there definitely exists more `ServiceDomain`s that we don't include. Those can be added in the future when we have a toolkit that would use those service domain.

### Resolves:
- https://linear.app/arcadedev/issue/TOO-386/tool-decorator-has-toolmetadata-parameter
- https://linear.app/arcadedev/issue/TOO-387/mcpapp-tools-have-annotations-and-metadata-in-meta-in-their-definition

### MCP Format
```json
{
  "name": "Testingmeta_GetPostsInSubreddit",
  "description": "Get posts from a specific subreddit",
  "inputSchema": {
    "type": "object",
    "properties": {
      "subreddit": {
        "type": "string",
        "description": "The name of the subreddit"
      }
    },
    "additionalProperties": false,
    "required": [
      "subreddit"
    ]
  },
  "annotations": {
    "title": "GetPostsInSubreddit",
    "readOnlyHint": true,
    "destructiveHint": false,
    "idempotentHint": true,
    "openWorldHint": true
  },
  "title": "GetPostsInSubreddit",
  "_meta": {
    "arcade": {
      "requirements": {
        "authorization": {
          "provider_id": "reddit",
          "provider_type": "oauth2",
          "oauth2": {
            "scopes": [
              "read"
            ]
          }
        }
      },
      "metadata": {
        "classification": {
          "service_domains": [
            "social_media"
          ]
        },
        "behavior": {
          "operations": [
            "read"
          ],
          "read_only": true,
          "destructive": false,
          "idempotent": true,
          "open_world": true
        }
      }
    }
  }
}
```

### Arcade Format
```json
{
  "name": "GetPostsInSubreddit",
  "fully_qualified_name": "Testingmeta.GetPostsInSubreddit",
  "description": "Get posts from a specific subreddit",
  "toolkit": {
    "name": "Testingmeta",
    "description": "ArcadeMCP provides access to a wide range of tools and toolkits.Use 'tools/list' to see available tools and 'tools/call' to execute them.",
    "version": "1.0.0"
  },
  "input": {
    "parameters": [
      {
        "name": "subreddit",
        "required": true,
        "description": "The name of the subreddit",
        "value_schema": {
          "val_type": "string",
          "inner_val_type": null,
          "enum": null,
          "properties": null,
          "inner_properties": null,
          "description": null
        },
        "inferrable": true
      }
    ]
  },
  "output": {
    "description": "No description provided.",
    "available_modes": [
      "value",
      "error"
    ],
    "value_schema": {
      "val_type": "json",
      "inner_val_type": null,
      "enum": null,
      "properties": null,
      "inner_properties": null,
      "description": null
    }
  },
  "requirements": {
    "authorization": {
      "provider_id": "reddit",
      "provider_type": "oauth2",
      "id": null,
      "oauth2": {
        "scopes": [
          "read"
        ]
      }
    },
    "secrets": null,
    "metadata": null
  },
  "deprecation_message": null,
  "metadata": {
    "classification": {
      "service_domains": [
        "social_media"
      ]
    },
    "behavior": {
      "operations": [
        "read"
      ],
      "read_only": true,
      "destructive": false,
      "idempotent": true,
      "open_world": true
    },
    "extras": null
  }
}
```

---

> [!NOTE]
> **Medium Risk**
> Touches core tool definition and MCP conversion paths, changing serialized tool shapes and validation behavior. Backward-compatible for tools without metadata, but metadata/annotation/meta output and schema generation changes could affect clients that depend on the previous MCP payload details.
> 
> **Overview**
> Adds first-class `ToolMetadata` support (classification, behavior, extras) to Arcade tools, including strict validation rules, and threads it through the tool lifecycle: `@tool`/`MCPApp.tool` accept `metadata`, `ToolCatalog` validates/stores it, and `ToolDefinition` now exposes `metadata`.
> 
> Updates MCP serialization so tool `annotations` are derived from `metadata.behavior` and `_meta.arcade` includes both tool `requirements` and serialized `metadata`; also includes MCP `outputSchema` when available and removes the previous input-schema fallback path.
> 
> Bumps package versions/dependency minimums, updates templates and the `echo` example to demonstrate metadata usage, and adds a new `tool_metadata` example server plus comprehensive tests for validation and MCP serialization.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 77c886cbf46d1787c3626b72ba5660047e9c5043. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libs/arcade-mcp-server/arcade_mcp_server/convert.py
+++ b/libs/arcade-mcp-server/arcade_mcp_server/convert.py
@@ -1,8 +1,7 @@
 import base64
 import json
 import logging
-from enum import Enum
-from typing import Any, get_args, get_origin
+from typing import Any
 
 from arcade_core.catalog import MaterializedTool
 from arcade_core.schema import ToolDefinition
@@ -218,116 +217,6 @@ def build_input_schema_from_definition(definition: ToolDefinition) -> dict[str, 
             properties[param.name] = schema
             if getattr(param, "required", False):
                 required.append(param.name)
-
-    input_schema: dict[str, Any] = {
-        "type": "object",
-        "properties": properties,
-        "additionalProperties": False,
-    }
-    if required:
-        input_schema["required"] = required
-    return input_schema
-
-
-def _build_input_schema_from_model(tool: MaterializedTool) -> dict[str, Any]:
-    """Build input schema from a tool's input_model as a fallback."""
-    properties: dict[str, Any] = {}
-    required: list[str] = []
-
-    context_param_name = None
-    tool_input = getattr(tool.definition, "input", None)
-    if tool_input is not None:
-        context_param_name = getattr(tool_input, "tool_context_parameter_name", None)
-
-    if (
-        hasattr(tool, "input_model")
-        and tool.input_model is not None
-        and hasattr(tool.input_model, "model_fields")
-    ):
-        for field_name, field in tool.input_model.model_fields.items():
-            if field_name == context_param_name:
-                continue
-
-            field_type = getattr(field, "annotation", None)
-            field_type_name = "string"  # default
-
-            if field_type is int:
-                field_type_name = "integer"
-            elif field_type is float:
-                field_type_name = "number"
-            elif field_type is bool:
-                field_type_name = "boolean"
-            elif field_type is list or (getattr(field_type, "__origin__", None) is list):
-                field_type_name = "array"
-            elif field_type is dict or (getattr(field_type, "__origin__", None) is dict):
-                field_type_name = "object"
-
-            field_description = getattr(field, "description", None) or f"Parameter: {field_name}"
-
-            param_def: dict[str, Any] = {
-                "type": field_type_name,
-                "description": field_description,
-            }
-
-            # Enum support: Enum classes or typing.Annotated[...] with Enum
-            enum_type = None
-            ann = getattr(field, "annotation", None)
-            if ann is not None:
-                origin = get_origin(ann)
-                args = get_args(ann)
-                # typing.Annotated[Enum, ...]
-                if origin is not None and args:
-                    for arg in args:
-                        if isinstance(arg, type) and issubclass(arg, Enum):
-                            enum_type = arg
-                            break
-                elif isinstance(ann, type) and issubclass(ann, Enum):
-                    enum_type = ann
-            if enum_type is not None:
-                param_def["enum"] = [e.value for e in enum_type]
-
-            # Literal[...] support for enum-like constraints
-            if ann is not None and get_origin(ann) is None:
-                pass  # no-op, handled above
-            elif ann is not None and get_origin(ann) is Any:
-                pass
-            else:
-                if get_origin(ann) is None:
-                    ...
-
-            # Attempt to infer inner list item types for list[T]
-            if field_type_name == "array":
-                inner = None
-                if get_origin(field_type) is list and get_args(field_type):
-                    inner = get_args(field_type)[0]
-                if inner is int:
-                    param_def["items"] = {"type": "integer"}
-                elif inner is float:
-                    param_def["items"] = {"type": "number"}
-                elif inner is bool:
-                    param_def["items"] = {"type": "boolean"}
-                elif inner is str:
-                    param_def["items"] = {"type": "string"}
-
-            properties[field_name] = param_def
-
-            # Required detection with multiple strategies
-            is_required_attr = getattr(field, "is_required", None)
-            try:
-                if callable(is_required_attr):
-                    if is_required_attr():
-                        required.append(field_name)
-                elif isinstance(is_required_attr, bool) and is_required_attr:
-                    required.append(field_name)
-                else:
-                    has_default = getattr(field, "default", None) is not None
-                    has_factory = getattr(field, "default_factory", None) is not None
-                    if not (has_default or has_factory):
-                        required.append(field_name)
-            except Exception:
-                logger.debug(
-                    f"Could not determine if field {field_name} is required, assuming optional"
-                )
 
     input_schema: dict[str, Any] = {
         "type": "object",


libs/arcade-cli/arcade_cli/templates/minimal/{{ toolkit_name }}/src/{{ toolkit_name }}/server.py
```diff
@@ -33,7 +41,22 @@ def whisper_secret(context: Context) -> Annotated[str, "The last 4 characters of
 
 # To use this tool locally, you need to install the Arcade CLI (uv tool install arcade-mcp)
 # and then run 'arcade login' to authenticate.
-@app.tool(requires_auth=Reddit(scopes=["read"]))
+@app.tool(
```
EricGustin, line 43: Moving forward, all new servers created with `arcade new` will contain a tool that has metadata defined

libs/arcade-core/arcade_core/metadata.py
```diff
@@ -0,0 +1,391 @@
+"""
```
EricGustin, line 1: This file is intentionally "overly" documented because it's essentially our public API. Thoroughly defining every single ServiceDomain, Verb, etc. is important so that users of the framework can feel confident about how to define metadata for their tools

libs/arcade-core/arcade_core/metadata.py
```diff
@@ -0,0 +1,391 @@
...
+    The capability areas this tool operates in.
+
+    Used for Arcade tool selection boosting.
+    Tools can specify multiple domains to accurately represent their capabilities.
+
+    Choose the domain that answers: "What is this tool fundamentally FOR?"
+    """
+
+    # === Communication ===
+    MESSAGING = "messaging"
```
Spartee: I don't think few shot should be static like this. I would remove the exampels.

Spartee: at very least I would remove the specific tool names.

libs/arcade-core/arcade_core/metadata.py
```diff
@@ -0,0 +1,391 @@
...
+
+from enum import Enum
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict
+
+from arcade_core.errors import ToolDefinitionError
+
+
+class Domain(str, Enum):
```
Spartee: I think there's too many here at different levels of abstraction to the point where the classification isn't appropriate. consistency matters most. for example

What domain does this go in?

```text
MyApp.SearchDocumentDatabase
```

`storage`?
`data`?
`document`?
`search`?

the goal is to be simple enough to make it impossible to make a mistake. 

The less overlap the categories have the better. I would actually prefer to have multiple combined like ``files`` ``search``. 

Think of it as a venn diagram.. the more circles the more overlap.

lmk if this doens't mke sense.

libs/arcade-core/arcade_core/metadata.py
```diff
@@ -0,0 +1,391 @@
...
+class SystemType(str, Enum):
+    """
+    The type of system this tool interfaces with.
+
+    Provides orthogonal signal that disambiguates tools with the same domain.
+    For example, "read temperature" could match IoT sensors (HARDWARE) or
+    web weather scrapers (WEB) - SystemType helps distinguish them.
+    """
+
+    SAAS_API = "saas_api"
```
Spartee: no acronyms. this has little semantics and is easily confused actually. this may get tokenized as ``SA as``

libs/arcade-core/arcade_core/metadata.py
```diff
@@ -0,0 +1,391 @@
...
+    """Third-party SaaS platforms accessed via their APIs. Examples: Slack, Salesforce, GitHub, Stripe"""
+
+    DATABASE = "database"
+    """Data storage systems with query interfaces. Examples: PostgreSQL, MongoDB, Redis, Elasticsearch"""
+
+    FILE_SYSTEM = "file_system"
```
Spartee: Files

libs/arcade-core/arcade_core/metadata.py
```diff
@@ -0,0 +1,391 @@
...
+    """File storage systems (local or cloud). Examples: Local disk, S3, Google Drive, Dropbox"""
+
+    WEB = "web"
+    """Web pages, browsers, or web scraping. Examples: Firecrawl, Playwright, web search"""
+
+    OPERATING_SYSTEM = "operating_system"
```
Spartee: Operating system AND files?

libs/arcade-core/arcade_core/metadata.py
```diff
@@ -0,0 +1,391 @@
...
+    """Autonomous AI agents that perform multi-step tasks. Examples: Research agents, planning agents"""
+
+    CUSTOM_API = "custom_api"
+    """Customer's internal or private APIs. Examples: Internal microservices, proprietary systems"""
+
+    SELF_CONTAINED = "self_contained"
+    """No external system — pure computation, fully self-contained. Examples: Math.Add, JSON.Parse, string formatting"""
+
+
+class Verb(str, Enum):
```
Spartee: this is the best one.

libs/arcade-core/arcade_core/metadata.py
EricGustin: I see where you're coming from, but the example provided is throwing me off. We don't decide on how to classify a tool based on only a vague name like `MyApp.SearchDocumentDatabase`. Tools have annotated parameters, docstrings that describe the tool, a toolkit name that is more descriptive than "MyApp", and actual implementation code - all of which provide important disambiguation when it is time to add metadata to it. I _think_ this issue goes away when looking at more concrete examples. Can you provide a concrete example where this problem still exists?

I agree with your goal of needing to be simple enough to make it impossible to make a mistake. But in a world of 100,000+ tools (not that far away), I see granularity as more of a feature than a bug. And if a tool genuinely spans more than one `Domain`, then the tool can have more than one - it's multi-select. That "overlap" allows for additional boosting signal for tools that are genuinely cross-domain, which at 100,000+ tools provides the granularity/distinction needed at that scale.

I'm open to the idea of less overlap between domains as long as I don't get bogged down in hours of naming nuances (again) 🙂

libs/arcade-core/arcade_core/metadata.py
EricGustin: intended to distinguish between talking to a runtime env like "run this script" (OPERATING_SYSTEM) vs some file-based storage layer like local disk, s3 (FILE_SYSTEM). I could see an argument for removing the FILE_SYSTEM `SystemType` entirely since we already have a STORAGE `Domain` though.

libs/arcade-core/arcade_core/metadata.py
EricGustin: I could be on board for one of these: THIRD_PARTY_SERVICE, CLOUD_PLATFORM

EricGustin: And then rename CUSTOM_API to FIRST_PARTY_SERVICE or PROPRIETARY_SERVICE

byrro:
`convert.py` [seems to still be emitting](https://github.com/ArcadeAI/arcade-mcp/blob/dbdec637b177c81fbff6d7850e2753aee1e2fa98/libs/arcade-mcp-server/arcade_mcp_server/convert.py#L74) `meta` in the old format

EricGustin:
> `convert.py` [seems to still be emitting](https://github.com/ArcadeAI/arcade-mcp/blob/dbdec637b177c81fbff6d7850e2753aee1e2fa98/libs/arcade-mcp-server/arcade_mcp_server/convert.py#L74) `meta` in the old format

oh, great catch!

libs/arcade-core/arcade_core/catalog.py
```diff
@@ -464,6 +464,10 @@ def create_tool_definition(
         tool_name = snake_to_pascal_case(raw_tool_name)
         fully_qualified_name = FullyQualifiedName.from_toolkit(tool_name, toolkit_definition)
         deprecation_message = getattr(tool, "__tool_deprecation_message__", None)
+        tool_metadata = getattr(tool, "__tool_metadata__", None)
+
+        if tool_metadata is not None:
+            tool_metadata.validate_for_tool()
```
byrro, line 476: More of a DX nit:

It'd be nice to have a guard here ensuring this is actually a `ToolMetadata` instance. If the developer use smth else like a `dict`, this is going to raise a cryptic `AttributeError` with no hint to what is wrong or how to fix the tool definition.

EricGustin:
Addressed your comments @byrro 

1. Delegated `_to_dto` to `create_mcp_tool` when converting Arcade format to MCP format
2. Added the guard w/ helpful error message

libs/tests/arcade_mcp_server/test_convert.py
```diff
@@ -371,50 +371,10 @@ def f() -> Annotated[str, "result"]:
...
-
-        assert schema["type"] == "object"
-        assert "properties" in schema
-        assert "name" in schema["properties"]
-        assert "age" in schema["properties"]
-
-        # Ensure the schema was built from the model and not the definition
-        assert schema["properties"]["name"]["type"] == "string"
-        assert schema["properties"]["age"]["type"] == "integer"
+    def test_output_schema_included(self, materialized_tool):
```
EricGustin, line 374: no longer doing that fallback logic - we want to propogate the error instead of silencing it with a sub-par fallback

PR #768, opened 2026-02-10
[TOO-326] Windows papercuts

jottakka:
> [!NOTE]
> **Medium Risk**
> Touches authentication/login flow, credentials-file permissions, and subprocess lifecycle behavior across platforms; while mostly defensive, regressions could impact login or process management on Windows/macOS runners.
> 
> **Overview**
> Improves Windows/cross-platform reliability across the CLI and MCP server: OAuth login now binds the callback server to `127.0.0.1`, avoids slow loopback reverse-DNS, adds a configurable callback timeout (`--timeout` + env default), and opens URLs via a Windows-friendly `_open_browser` to avoid flashing console windows.
> 
> Centralizes CLI output via a shared `console` that forces UTF-8 on Windows, standardizes UTF-8 file reads/writes throughout, tightens credentials-file permissions on Windows using `icacls`, and adds shared Windows subprocess helpers for **no-window** process creation and graceful termination (used by `deploy`, MCP reload, and usage-tracking worker).
> 
> Updates client configuration UX/robustness (Windows AppData resolution via `platformdirs`, Cursor config path fallbacks + compatibility writes, overwrite warnings, absolute `uv` path for GUI clients, safer path display) and improves `deploy` child-process handling to avoid pipe-buffer deadlocks while giving better debug-aware error messages.
> 
> Expands CI to run tests on Linux/Windows/macOS, adds a no-auth CLI integration workflow, disables usage tracking in toolkits CI, and adds extensive regression tests for Windows signals, subprocess cleanup, UTF-8, and config-path edge cases; bumps `arcade-core` to `4.4.2` and `arcade-mcp-server` to `1.17.2` (with updated dependency pin).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0fabd8ca1cd647039ba6ddbdf3f7809c330bab9e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libs/tests/arcade_mcp_server/integration/test_end_to_end.py
+++ b/libs/tests/arcade_mcp_server/integration/test_end_to_end.py
@@ -8,8 +8,9 @@ import asyncio
 import json
 import os
 import queue
-import random
+import socket
 import subprocess
+import sys
 import threading
 import time
 from pathlib import Path
@@ -24,6 +25,43 @@ import pytest
 def get_entrypoint_path() -> str:
     """Get the path to the test server entrypoint."""
     return str(Path(__file__).parent / "server" / "src" / "server" / "entrypoint.py")
+
+
+HTTP_STARTUP_TIMEOUT_SECONDS = 30 if sys.platform == "win32" else 10
+
+
+def _find_open_tcp_port() -> int:
+    """Reserve a free loopback TCP port and return it."""
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.bind(("127.0.0.1", 0))
+        return int(sock.getsockname()[1])
+
+
+def _cleanup_process(process: subprocess.Popen, timeout: float = 5.0) -> None:
+    """Terminate subprocesses reliably, including uv child trees on Windows."""
+    if process.poll() is not None:
+        return
+
+    if sys.platform == "win32":
+        subprocess.run(
+            ["taskkill", "/PID", str(process.pid), "/T", "/F"],
+            check=False,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+        try:
+            process.wait(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            process.kill()
+            process.wait()
+        return
+
+    process.terminate()
+    try:
+        process.wait(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        process.kill()
+        process.wait()
 
 
 def start_mcp_server(
@@ -58,7 +96,7 @@ def start_mcp_server(
 
     elif transport == "http":
         if port is None:
-            port = random.randint(8000, 9000)  # noqa: S311
+            port = _find_open_tcp_port()
 
         env = {
             **os.environ,
@@ -72,8 +110,8 @@ def start_mcp_server(
         cmd = ["uv", "run", entrypoint_path, "http"]
         process = subprocess.Popen(
             cmd,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
             text=True,
             env=env,
             cwd=str(package_path),
@@ -106,7 +144,7 @@ def start_mcp_server_direct_python(
         pytest.skip("Server venv not found - run 'uv sync' in integration/server first")
 
     if port is None:
-        port = random.randint(8000, 9000)  # noqa: S311
+        port = _find_open_tcp_port()
 
     env = {
         **os.environ,
@@ -120,8 +158,8 @@ def start_mcp_server_direct_python(
     cmd = [str(venv_python), entrypoint_path, transport]
     process = subprocess.Popen(
         cmd,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
         text=True,
         env=env,
         cwd=str(package_path),
@@ -129,21 +167,32 @@ def start_mcp_server_direct_python(
     return process, port if transport == "http" else None
 
 
-def wait_for_http_server_ready(port: int, timeout: int = 30) -> None:
+def wait_for_http_server_ready(
+    port: int,
+    timeout: int = HTTP_STARTUP_TIMEOUT_SECONDS,
+    process: subprocess.Popen | None = None,
+) -> None:
     """
     Wait for HTTP server to become healthy.
 
     Args:
         port: Server port
         timeout: Maximum time to wait in seconds
+        process: Optional subprocess handle for early-exit detection
 
     Raises:
         TimeoutError: If server doesn't become healthy within timeout
+        RuntimeError: If process exits before becoming healthy
     """
     health_url = f"http://127.0.0.1:{port}/worker/health"
-    start_time = time.time()
+    deadline = time.monotonic() + timeout
 
-    while time.time() - start_time < timeout:
+    while time.monotonic() < deadline:
+        if process is not None and process.poll() is not None:
+            raise RuntimeError(
+                f"Server process exited with code {process.returncode} "
+                f"before becoming healthy on port {port}"
+            )
         try:
             response = httpx.get(health_url, timeout=2.0)
             if response.status_code == 200:
@@ -573,12 +622,7 @@ async def test_stdio_e2e():
 
     finally:
         # Clean shutdown
-        process.terminate()
-        try:
-            process.wait(timeout=5)
-        except subprocess.TimeoutExpired:
-            process.kill()
-            process.wait()
+        _cleanup_process(process)
 
 
 @pytest.mark.asyncio
@@ -590,7 +634,11 @@ async def test_http_e2e():
     base_url = f"http://127.0.0.1:{port}"
 
     try:
-        wait_for_http_server_ready(port, timeout=10)
+        wait_for_http_server_ready(
+            port,
+            timeout=HTTP_STARTUP_TIMEOUT_SECONDS,
+            process=process,
+        )
 
         headers = {
             "Content-Type": "application/json",
@@ -715,12 +763,7 @@ async def test_http_e2e():
         client.close()
 
     finally:
-        process.terminate()
-        try:
-            process.wait(timeout=5)
-        except subprocess.TimeoutExpired:
-            process.kill()
-            process.wait()
+        _cleanup_process(process)
 
 
 @pytest.mark.asyncio
@@ -732,7 +775,11 @@ async def test_http_mcp_concurrent_tool_execution():
     base_url = f"http://127.0.0.1:{port}"
 
     try:
-        wait_for_http_server_ready(port, timeout=10)
+        wait_for_http_server_ready(
+            port,
+            timeout=HTTP_STARTUP_TIMEOUT_SECONDS,
+            process=process,
+        )
 
         headers = {
             "Content-Type": "application/json",
@@ -815,12 +862,7 @@ async def test_http_mcp_concurrent_tool_execution():
             assert total_time < max_expected_time
 
     finally:
-        process.terminate()
-        try:
-            process.wait(timeout=5)
-        except subprocess.TimeoutExpired:
-            process.kill()
-            process.wait()
+        _cleanup_process(process)
 
 
 @pytest.mark.asyncio
@@ -832,7 +874,11 @@ async def test_http_worker_concurrent_tool_execution():
     base_url = f"http://127.0.0.1:{port}"
 
     try:
-        wait_for_http_server_ready(port, timeout=10)
+        wait_for_http_server_ready(
+            port,
+            timeout=HTTP_STARTUP_TIMEOUT_SECONDS,
+            process=process,
+        )
 
         headers = {
             "Content-Type": "application/json",
@@ -891,12 +937,7 @@ async def test_http_worker_concurrent_tool_execution():
             assert total_time < max_expected_time
 
     finally:
-        process.terminate()
-        try:
-            process.wait(timeout=5)
-        except subprocess.TimeoutExpired:
-            process.kill()
-            process.wait()
+        _cleanup_process(process)
 
 
 @pytest.mark.asyncio
@@ -908,7 +949,11 @@ async def test_http_mixed_route_concurrent_execution():
     base_url = f"http://127.0.0.1:{port}"
 
     try:
-        wait_for_http_server_ready(port, timeout=10)
+        wait_for_http_server_ready(
+            port,
+            timeout=HTTP_STARTUP_TIMEOUT_SECONDS,
+            process=process,
+        )
 
         headers = {
             "Content-Type": "application/json",
@@ -999,12 +1044,7 @@ async def test_http_mixed_route_concurrent_execution():
             assert total_time < max_expected_time
 
     finally:
-        process.terminate()
-        try:
-            process.wait(timeout=5)
-        except subprocess.TimeoutExpired:
-            process.kill()
-            process.wait()
+        _cleanup_process(process)
 
 
 @pytest.mark.asyncio
@@ -1014,7 +1054,11 @@ async def test_http_direct_python_invocation():
     assert port is not None
 
     try:
-        wait_for_http_server_ready(port, timeout=10)
+        wait_for_http_server_ready(
+            port,
+            timeout=HTTP_STARTUP_TIMEOUT_SECONDS,
+            process=process,
+        )
 
         # Verify server is healthy and tools are discoverable
         headers = {
@@ -1056,9 +1100,4 @@ async def test_http_direct_python_invocation():
         client.close()
 
     finally:
-        process.terminate()
-        try:
-            process.wait(timeout=5)
-        except subprocess.TimeoutExpired:
-            process.kill()
-            process.wait()
+        _cleanup_process(process)


libs/arcade-cli/arcade_cli/authn.py
```diff
@@ -357,21 +384,52 @@ def __init__(self, state: str, port: int = LOCAL_CALLBACK_PORT):
         self.port = port
         self.httpd: HTTPServer | None = None
         self.result: dict[str, Any] = {}
+        self.result_event = threading.Event()
+        self.ready_event = threading.Event()
 
     def run_server(self) -> None:
```
jottakka, line 426: To avoid affecting previous behavior it should only be called for the widows flow

libs/arcade-cli/arcade_cli/authn.py
```diff
@@ -496,36 +630,38 @@ def build_coordinator_url(host: str, port: int | None) -> str:
...
         with oauth_callback_server(state) as server:
             # server is running and waiting for callback
             ...
-        # After the with block, the server has received the callback
+        # After the with block, the server has been shut down
     """
-    server = OAuthCallbackServer(state)
-    server_thread = threading.Thread(target=server.run_server)
+    server = OAuthCallbackServer(state, port=port)
+    server_thread = threading.Thread(target=server.run_server, daemon=True)
```
jottakka, line 686: how daemon true affects it ?

libs/arcade-cli/arcade_cli/configure.py
```diff
@@ -53,6 +52,73 @@ def get_windows_username() -> str | None:
...
+    """Resolve Windows roaming AppData via optional ``platformdirs``.
+
+    ``platformdirs`` is the de-facto standard Python library for OS-specific
+    user directory resolution. We import it lazily and treat it as optional so
+    existing environments keep working without new hard dependencies.
+    """
+    try:
+        from platformdirs import user_data_dir
+    except ImportError:
+        return None
```
jottakka: for windows here should return some info saying that this install is required instead of silent fail

libs/arcade-cli/arcade_cli/configure.py
```diff
@@ -53,6 +52,73 @@ def get_windows_username() -> str | None:
...
+
+    try:
+        candidate = user_data_dir(appname=None, appauthor=False, roaming=True)
+    except TypeError:
+        # Compatibility with older signatures that rely on positional args.
```
jottakka: add a debug log here

libs/arcade-cli/arcade_cli/configure.py
```diff
@@ -53,6 +52,73 @@ def get_windows_username() -> str | None:
...
+        candidate = user_data_dir(None, False, roaming=True)
+
+    if not candidate:
+        return None
+    return Path(candidate)
+
+
+def _resolve_windows_appdata() -> Path:
+    """Resolve the Windows roaming AppData directory with fallbacks."""
```
jottakka: are all those fallbacks really needed ? lets stick with one path if it is works always.

libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -362,6 +367,48 @@ def exclude_filter(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo | None:
...
+                exc,
+            )
+    try:
+        process.terminate()
+    except OSError as exc:
+        logger.debug("terminate() failed during graceful shutdown: %s", exc)
+
+
+def _resolve_server_process_stdio(debug: bool) -> tuple[int | None, int | None]:
+    """Choose stdio strategy for the temporary validation server process.
```
jottakka: explain this better

libs/arcade-core/arcade_core/config_model.py
```diff
@@ -1,11 +1,43 @@
...
 from typing import Any
 
 import yaml
 from pydantic import BaseModel, ConfigDict, ValidationError
 
+logger = logging.getLogger(__name__)
+
+
+def _set_windows_owner_acl(config_file_path: Path) -> None:
+    """Apply owner-only style ACLs on Windows using ``icacls``.
```
jottakka: add better comments here pls

libs/arcade-core/arcade_core/config_model.py
```diff
@@ -230,7 +262,18 @@ def save_to_file(self) -> None:
 
         # Convert to dict, excluding None values for cleaner output
         data = {"cloud": self.model_dump(exclude_none=True, mode="json")}
-        config_file_path.write_text(yaml.dump(data, default_flow_style=False))
+        config_file_path.write_text(yaml.dump(data, default_flow_style=False), encoding="utf-8")
 
-        # Set restrictive permissions (owner read/write only)
-        config_file_path.chmod(0o600)
+        # Set restrictive permissions (owner read/write only).
+        # On Windows, prefer ACL hardening via icacls.
```
jottakka: better explanation of what it does and what it means

.pre-commit-config.yaml
```diff
@@ -17,6 +17,6 @@ repos:
     hooks:
       - id: ruff
         args: [--fix]
-        exclude: "(.*/templates/.*|libs/tests/.*)"
```
jottakka, line 20: wouldnt this change what it is being ignored ? pls check

pyproject.toml
```diff
@@ -42,6 +43,14 @@ all = [
     "numpy>=2.0.0",
     "scikit-learn>=1.5.0",
     "pytz>=2024.1",
+    # platform-aware user directories (optional Windows path resolution)
+    "platformdirs>=4.3.6",
+    # mcp server
+    "arcade-mcp-server>=1.15.3,<2.0.0",
```
EricGustin: this is already included in the required dependencies, so not needed here as an optional dependency

pyproject.toml
```diff
@@ -42,6 +43,14 @@ all = [
...
+    # serve
+    "arcade-serve>=3.2.0,<4.0.0",
+    # tdk
+    "arcade-tdk>=3.4.0,<4.0.0",
```
EricGustin: arcade-serve and arcade-tdk are transient dependencies of arcade-mcp-server (a required dep), so they don't need to be here

pyproject.toml
```diff
@@ -154,3 +167,4 @@ ignore = ["E501", "S105"]
 
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401"]
+"tests/**/*.py" = ["C901"]
```
EricGustin: We should reduce the cyclomatic complexity of the test cases instead of ignoring the problem

libs/arcade-cli/arcade_cli/authn.py
```diff
@@ -45,11 +52,22 @@ def _render_template(template_name: str, **context: Any) -> bytes:
...
+def _get_default_oauth_timeout_seconds() -> int:
+    value = os.environ.get("ARCADE_LOGIN_TIMEOUT_SECONDS", str(DEFAULT_OAUTH_TIMEOUT_SECONDS))
+    try:
+        parsed = int(value)
+        return parsed if parsed > 0 else DEFAULT_OAUTH_TIMEOUT_SECONDS  # noqa: TRY300
+    except ValueError:
+        return DEFAULT_OAUTH_TIMEOUT_SECONDS
+
+
+DEFAULT_OAUTH_TIMEOUT_SECONDS = _get_default_oauth_timeout_seconds()
```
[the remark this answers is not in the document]
EricGustin: please address this

libs/arcade-cli/arcade_cli/authn.py
```diff
@@ -562,21 +732,39 @@ def status(msg: str) -> None:
...
+            # Step 4: Generate authorization URL and open browser
+            auth_url, code_verifier = generate_authorization_url(
+                oauth_client, cli_config, redirect_uri, state
+            )
+
+            # Always print the auth link first so users can click/copy it
+            # immediately, even if automatic browser launch fails.
+            status("Use this authorization link if needed:\n" f"{auth_url}")
+            status("Opening a browser to log you in...")
+            browser_opened = _open_browser(auth_url)
```
EricGustin: Were you running into issues with the previous logic which only printed the auth link if browser opening failed? I'm of the perspective that we should only print the auth link if opening the browser fails to prevent polluting the output of `arcade login`

libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -362,6 +367,57 @@ def exclude_filter(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo | None:
...
+    On Windows, ``process.terminate()`` calls ``TerminateProcess`` which kills
+    the child immediately.  If the child was started with
+    ``CREATE_NEW_PROCESS_GROUP`` we can send ``CTRL_BREAK_EVENT`` first so
+    Python's default handler raises ``KeyboardInterrupt`` — giving the child
+    a chance to clean up.
+    """
+    if sys.platform == "win32":
+        try:
+            process.send_signal(signal.CTRL_BREAK_EVENT)  # type: ignore[attr-defined]
+            return  # noqa: TRY300
```
EricGustin: this should be in an `else` after the `except`

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -144,7 +148,16 @@ def logout(
...
         if os.path.exists(CREDENTIALS_FILE_PATH):
-            os.remove(CREDENTIALS_FILE_PATH)
+            try:
+                os.remove(CREDENTIALS_FILE_PATH)
+            except PermissionError:
+                # On Windows, the file may be locked by another process.
+                handle_cli_error(
+                    "Could not remove credentials file — it may be in use by another process. "
+                    "Close other Arcade instances and try again.",
+                    should_exit=False,
```
EricGustin: This should be `should_exit=True`

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -163,8 +176,11 @@ def whoami(
 
     try:
         config = Config.load_from_file()
+    except FileNotFoundError:
+        console.print("Not logged in. Run 'arcade login' to authenticate.", style="bold red")
+        return
     except Exception as e:
-        handle_cli_error("Failed to read credentials", e, debug)
+        handle_cli_error("Failed to read credentials", e, debug, should_exit=False)
```
EricGustin: Any reason why we're changing this from the default `should_exit-False` to `should_exit=True`? Seems like this should exit

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -311,8 +327,15 @@ def mcp(
...
-        # Execute the command and pass through all output
-        result = subprocess.run(cmd, check=False)
+        # Execute the command and pass through all output.
+        # On Windows, set CREATE_NO_WINDOW to prevent a phantom console
+        # window from appearing (e.g. when an MCP client spawns this
+        # command without an attached console).  The child process still
+        # inherits stdin/stdout/stderr for stdio transport communication.
+        run_kwargs: dict[str, object] = {"check": False}
+        if sys.platform == "win32":
+            run_kwargs["creationflags"] = subprocess.CREATE_NO_WINDOW
```
EricGustin: There is a lot of varying "suppress phantom window" logic spread across various parts of this PR. Is it possible to consolidate this logic into a single helper? Something that would help the future developer?

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -943,6 +966,7 @@ def main_callback(
     public_commands = {
         login.__name__,
         logout.__name__,
+        whoami.__name__,
```
EricGustin: "whoami" seems like a command that would require the user to be logged in. Why do we add it as a public command here?

libs/arcade-cli/arcade_cli/new.py
```diff
@@ -271,9 +273,14 @@ def create_new_toolkit_minimal(output_directory: str, toolkit_name: str) -> None
 
     try:
         create_package(env, template_directory, toolkit_directory, context, ignore_pattern)
+        console.print("")
         console.print(
             f"[green]Server '{toolkit_name}' created successfully at '{toolkit_directory}'.[/green]"
         )
+        console.print("Next steps:", style="bold")
+        console.print(f"  cd {toolkit_directory / toolkit_name / 'src' / toolkit_name}")
+        console.print("  uv run server.py")
```
EricGustin: DX: give them the command for stdio and for http

libs/arcade-core/arcade_core/usage/identity.py
```diff
@@ -45,7 +45,7 @@ def load_or_create(self) -> dict[str, Any]:
 
         if os.path.exists(self.usage_file_path):
             try:
-                with open(self.usage_file_path) as f:
+                with open(self.usage_file_path, encoding="utf-8") as f:
```
EricGustin, line 48: what happens if a future developer forgets to explicitly set `encoding="utf-8"` ? We need a linting rule, or a pre-commit hook, or a unit test to prevent a regression here

tests/install/README.md
```diff
@@ -158,7 +192,18 @@ If you see import errors, make sure you're running from the project root and tha
...
 ```
 
+### Optional eval/display test dependencies
+
+Some CLI test files (for eval/display formatting paths) require optional eval
+dependencies such as `openai`, `pytz`, and `numpy`. Install the eval extras
+before running those suites:
+
+```bash
+uv sync --dev --extra evals
```
EricGustin: It's no longer `uv sync --dev`, but instead it's `uv sync --extra dev`. Please update in other locations too

tests/integration/mcp_protocol_smoke.py
```diff
@@ -0,0 +1,670 @@
+#!/usr/bin/env python3
```
EricGustin, line 1: Does this differ much from the existing https://github.com/ArcadeAI/arcade-mcp/blob/main/libs/tests/arcade_mcp_server/integration/test_end_to_end.py ? Seems like it's testing a lot of the same stuff and duplicating a lot of the boilerplate.

pyproject.toml
```diff
@@ -42,6 +43,14 @@ all = [
     "numpy>=2.0.0",
     "scikit-learn>=1.5.0",
     "pytz>=2024.1",
+    # platform-aware user directories (optional Windows path resolution)
+    "platformdirs>=4.3.6",
```
EricGustin: this is already included in the required dependencies, so not needed here as an optional dependency

.github/workflows/test-install.yml
```diff
@@ -40,9 +40,19 @@ jobs:
                   uv pip install -e .
 
             - name: Run installation test
+              env:
+                  ARCADE_USAGE_TRACKING: "0"
```
EricGustin: Consider adding something like [this](https://github.com/ArcadeAI/arcade-mcp/blob/a918eef0371b83cab23b39be487fdf861977516a/libs/tests/conftest.py#L43-L61) to the integration tests

.github/workflows/test-install.yml
```diff
@@ -40,9 +40,19 @@ jobs:
...
               run: |
                   python tests/install/test_install.py
 
+            - name: Restore full test environment
+              run: |
+                  uv sync --extra all --extra dev
+
+            - name: Run CLI unit tests
+              run: |
+                  uv run python -m pytest libs/tests/cli/ -v --tb=short --ignore=libs/tests/cli/test_display.py --ignore=libs/tests/cli/test_main_evals.py --ignore=libs/tests/cli/deploy/
```
EricGustin: The `.github/workflows/main.yml` workflow already runs the CLI unit tests. What’s the reason we’re running them again in `test-install.yml`? We'd be running CLI unit tests for linux twice.

It looks like `test-install.yml`’s scope has shifted beyond “test installing the CLI,” so the name feels misleading now. I’d prefer we split responsibilities more clearly: one workflow dedicated to unit tests (with an OS matrix), and a separate workflow dedicated to integration tests (also with an OS matrix).

.github/workflows/windows-cli-integration-no-auth.yml
```diff
@@ -0,0 +1,38 @@
...
+        strategy:
+            matrix:
+                python-version: ["3.10", "3.11", "3.12"]
+            fail-fast: false
+
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v4
+
+            - name: Install uv
```
EricGustin: we have a github action for setting up the uv env. Maybe it will help. `.github/actions/setup-uv-env`

libs/tests/cli/test_windows_subprocess.py
```diff
@@ -0,0 +1,239 @@
...
+        """Source must use CTRL_BREAK_EVENT for graceful shutdown on Windows."""
+        import arcade_mcp_server.mcp_app as mcp_mod
+        source = inspect.getsource(mcp_mod)
+        assert "CTRL_BREAK_EVENT" in source
+
+    def test_checks_platform(self) -> None:
+        """Source must check sys.platform before setting Windows flags."""
+        import arcade_mcp_server.mcp_app as mcp_mod
+        source = inspect.getsource(mcp_mod)
+        assert 'sys.platform == "win32"' in source or "sys.platform == 'win32'" in source
```
EricGustin: testing source code string matching is very brittle. They're not really testing behavior

libs/tests/cli/test_cross_platform.py
```diff
@@ -0,0 +1,1047 @@
...
+  3. Console encoding safety (UnicodeEncodeError prevention)
+  4. Browser opening without CMD flash
+  5. File locking error handling (os.remove, shutil.rmtree)
+  6. Config model UTF-8 + chmod safety
+  7. Path formatting for display
+  8. OAuth callback loopback binding
+  9. Signal handling logging level
+
+All tests run on BOTH Unix and Windows — platform-specific behaviour is
+tested by mocking ``sys.platform``.
```
EricGustin: I'm hesitant to always mock `sys.platform`. Mocking `sys.platform` seems like it should only be done for very specific scenarios (like testing fallback logic or error handling branches). 

What would be better is if ALL of our tests are run in CI across an OS matrix. If we ran all of our tests in an OS matrix, then this entire file could go away (it's mostly just duplicating existing tests)

EricGustin: For tests that are truly only for a specific OS, then just use `pytest.mark.skipif(sys.platform != "win32", reason="...")`

pyproject.toml
```diff
@@ -20,7 +20,7 @@ requires-python = ">=3.10"
 dependencies = [
     # CLI dependencies
     "arcade-mcp-server>=1.17.0,<2.0.0",
-    "arcade-core>=4.4.0,<5.0.0",
+    "arcade-core>=4.4.2,<5.0.0",
```
EricGustin, line 23: arcade-core was bumped to 4.4.1 in this PR, but the dependency is requiring at least 4.4.2. This will completely break the CLI because it will fail to resolve when someone tries to install the CLI

EricGustin (changes requested):
Thank you for tackling this. The callback work, binding fix, and smoke tests are great. 

There are a few blockers before this can merge:
1. The version mismatch in the CLI's dependency will break the CLI when released
2. We should be running the full test suite on the OS matrix.
3. The phantom-window suppression pattern is copied many times throughout with the same Windows logic. Someone in the future is going to forget to paste it again. I would like to see shared subprocess helpers.
4. We need a linting rule to prevent encoding=utf-8 regressions in the future
5. The 1000+ line file `test_cross_platform.py` can arguably be deleted since it's nearly entirely duplicated tests

See inline comments for specifics

libs/arcade-cli/arcade_cli/authn.py
jottakka: There were times were it openned in other browser that I wasnt logged in, then I had to login and it would skip the oauth flow other times i just wanted to open in other browser session. I believe having it there is kinda important for usability and give users the option to copy and paste it.

libs/tests/core/test_dependency_alignment.py
jottakka, line 1: @EricGustin need some feedback about this test, should we have something as its own step?

.github/workflows/cli-integration-no-auth.yml
```diff
@@ -0,0 +1,39 @@
...
+        name: CLI Integration No-Auth (${{ matrix.os }}, Python ${{ matrix.python-version }})
+        runs-on: ${{ matrix.os }}
+        timeout-minutes: 35
+        env:
+            ARCADE_USAGE_TRACKING: "0"
+
+        strategy:
+            matrix:
+                os: [ubuntu-latest, windows-latest, macos-latest]
+                python-version: ["3.10", "3.11", "3.12"]
```
EricGustin, line 21: We also support Python 3.13 and 3.14. This would bring this matrix up to 15 jobs. How many jobs are too many jobs (genuine question - I don't know if we have an answer)?

.github/workflows/cli-integration-no-auth.yml
```diff
@@ -0,0 +1,39 @@
...
+              uses: actions/checkout@v4
+
+            - name: Set up environment
+              uses: ./.github/actions/setup-uv-env
+              with:
+                  python-version: ${{ matrix.python-version }}
+
+            - name: Install arcade-mcp from source
+              run: |
+                  uv pip install -e .
```
EricGustin, line 35: Is this step necessary? The previous "Set up environment" step should already do this

.github/workflows/test-toolkits.yml
```diff
@@ -32,11 +35,13 @@ jobs:
 
   test-toolkits:
     needs: setup
-    runs-on: ubuntu-latest
+    name: test-toolkits (${{ matrix.toolkit }}, ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
EricGustin, line 42: IMO, and feel free to disagree, but I think doing an OS matrix for non-integration tests and especially for toolkit tests is overkill. We're looking at a ~13x price increase for this matrix for tests that are mostly, if not entirely, testing OS-agnostic things.

libs/arcade-cli/arcade_cli/new.py
```diff
@@ -277,9 +286,15 @@ def create_new_toolkit_minimal(output_directory: str, toolkit_name: str) -> None
...
-        console.print("  uv run server.py")
+        server_dir = toolkit_directory / toolkit_name / "src" / toolkit_name
+        console.print("\nNext steps:", style="bold")
+        console.print(f"  cd {server_dir}")
+        console.print("")
+        console.print("  Run with stdio transport (for MCP clients):", style="dim")
+        console.print("    uv run server.py")
+        console.print("")
+        console.print("  Run with HTTP transport (for development/testing):", style="dim")
+        console.print("    uv run server.py --transport http --port 8000")
```
EricGustin, line 297: I missed the first step to change my directory, and ended up getting an error. Not sure if this is just a me problem or if others would run into this. Maybe numbering the steps would help. 

Image

.github/workflows/cli-integration-no-auth.yml
```diff
@@ -0,0 +1,39 @@
...
+        branches: [main]
+    workflow_dispatch:
+
+jobs:
+    cli-integration-no-auth:
+        name: CLI Integration No-Auth (${{ matrix.os }}, Python ${{ matrix.python-version }})
+        runs-on: ${{ matrix.os }}
+        timeout-minutes: 35
+        env:
+            ARCADE_USAGE_TRACKING: "0"
```
EricGustin, line 16: is this needed in CI anymore since you added that function in conftest.py?

PR #797, opened 2026-03-19
feat: Add TelemetryPassbackMiddleware for serverExecutionTelemetry capability

savula15:
**Implements**: [SEP-2448: server execution telemetry] (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2448)

**Description:**

**The Observability Gap (The Problem)**

MCP clients propagate trace context to servers, but server-side execution remains a black box. The client sees a single tools/call or resources/read span; everything the server does (auth checks, policy evaluation, API calls, sub-tool invocations) is invisible. In cross-organization deployments, clients and servers use separate observability backends with no shared collector access, making traditional span export useless.

Screenshot 2026-03-23 at 3 43 21 PM

**Server Execution Telemetry (The Solution)**

Servers advertise serverExecutionTelemetry and return a curated slice of their execution spans directly in _meta.otel of the response. Clients ingest these verbatim OTLP spans into their own collector, stitching server-side execution into their distributed trace; no shared infrastructure required. The black box becomes transparent.

Screenshot 2026-03-23 at 3 43 44 PM

.
**Summary:**

Implement MCP serverExecutionTelemetry capability that enables cross-organization distributed tracing by returning server-side OpenTelemetry spans to clients inline via _meta.otel.traces.

  Server-side (middleware):
  - TelemetryPassbackMiddleware intercepts tools/call and resources/read
  - ContextVarSpanCollector isolates span collection per-request via ContextVar
  - Propagates traceparent from client request for distributed trace stitching
  - Serializes collected spans to verbatim OTLP JSON (resourceSpans format), directly POSTable to /v1/traces
  - Top-level span filtering by default; full span tree via detailed opt-in
  - Middleware advertises capabilities via get_capabilities() on the Middleware base class
  - Provisional API: FutureWarning emitted until SEP-2448 is ratified

  Client-side (reference agent):
  - LangChain ReAct agent connects to MCP server via streamable_http_client with OAuth 2.1
  - Detects serverExecutionTelemetry capability at initialization
  - Dynamically wraps discovered MCP tools with traceparent propagation and _meta.otel span request
  - Ingests returned server spans into Jaeger (OTLP JSON) and Galileo (OTLP protobuf)
  - Two-act demo: --no-passback (black box) vs default (full server-side visibility)

  Dependencies:
  - opentelemetry-api and opentelemetry-sdk added to arcade-mcp-server

  Bump arcade-mcp-server version to 1.18.0.

### DIFF
--- a/libs/arcade-mcp-server/arcade_mcp_server/middleware/telemetry.py
+++ b/libs/arcade-mcp-server/arcade_mcp_server/middleware/telemetry.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import logging
+import warnings
 from contextvars import ContextVar
 from typing import Any
 
@@ -167,6 +168,12 @@ def spans_to_otlp_json(
 class TelemetryPassbackMiddleware(Middleware):
     """MCP middleware implementing SEP serverExecutionTelemetry.
 
+    .. warning:: **Provisional API** — This middleware implements a draft SEP
+       (SEP-2448: server execution telemetry) that has not yet been finalized in
+       the MCP specification.  The capability schema, response payload shape,
+       and constructor signature may change in a future release once the SEP is
+       ratified.  Pin your ``arcade-mcp-server`` version if you need stability.
+
     Intercepts tools/call and resources/read to:
 
     1. Propagate traceparent from the client for distributed tracing.
@@ -176,6 +183,13 @@ class TelemetryPassbackMiddleware(Middleware):
     """
 
     def __init__(self, service_name: str, tracer_provider: TracerProvider) -> None:
+        warnings.warn(
+            "TelemetryPassbackMiddleware implements a draft version of SEP "
+            "(SEP-2448: server execution telemetry) that is not yet finalized. "
+            "The API may change in a future release.",
+            FutureWarning,
+            stacklevel=2,
+        )
         self._service_name = service_name
         self._tracer = tracer_provider.get_tracer(service_name)
         self._propagator = TraceContextTextMapPropagator()
--- a/libs/tests/arcade_mcp_server/test_telemetry_middleware.py
+++ b/libs/tests/arcade_mcp_server/test_telemetry_middleware.py
@@ -333,6 +333,7 @@ def _make_otel_context(*, request: bool = True, detailed: bool = False, tracepar
     return mcp_ctx
 
 
+@pytest.mark.filterwarnings("ignore::FutureWarning")
 class TestTelemetryPassbackMiddleware:
     """Test TelemetryPassbackMiddleware class."""
 


examples/mcp_servers/telemetry_passback/src/telemetry_passback/agent.py
```diff
@@ -0,0 +1,555 @@
...
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from pathlib import Path
+from typing import Any
+from urllib.parse import parse_qs, urlparse
+
+import httpx
+from dotenv import load_dotenv
+from langchain_core.tools import StructuredTool
+from langchain_openai import ChatOpenAI
+from langgraph.prebuilt import create_react_agent
```
EricGustin: Please update the import

```
LangGraphDeprecatedSinceV10: create_react_agent has been moved to
langchain.agents. Please update your import to `from langchain.agents import
create_agent`. Deprecated in LangGraph V1.0 to be removed in V2.0.
````

File: examples/mcp_servers/telemetry_passback/pyproject.toml [hunk 2]
```diff
@@ -0,0 +1,43 @@
+[project]
+name = "telemetry-passback"
+version = "0.1.0"
+description = "SEP-0000 reference implementation: cross-org distributed tracing via MCP serverExecutionTelemetry"
+requires-python = ">=3.11"
+dependencies = [
+    "arcade-mcp-server>=1.17.4,<2.0.0",
```
EricGustin: Since the telemetry middleware won't be published until 1.18.0
```suggestion
    "arcade-mcp-server>=1.18.0,<2.0.0",
```

libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -681,14 +688,18 @@ async def _handle_initialize(
...
             session.set_client_params(message.params)
 
+        caps_kwargs: dict[str, Any] = {
+            "tools": {"listChanged": True},
+            "logging": {},
+            "prompts": {"listChanged": True},
+            "resources": {"subscribe": True, "listChanged": True},
+        }
+        if self._extra_capabilities:
+            caps_kwargs.update(self._extra_capabilities)
```
EricGustin, line 696: The [MCP specification's schema ](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/0c439afa28daba012198c537b2e89bcddbd64b77/schema/2025-06-18/schema.ts#L265) defines a dedicated `experimental` namespace for non-standard capabilities. Perhaps the "extra capabilities" should be defined there instead. 

Thoughts/opinions?

libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -123,6 +123,7 @@ def __init__(
         auth_disabled: bool = False,
         arcade_api_key: str | None = None,
         arcade_api_url: str | None = None,
+        extra_capabilities: dict[str, Any] | None = None,
```
EricGustin: Is there a concrete use-case beyond middleware-declared capabilities? If not, then I think we should remove this from the `MCPServer`'s `__init__` to enforce co-location of extra capabilities to the server's middleware.

.gitignore
```diff
@@ -179,3 +179,7 @@ cython_debug/
 
 # Docs
 libs/arcade-mcp-server/site/*
+
+# exlcude OAuth tokens
```
EricGustin:
```suggestion
# exclude OAuth tokens
```

examples/mcp_servers/telemetry_passback/src/telemetry_passback/agent.py
savula15: Good catch. Upgraded from langchain 0.3.28 → 1.2.13 (latest stable) and changed the import to from langchain.agents import create_agent. All other dependencies bumped to latest stable as well.

examples/mcp_servers/telemetry_passback/pyproject.toml
savula15: Fixed

libs/arcade-mcp-server/arcade_mcp_server/server.py
savula15: Great point and we already considered that route. 

It's Intentional. We're positioning serverExecutionTelemetry as a standards-track capability, not an experimental one. The SEP argues that cross-org observability is fundamental to MCP; not niche or optional and placing it under experimental would signal lack of commitment to the ecosystem.

libs/arcade-mcp-server/arcade_mcp_server/server.py
savula15: Agreed. _init_middleware() already collects capabilities from middleware via get_capabilities(), so the constructor param was redundant. Removed it from MCPServer.__init__ and the example.

.gitignore
savula15: Fixed the typo.

savula15:
> Thanks for your contribution! I've left a few comments before this can be merged. Also, you can run `make check` from the root of the repository to see why the CI check is failing

Thanks @EricGustin and great observations. I've iterated and addressed all comments. Please have a look at the latest. 'make check' is also clear now.

EricGustin (commented):
@savula15, fair point about not placing `serverExecutionTelemetry` capability under `experimental`. Since the SEP may change by the time it gets into the specification, I still want to protect `arcade-mcp-server`'s API surface from breakage when the SEP evolves. Given that it's already an opt-in middleware, can you please add something like a `FutureWarning` + clear docstring/changelog markers on the class that convey the possibility of changes before being finalized.

savula15:
> @savula15, fair point about not placing `serverExecutionTelemetry` capability under `experimental`. Since the SEP may change by the time it gets into the specification, I still want to protect `arcade-mcp-server`'s API surface from breakage when the SEP evolves. Given that it's already an opt-in middleware, can you please add something like a `FutureWarning` + clear docstring/changelog markers on the class that convey the possibility of changes before being finalized.

Ah, Good idea @EricGustin. 

Added a FutureWarning that calls out the draft SEP status and the possibility of breaking changes before ratification.

savula15:
> Looks good to me! It looks like the pyproject.toml's version for arcade-mcp-server got overwritten with what's on main branch though

Great. Will clean up versions in future PR. For now, tested with local version and will update .toml later once new rev is released.

PR #800, opened 2026-03-21
Add background update check & notification for arcade CLI

EricGustin:
## Summary
- On every CLI command invocation (except `update`/`upgrade`/`mcp`), a detached subprocess checks PyPI for newer versions (throttled to once per 4 hours) and caches the result at `~/.arcade/update_cache.json`
- On the next invocation, if a newer version is known, a yellow one-liner notification is printed suggesting `arcade update`
- Respects `ARCADE_DISABLE_AUTOUPDATE=1` environment variable to opt out entirely

---

> [!NOTE]
> **Medium Risk**
> Adds a background PyPI version check that spawns detached subprocesses and may print update notifications on most CLI invocations; mistakes could impact CLI reliability or corrupt MCP stdio output (mitigated by explicit command exclusions).
> 
> **Overview**
> Adds `arcade update` (and hidden `arcade upgrade` alias) to self-upgrade the `arcade-mcp` CLI by detecting the original install method (`uv tool`, `pipx`, `uv pip`, or `pip`) and running the appropriate upgrade command.
> 
> Introduces a **throttled background update check** on most CLI invocations: a detached subprocess queries PyPI, writes `~/.arcade/update_cache.json`, and on subsequent runs prints a one-line notification when a newer version is cached; this is disabled via `ARCADE_DISABLE_AUTOUPDATE=1` and explicitly skipped for `update`/`upgrade`/`mcp` to avoid MCP stdio output corruption.
> 
> Bumps the package version to `1.13.0`, adds a `packaging` dependency, and includes comprehensive tests covering PyPI/yanked/prerelease handling, install-method detection, caching, and callback integration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2d9646ecc2211e8cfecd6e4901d14b1f5b7bb306. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libs/arcade-cli/arcade_cli/update.py
+++ b/libs/arcade-cli/arcade_cli/update.py
@@ -249,7 +249,15 @@ def detect_install_method() -> InstallMethod:
 _UPGRADE_COMMANDS: dict[InstallMethod, list[str]] = {
     InstallMethod.UV_TOOL: ["uv", "tool", "upgrade", PACKAGE_NAME],
     InstallMethod.PIPX: ["pipx", "upgrade", PACKAGE_NAME],
-    InstallMethod.UV_PIP: ["uv", "pip", "install", "--upgrade", PACKAGE_NAME],
+    InstallMethod.UV_PIP: [
+        "uv",
+        "pip",
+        "install",
+        "--upgrade",
+        "--python",
+        sys.executable,
+        PACKAGE_NAME,
+    ],
     InstallMethod.PIP: ["pip", "install", "--upgrade", PACKAGE_NAME],
 }
 
--- a/libs/tests/cli/test_update.py
+++ b/libs/tests/cli/test_update.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import sys
 import time
 from unittest.mock import MagicMock, patch
 
@@ -281,7 +282,7 @@ class TestUpdateCommand:
             assert "recommend" in result.output.lower()
             assert "uv tool" in result.output.lower()
             cmd = mock_run.call_args[0][0]
-            assert cmd == ["uv", "pip", "install", "--upgrade", PACKAGE_NAME]
+            assert cmd == ["uv", "pip", "install", "--upgrade", "--python", sys.executable, PACKAGE_NAME]
 
     def test_update_available_pip_shows_warning(self) -> None:
         with (


libs/arcade-cli/arcade_cli/main.py
```diff
@@ -891,6 +893,37 @@ def deploy(
         handle_cli_error("Failed to deploy server", e, debug)
 
 
+@cli.command(help="Check for and install CLI updates", rich_help_panel="Manage")
+def update(
+    debug: bool = typer.Option(False, "--debug", "-d", help="Show debug information"),
+) -> None:
+    """Check for updates to the Arcade CLI and install if available."""
+    from arcade_cli.update import run_update
```
EricGustin: top of file imports only pls. only do imports inside of function if absolutely necessary. if there would be a circular dependency, then first prefer refactoring instead of defaulting to importing inside the function

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -891,6 +893,37 @@ def deploy(
...
+        handle_cli_error("Failed to check for updates", e, debug)
+
+
+@cli.command(
+    name="upgrade",
+    help="Check for and install CLI updates (alias for update)",
+    rich_help_panel="Manage",
+    hidden=True,
+)
+def upgrade(
```
EricGustin, line 914: if it's possible to have upgrade an alias of update, then do that instead of duplicating essentially the same wrapper function twice

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,245 @@
...
+from packaging.version import Version
+
+from arcade_cli.console import console
+
+logger = logging.getLogger(__name__)
+
+PACKAGE_NAME = "arcade-mcp"
+PYPI_URL = f"https://pypi.org/pypi/{PACKAGE_NAME}/json"
+
+UPDATE_CACHE_PATH = os.path.join(ARCADE_CONFIG_PATH, "update_cache.json")
```
EricGustin, line 36: ensure cross-OS compatible

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,245 @@
...
+CHECK_INTERVAL_SECONDS = 4 * 60 * 60  # 4 hours
```
EricGustin, line 38: check interval for what?

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,245 @@
...
+    checked_at: float  # time.time()
+
+
+def read_update_cache(cache_path: str) -> UpdateCache | None:
+    """Read the update cache from disk. Returns None if missing or corrupt."""
+    try:
+        with open(cache_path) as f:
+            data = json.load(f)
+        return UpdateCache(
+            latest_version=data["latest_version"],
```
EricGustin: what if UpdateCahce's schema changes and this line doesnt?

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+    except Exception:
+        return None
+
+
+def write_update_cache(cache_path: str, cache: UpdateCache) -> None:
+    """Write the update cache to disk."""
+    os.makedirs(os.path.dirname(cache_path), exist_ok=True)
+    with open(cache_path, "w") as f:
+        json.dump({"latest_version": cache.latest_version, "checked_at": cache.checked_at}, f)
```
EricGustin: why not use the dedicated type that was defined for this?

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,245 @@
...
+        subprocess.Popen(
+            [
+                sys.executable,
+                "-c",
+                "from arcade_cli.update import _background_check; _background_check()",
+            ],
+            start_new_session=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
```
EricGustin: is this cross-OS compatible?

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,245 @@
...
+        return
+    cache = read_update_cache(UPDATE_CACHE_PATH)
+    if cache:
+        try:
+            current = metadata.version(PACKAGE_NAME)
+            if Version(cache.latest_version) > Version(current):
+                console.print(
+                    f"Update available: {current} → {cache.latest_version}  "
+                    f"Run `arcade update` to upgrade.",
+                    style="dim",
```
EricGustin: i'd prefer yellow

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,245 @@
...
+
+    if result.returncode == 0:
+        console.print(f"Successfully updated to {latest}!", style="bold green")
+    else:
+        console.print(
+            f"\nAuto-upgrade failed. You can try manually:\n"
+            f"  uv tool upgrade {PACKAGE_NAME}\n"
+            f"  uv pip install -U {PACKAGE_NAME}\n"
+            f"  pip install -U {PACKAGE_NAME}",
+            style="red",
```
EricGustin, line 318: prefer uv tool upgrade. Make it clear that they should not run all three of these commands. It currently is not clear and someone who isn't thinking would just copy and paste all 3

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,245 @@
...
+    """Determine whether a background PyPI check is needed."""
+    if os.environ.get("ARCADE_DISABLE_AUTOUPDATE") == "1":
+        return False
+    if cache is None:
+        return True
+    return (time.time() - cache.checked_at) > CHECK_INTERVAL_SECONDS
+
+
+def fork_background_check() -> None:
+    """Spawn a detached process that checks PyPI and writes the cache."""
```
EricGustin, line 81: this better not recommend an install if the lastest version is a release candidate (-rc1, rc1, etc.)

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,272 @@
...
+from packaging.version import Version
+
+from arcade_cli.console import console
+
+logger = logging.getLogger(__name__)
+
+PACKAGE_NAME = "arcade-mcp"
+PYPI_URL = f"https://pypi.org/pypi/{PACKAGE_NAME}/json"
+
+# Cross-OS: ARCADE_CONFIG_PATH uses os.path.join / os.path.expanduser
```
EricGustin:
```suggestion
```

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,271 @@
...
+def fork_background_check() -> None:
+    """Spawn a detached process that checks PyPI and writes the cache."""
+    cache = read_update_cache(UPDATE_CACHE_PATH)
+    if not should_check_for_update(cache):
+        return
+    cmd = [
+        sys.executable,
+        "-c",
+        "from arcade_cli.update import _background_check; _background_check()",
+    ]
```
EricGustin, line 89: we need some kind of test to ensure that this doesn't drift from what actually exists in code

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,271 @@
...
+            version: str = data["info"]["version"]
+            if Version(version).is_prerelease:
+                return None
+            return version
+    except Exception:
+        return None
+
+
+def detect_install_method() -> InstallMethod:
+    """Auto-detect how the user originally installed the CLI."""
```
EricGustin: just define the steps in the docstring instead

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,294 @@
...
+    InstallMethod.UV_TOOL: "uv tool",
+    InstallMethod.PIPX: "pipx",
+    InstallMethod.UV_PIP: "uv pip",
+    InstallMethod.PIP: "pip",
+}
+
+
+def _upgrade_command(method: InstallMethod) -> list[str]:
+    """Return the shell command to run for a given install method."""
+    return _UPGRADE_COMMANDS[method]
```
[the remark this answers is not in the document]
EricGustin: ignored

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,299 @@
...
+    cache = read_update_cache(UPDATE_CACHE_PATH)
+    if cache:
+        try:
+            cached_version = Version(cache.latest_version)
+            if cached_version > Version(metadata.version(PACKAGE_NAME)):
+                console.print(
+                    f"Update available: {metadata.version(PACKAGE_NAME)} → {cache.latest_version}  "
+                    f"Run `arcade update` to upgrade.",
+                    style="yellow",
+                )
```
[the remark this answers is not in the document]
EricGustin: The write path already guarentees. doing it on read too is redundant

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,298 @@
...
+                capture_output=True,
+                text=True,
+                timeout=10,
+            )
+            if result.returncode == 0 and _PACKAGE_RE.search(result.stdout):
+                return InstallMethod.PIPX
+        except Exception:
+            logger.debug("Failed to check pipx list", exc_info=True)
+
+    if shutil.which("uv"):
```
byrro, line 230: This fallback may mis-detect the install method. If `uv` is on `PATH` but the library was originally installed via plain pip, this returns `UV_PIP` and we run `uv pip install --upgrade arcade-mcp`, which may upgrade a different environment than the one backing the current `arcade` executable.

Perhaps we should only fallback to `UV_PIP` when there is stronger evidence. I don't know, maybe looking for an active venv context that maps to current interpreter?

libs/arcade-cli/arcade_cli/update.py
```diff
@@ -0,0 +1,298 @@
...
+from enum import Enum
+from importlib import metadata
+from urllib.request import urlopen
+
+from arcade_core.constants import ARCADE_CONFIG_PATH
+from arcade_core.subprocess_utils import (
+    build_windows_hidden_startupinfo,
+    get_windows_no_window_creationflags,
+)
+from packaging.version import Version
```
byrro, line 25: the `packaging` lib is not listed in `pyproject.dependencies`

this seems to work because some other downstream dependency rely on it and gets it installed in the environment, but it'd be a good practice to declare a direct dependency in the `arcade-mcp` pyproject, I think

EricGustin: Agreed. Added as an explicit dependency of the CLI.

libs/arcade-cli/arcade_cli/update.py
EricGustin: Addressed

PR #802, opened 2026-03-25
Add pctx Code Mode with eval suite with ProjectTracker example

pk8189:
Adds examples/evals/eval_pctx_code_mode.py, an evaluation suite for the pctx Code Mode MCP server example. Unlike the standard arcade-evals runner (which scores LLM tool calls directly), Code Mode requires a custom execution loop: the LLM generates TypeScript, pctx executes it in a sandbox, and tool invocations are extracted from the ExecutionTrace and scored against expected calls using arcade-evals critics and rubrics.

**What's included:**

- 7 eval cases covering the full ProjectTracker API: listing projects, getting/creating projects and sprints, task lifecycle, sprint metrics, search, and a multi-step chained workflow
- Soft rubric (fail_threshold=0.7) that scores via cost matrix rather than hard-failing on tool selection or call count, since Code Mode may inject extra helper calls
- `_build_system_prompt(declarations)` — system prompt is constructed at runtime from pctx.get_function_details(), injecting the actual TypeScript function signatures rather than hardcoding calling conventions; this keeps the prompt accurate as the tool surface evolves
- Supports OpenRouter or OpenAI via OPENROUTER_API_KEY / OPENAI_API_KEY, with model override via EVAL_MODEL

To run:

`pctx start
uv run --directory examples/mcp_servers/pctx_code_mode examples/evals/eval_pctx_code_mode.py`

### DIFF
--- a/examples/evals/eval_pctx_code_mode.py
+++ b/examples/evals/eval_pctx_code_mode.py
@@ -1,0 +1,401 @@
+"""
+ProjectTracker Code Mode evaluations.
+
+Uses pctx Code Mode: the LLM writes TypeScript executed in the pctx sandbox.
+Tool calls within the TypeScript are captured via ExecutionTrace and evaluated
+using arcade-evals critics and rubrics.
+
+Unlike the standard arcade-evals runner (which captures LLM tool calls directly),
+Code Mode requires a custom execution loop:
+  LLM generates TypeScript → pctx executes it → trace.events extracted as tool calls
+  → EvalCase.evaluate() scores against expected calls using critics and rubrics.
+
+Requires:
+  - pctx server running: pctx start
+  - OpenRouter (or OpenAI) key: OPENROUTER_API_KEY or OPENAI_API_KEY
+
+Run:
+    uv run --directory examples/mcp_servers/pctx_code_mode examples/evals/eval_pctx_code_mode.py
+"""
+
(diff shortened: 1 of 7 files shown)


JiwaniZakir (commented):
The `_PLACEHOLDER_SYSTEM = ""` pattern in `eval_pctx_code_mode.py` is fragile: every `EvalCase` is initialized with `system_message=_PLACEHOLDER_SYSTEM`, which passes the empty string *by value* at construction time, not a reference to the variable. This means reassigning `_PLACEHOLDER_SYSTEM` later at runtime won't update any case — the runtime loop must be explicitly mutating `case.system_message` on each `EvalCase` object directly. Since that mutation logic isn't visible in the diff, it's easy for a future refactor to silently break all evals by running them with an empty system prompt, producing meaningless scores with no obvious error. A more robust approach would be to build the cases lazily inside a factory function that accepts `declarations` as a parameter, eliminating the placeholder entirely. Additionally, the hardcoded IDs (`task_001`, `spr_001`, `user_eval`) in the eval cases implicitly depend on specific seed data being present in the pctx sandbox — this dependency should be documented in the module docstring or enforced with a setup/teardown fixture so the suite doesn't silently pass against stale or incorrect data.

EricGustin (approved):

PR #816, opened 2026-04-09
fix: TypedDict total=False output breaks validation

EricGustin:
When a tool’s output TypedDict uses total=False, MCP clients reject the response with:
```
MCP error -32602: Structured content does not match the tool's output schema
```
Note that the bug also exists for the Engine transport (/worker/tools/execute), but since the engine doesn't validate the output schema, the bug never surfaced. This PR addresses the problem holistically (MCP and Engine) in preparation for a future where the Engine transport validates output schemas.

Two bugs combined to cause this:
1. Schema: The outputSchema had no required array and declared all fields as strict types (e.g. "type": "string"), making every field look mandatory and non-null.
2. Serialization: model_dump() on TypedDict-derived Pydantic models emitted None for absent optional fields. A tool returning {"name": "hello"} produced {"name": "hello", "optional_field": null} which is a value the schema forbids.

---

> [!NOTE]
> **Medium Risk**
> Adjusts core schema generation and MCP JSON Schema conversion for TypedDicts, affecting how tool input/output contracts are emitted and validated across clients; mistakes could break compatibility or validation behavior.
> 
> **Overview**
> Fixes MCP/engine validation failures for `TypedDict(total=False)` outputs by ensuring absent optional keys are **omitted from serialized output** and that emitted schemas correctly describe **required vs optional** keys.
> 
> `arcade-core` now tracks `required_keys`/`inner_required_keys` and per-field `nullable` in `ValueSchema`, derives required sets from TypedDict `__required_keys__`, and unwraps `Optional[T]` to support optional nested TypedDicts; TypedDict-derived Pydantic models now `model_dump(exclude_unset=True)` to avoid leaking missing fields as `null`.
> 
> `arcade-mcp-server` JSON Schema conversion now emits `required` arrays (including for arrays of objects), supports `nullable` by generating `type: [<type>, "null"]` (and `enum` including `None`), and treats nullable top-level objects as valid unwrapped output schemas. Adds focused unit/end-to-end tests plus an expanded example server demonstrating total-false, mixed required/optional, nullable, and optional-nested TypedDict outputs, and bumps package versions/dependencies accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53fe8365f613053599130520b75f30b614b465ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/mcp_servers/tools_with_output_schema/pyproject.toml
+++ b/examples/mcp_servers/tools_with_output_schema/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "MCP Server created with Arcade.dev"
 requires-python = ">=3.10"
 dependencies = [
-    "arcade-mcp-server>=1.19.0,<2.0.0",
+    "arcade-mcp-server>=1.19.1,<2.0.0",
 ]
 
 [project.optional-dependencies]
--- a/examples/mcp_servers/tools_with_output_schema/src/tools_with_output_schema/server.py
+++ b/examples/mcp_servers/tools_with_output_schema/src/tools_with_output_schema/server.py
@@ -7,17 +7,21 @@ fully-expanded JSON Schema output schemas, so MCP clients can validate and
 display tool results without guessing the shape of the data.
 
 Tools in this server progress from simple to complex:
-  - calculate_statistics  — flat TypedDict (all scalar fields)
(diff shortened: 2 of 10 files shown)


byrro (approved):
LGTM

PR #818, opened 2026-04-11
Bootstrap Arcade login from arcade-mcp-server (stdio); harden token refresh

EricGustin:
## What this changes

This branch closes two gaps in stdio `arcade-mcp-server` auth:

1. **Token refresh in long-running sessions** could lose the configured user email (the original bug this PR was opened for).
2. **Bootstrap login when no token exists at all** — `arcade-mcp-server` previously had a runtime data dependency on `~/.arcade/credentials.yaml`, a file only written by `arcade login` from `arcade-cli`. A user who installed only `arcade-mcp-server` had no way to authenticate to Arcade Cloud; every auth-requiring tool failed with no recovery path. This gap is **stdio-only** — HTTP transport uses `ResourceServerAuth` (OAuth 2.1 bearer-token validation) and is unaffected.

After this PR, stdio `arcade-mcp-server` is self-sufficient: it can complete an Arcade login on its own, surface the auth URL through MCP elicitation when the client supports it, and fall back to surfacing the URL in the tool error response when it doesn't. In both cases the server actively listens for the OAuth callback so login completes without `arcade-cli` installed.

## Token-refresh hardening (commits 14e6ca64..d3cd9778)

- Decouple token state from user-ID resolution: split `_load_config_values` into `_load_access_token` and `_load_config_user_id` so an expired token no longer makes the server lose the configured email.
- `asyncio.Lock` to serialize concurrent refresh attempts and prevent racing on the same refresh token.
- `asyncio.to_thread` so the synchronous credentials-file load doesn't block the event loop.
- Guard against `arc_*` service keys being overwritten by a CLI OAuth token in the refresh path (service keys never expire).
- Lazy refresh hook in `_check_authorization` updates `self.arcade.api_key` in place without rebuilding the client.

### Phase A: `arcade_core.oauth_login`

Move the OAuth protocol primitives into `arcade-core` so both `arcade-cli` and `arcade-mcp-server` can drive a login without duplicating logic. Implementation is **pure stdlib + httpx** — no `authlib`, no `Jinja2`, no new core deps.

- `generate_authorization_url` (PKCE S256 via stdlib `secrets`/`hashlib`)
- `async exchange_code_for_tokens`, `async fetch_whoami` (`httpx.AsyncClient`)
- `validate_whoami_org_project` (raises `NoOrgsError`/`NoProjectsError`)
- `async save_credentials_from_whoami` (YAML write via `asyncio.to_thread`; invalidates the `get_config` lru_cache)
- `OAuthCallbackServer` — owns its thread; bind errors caught (no traceback to stderr); `handle_error` routes through a stdlib logger with `NullHandler` and `propagate=False` so stdio paths stay silent; `BrokenPipeError` suppressed; unknown paths return 404 without advancing the listener.

### Phase B: `arcade_cli.authn` delegates to core

The CLI's OAuth flow now imports from `arcade_core.oauth_login` instead of carrying its own copy. CLI-specific UX (`rich.console`, `webbrowser`, status callbacks, `ARCADE_LOGIN_TIMEOUT_SECONDS`) is preserved.

- Drop `authlib` + `types-Authlib` deps (no longer imported anywhere in `arcade-cli`).
- Delete unused Jinja2 callback templates (core uses static HTML). `Jinja2` stays — `arcade_cli/new.py` still uses it for project scaffolding.
- Public CLI symbols stay importable; moved primitives are re-exported from `arcade_cli.authn`.

### Phase C: `arcade_mcp_server.auth_bootstrap`

`bootstrap_login` orchestrates an Arcade OAuth login from inside the MCP server. A per-server `AttemptSlot` owns the in-flight `LoginAttempt` so concurrent missing-token tool calls share one URL, one listener, one `fetch_cli_config`.

Behavior matrix:

| slot state | elicit provided | bootstrap_login does |
|---|---|---|
| empty / expired / done | yes | build URL, start listener, store attempt; send elicit; race elicit-result vs completion; on callback resolve via exchange/whoami/validate/save → COMPLETED or FAILED |
| empty / expired / done | no | build URL, start listener, store attempt; spawn background task; return URL_FOR_FALLBACK synchronously |
| populated, alive | yes | send fresh elicit (so joiner sees the URL too); await same completion; return its result |
| populated, alive | no | return URL_FOR_FALLBACK with the existing URL; never bind a second listener |

Elicit `accept` keeps waiting for the IdP callback; `decline`/`cancel`/`timeout`/exception map to `FAILED`. `AttemptSlot.shutdown` cancels the background task and shuts down the listener (idempotent).

### Phase D: wire into `MCPServer`

- **Capability rewrite**: `session.check_client_capability` now uses presence semantics (`model_fields_set` + `__pydantic_extra__` + `is not None`). The old helper used truthiness, so `ClientCapabilities(elicitation={})` (empty dict, falsy) short-circuited to `True` for every initialized client. Also extends to sampling, which the old helper ignored entirely.
- **Lazy startup**: `MCPServer.__init__` no longer does network/disk I/O. With no explicit `arcade_api_key` (constructor arg or `ARCADE_API_KEY` env), `self.arcade` stays `None` at startup. With an explicit key, the client is built immediately (just a string, no I/O).
- **`_login_slot`** stored on the server; shut down idempotently in `_stop`.
- **`_load_config_user_id` / `_load_org_project_context`** call `Config.load_from_file()` directly (not the cached singleton), so post-bootstrap reads see fresh data without depending on cache invalidation timing. `_load_access_token` still uses `get_valid_access_token` (preserves refresh path).
- **`ArcadeSettings.coordinator_url`** (default `https://cloud.arcade.dev`, env `ARCADE_COORDINATOR_URL`) — distinct from `api_url` which is the Engine. Bootstrap reads only `coordinator_url`.
- **`_check_tool_requirements` wiring** for the auth-required branch:
  - Service-key short-circuit preserved.
  - `requires_secrets`-only path untouched.
  - HTTP transport with `self.arcade is None`: existing missing-key error response.
  - stdio + auth-required + `self.arcade is None`:
    - **Retry-first**: load token + build client (creds may have appeared from a prior fallback bootstrap).
    - Otherwise probe elicitation capability via the corrected helper; call `bootstrap_login(elicit=session.elicit if supported else None)`.
    - `COMPLETED` → rebuild client; refresh `tool_context.user_id` from the freshly-loaded config email **before** `_check_authorization`.
    - `URL_FOR_FALLBACK` → error response carrying only `{message, llm_instructions, authorization_url}`. No `developer_message` / `reason` / `detail` leak (logged server-side only).
    - `FAILED` → clean error with humanized reason; reason machine-readable in `llm_instructions`; detail logged but never returned.

## End-to-end UX

**Fresh venv with only `arcade-mcp-server`** (no `arcade-cli`), no `~/.arcade/credentials.yaml`:

- Elicitation-capable client (e.g. Claude Desktop): call an auth-required tool → elicitation prompt with login URL appears → user clicks → callback lands → credentials written → tool result returns in the same call.
- Non-elicitation client: call an auth-required tool → response includes login URL in user-visible message → user logs in in browser → call again → second call hits the retry-first path and succeeds without starting a second listener.

## Versioning

- `arcade-core`: 4.7.0 → 4.8.0 (new `oauth_login` public surface)
- `arcade-mcp-server`: 1.21.1 → 1.22.0; `arcade-core` floor bumped to `>=4.8.0`
- `arcade-mcp` (root, the CLI): 1.14.1 → 1.15.0; `arcade-core` floor bumped to `>=4.8.0`; `authlib` + `types-Authlib` removed

## Tests

Strict TDD throughout. Stdio safety asserted via `capfd` silence (file-descriptor level — catches anything the BaseHTTPServer or threads might leak past `capsys`).

| Suite | Tests |
|---|---|
| `libs/tests/core/test_oauth_login.py` | 17 — happy path, CSRF state mismatch, unknown path 404 (non-fatal), timeout, port-in-use, broken pipe, `handle_error` silence, full-flow capfd silence, save chmod 600 + cache invalidation |
| `libs/tests/arcade_mcp_server/test_auth_bootstrap.py` | 22 — both paths (elicit + fallback), `accept`-keeps-waiting, decline/cancel/timeout, joiner UX (elicit + non-elicit + mixed), expired-attempt replacement, port-in-use, every downstream failure (cli_config, token exchange, whoami, no_orgs, no_projects, save), capfd silence, slot shutdown |
| `libs/tests/arcade_mcp_server/test_session.py` | 4 capability-rewrite tests; existing wrong assertions updated |
| `libs/tests/arcade_mcp_server/test_settings.py` | `coordinator_url` env + default |
| `libs/tests/arcade_mcp_server/test_server.py` | 16 in `TestNoTokenBootstrap` — no startup I/O, slot shutdown on `_stop`, retry-first, elicit-capable success, fallback URL, no-elicit-when-uncapable, clean failure surfacing without dev-message leak, in-call user_id propagation, concurrent calls share one slot, end-to-end fallback simulation, capfd silence, transport gating, secrets-only untouched, service-key short-circuit. Plus the existing `TestLoadAccessToken` / `TestLoadConfigUserId` / `TestInitArcadeClient` / `TestSelectUserId` / `TestLazyTokenRefresh` / `TestTokenEmailDecoupling` / `TestToolErrorResponse` / `TestLogToolCallError` classes from earlier in the branch (4 minor updates for the lazy-init / direct `Config.load_from_file()` change). |

**Full suite**: 2887 passed, 1 skipped. `make check` clean (ruff + ruff-format + mypy across all 6 libraries).


0xbrainkid:
The decoupled token vs. email loading is the right fix — tying identity loss to token expiry was the root cause, and separating the two concerns makes the auth state machine much cleaner.

One dimension worth noting in the fix: **silent identity loss is a trust boundary violation in agentic contexts.** When the server falls back from `user@example.com` to a session ID because of an expired token, any downstream tool calls or agent actions during that window are attributed to the wrong principal — or to no traceable principal at all. For audit trails and behavioral trust registries, an unattributable action is worse than a failed action.

Two additions that would make the fix more robust for long-running agentic sessions:

**1. Emit an identity-loss event before falling back.** Before switching to session ID mode, emit a structured event that upstream agents or orchestrators can intercept — giving them the opportunity to trigger a re-auth flow rather than silently degrading. An unhandled identity loss in a tool pipeline may compound downstream.

**2. Behavioral trust continuity.** If the Arcade server tracks behavioral trust scores per user identity, a mid-session identity switch (even temporary) should not reset or corrupt the behavioral history. The lazy refresh path you added is the right mechanism — ensuring the identity is re-established before any trust-relevant actions occur in the refreshed session.

The serialized lazy refresh with no concurrent refresh races is a solid implementation of this. The behavioral trust angle is an additional reason why getting the refresh right matters beyond just credential correctness.

PR #826, opened 2026-04-19
feat(arcade-core): opt-in debug leak flags for toolkit authors

jottakka:
## Summary

Adds two strictly opt-in env vars that let toolkit developers see `developer_message` / `stacktrace` content *in* the agent-facing error message while debugging. Off by default; activation requires a specific acknowledgement string, not a boolean — `true`/`1` is explicitly rejected with a warning log.

- `ARCADE_UNSAFE_DEBUG_LEAK_DEVELOPER_MESSAGE_TO_AGENT`
- `ARCADE_UNSAFE_DEBUG_LEAK_STACKTRACE_TO_AGENT`
- Magic ack: `yes-i-accept-leaking-internals-to-the-agent`

Everything goes through a single funnel — `ToolOutputFactory.fail` / `fail_retry` in `arcade_core/output.py` — so the behavior covers both the MCP server path and the Arcade Worker path with no call-site changes. A loud `logger.warning` fires once per process on activation, and a big header comment in `output.py` tells future maintainers not to add more flags of this shape (debug info belongs in `logger.debug`, not in a field that gets shipped to the model and often to end users).

Bumps `arcade-core` 4.6.2 → 4.7.0. Non-breaking, additive.

## Why

Today the project does a lot of work to keep `developer_message` and `stacktrace` off the agent's context. That's the right default, but it makes iterating on a new toolkit painful — you end up adding temporary logging or rebuilds just to see what blew up. This gives toolkit authors a safe, ugly, loud-on-activation escape hatch.

## Safety design

- Two separate flags so you only leak what you need.
- Magic string (not a boolean) activates the flag. Boolean-style values are rejected and log a pointer to `output.py`.
- First activation logs a `WARNING` identifying the flag and the risk.
- Flags documented only in `CLAUDE.md`, not in the public README.
- Top-of-file banner in `output.py` explicitly tells maintainers not to add more flags of this shape.

## Test plan

- [x] Existing test suite passes (1154 tests — `libs/tests/{core,tool,arcade_mcp_server}`).
- [x] End-to-end smoke test against the built `arcade_core-4.7.0` wheel, driven through `ToolExecutor.run` (same path toolkits hit). Covered cases:
  - flags off → message unchanged
  - `ARCADE_UNSAFE_..._DEVELOPER_MESSAGE_TO_AGENT=true` → flag rejected, warning logged, message unchanged
  - `ARCADE_UNSAFE_..._DEVELOPER_MESSAGE_TO_AGENT=<magic>` → `[DEBUG] developer_message: ...` appended
  - both flags with magic, `ToolRuntimeError` path → developer_message appended (stacktrace absent because `ToolRuntimeError.stacktrace()` returned `None`, which is existing behavior)
  - stacktrace flag with magic, generic `Exception` path → full `traceback.format_exc()` appended, activation `WARNING` visible

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Adds an opt-in path to include `developer_message` and stacktraces in agent-facing MCP error messages, which could leak sensitive data if misconfigured; safeguards (magic ack string + CI/pre-commit guard) reduce but don’t eliminate risk.
> 
> **Overview**
> Adds `arcade_mcp_server/_debug_exposure.py` with two env-gated debug flags that, only when set to a specific acknowledgement string, append `developer_message` and/or `stacktrace` into the agent-visible MCP tool error `message` (and logs one-shot warnings on rejection/activation).
> 
> Wires this into the MCP error path in `MCPServer._handle_call_tool`, documents the flags in `CLAUDE.md`, bumps `arcade-mcp-server` to `1.21.0`, and adds unit + integration tests plus a pre-commit hook and GitHub Actions workflow (`scripts/check_debug_leak_flags_off.py`) to ensure the magic ack string can’t be committed outside a small allowlist.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30e242c454128ec7cc62e169c2afd116be735cb5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jottakka:
Good catch from Bugbot — fixed in 321fdbac.

**Root cause:** `_warned_flags` was one shared set used by both the rejection-warning path (truthy but non-magic value) and the activation-warning path (magic value). Once a flag had been "warned" for rejection, the critical `ENABLED … DO NOT USE IN PRODUCTION` warning was silently suppressed if the developer then corrected the value to the magic string within the same process.

**Fix:** split the one-shot state into two independent sets — `_warned_rejected` and `_warned_activated`. Each category is still one-shot per flag per process, but they no longer starve each other.

**Regression coverage:** `test_fail_rejection_does_not_suppress_later_activation_warning` reproduces the exact sequence Bugbot described and asserts both warnings fire.

libs/arcade-core/arcade_core/output.py
```diff
@@ -8,6 +34,65 @@
...
 
+_logger = logging.getLogger(__name__)
+
+# Acknowledgement string a developer must set as the env value. Picked to be
+# impossible to set by mistake — no sane config management or CI will ever
+# emit this string.
+_DEBUG_LEAK_MAGIC = "yes-i-accept-leaking-internals-to-the-agent"
+
+_ENV_LEAK_DEVELOPER_MESSAGE = "ARCADE_UNSAFE_DEBUG_LEAK_DEVELOPER_MESSAGE_TO_AGENT"
+_ENV_LEAK_STACKTRACE = "ARCADE_UNSAFE_DEBUG_LEAK_STACKTRACE_TO_AGENT"
```
EricGustin: The `_TO_AGENT` framing assumes a lot about the caller of the tool. What do you think about that following env var names so that they describe the actual mutation?

1. `ARCADE_DEBUG_EXPOSE_DEVELOPER_MESSAGE_IN_TOOL_ERROR_RESPONSES`
2. `ARCADE_DEBUG_EXPOSE_STACKTRACE_IN_TOOL_ERROR_RESPONSES`

libs/arcade-core/arcade_core/output.py
```diff
@@ -1,3 +1,29 @@
...
+#   * require a specific magic acknowledgement string — not `true`, not `1` —
+#     to avoid being flipped on by generic "enable all debug flags" scripts,
+#   * log a loud WARNING the first time they are read as enabled,
+#   * are documented ONLY in CLAUDE.md, not in public READMEs.
+#
+# If you find yourself wanting to add a third such flag: don't. Put the debug
+# info in logs (`logger.debug`) or in a dedicated diagnostic endpoint instead.
+# Anything appended to `ToolCallError.message` WILL be shipped to the model
+# and, for most tools, also shown to an end user in some UI.
+#
```
EricGustin: nit; by the time someone is editing this file, they've already seen the flag name, hit the runtime warning log, tripped the CI guard, and their agent has read the CLAUDE.md section. The top of file banner feels like a fifth copy of the same signal. Would a 2-3 line version of this immediately above the actual declaration of the constants (instead of top-of-file)  land the same intent? Happy to leave it if you feel the banner is doing work im missing

.github/workflows/check-debug-leak-flags.yml
```diff
@@ -0,0 +1,30 @@
+name: Debug Leak Flag Guard
```
EricGustin, line 1: nice

EricGustin (changes requested):
This doesn't seem to belong in `arcade-core`. `ToolOutputFactory.fail` is shared by every consumer (MCP, the worker, and evals). Only MCP has the problem that this PR wants to solve (clients drop `developer_message` and only surface `message` which makes server development difficult). The worker path already returns the full `ToolCallOutput` to the Engine, which surfaces `developer_message` natively, so augmenting `message` for this path is redundant. 

I'd suggest moving into an `arcade-mcp-server` error handling middleware so that transform lives at the MCP boundary. This way we can keep `arcade-core` as a pure data transform.

libs/arcade-core/pyproject.toml
EricGustin, line 3: arcade-core isn't touched in this PR, but version is being bumped

EricGustin (approved):
approving with small comment

PR #860, opened 2026-06-03
fix(arcade-core,arcade-mcp-server): Nested-object schemas across all tool-schema converters

EricGustin:
## Summary

Arcade renders a tool's `ValueSchema` into JSON Schema in three independent places, and they had drifted into the same class of bug: nested object subschemas (especially `list[<TypedDict>]` items) were emitted incompletely, producing OpenAI-strict-invalid schemas that the OpenAI Responses API rejects with a hard 400. Because OpenAI rejects the whole tool list when one schema is invalid, every eval/tool batch containing that tool fails.

This PR fixes all three converter paths and the shared root cause they all read from, so correct schemas come from one corrected source rather than per-converter patches.

Resolves: https://linear.app/arcadedev/issue/TOO-990/arcade-core-tool-schema-converters-drop-nested-object-fields-for

## What was wrong (three layers)

1. **`arcade_core/converters/openai.py`** (OpenAI strict) and **`anthropic.py`** (standard) built array `items` from the inner scalar type alone, emitting bare `{"type": "object"}` and dropping object fields. OpenAI strict also requires `additionalProperties: false` + every key in `required` on each object.
2. **`arcade_mcp_server/convert.py`** (MCP `inputSchema`) preserved nested properties/required but never set `additionalProperties: false` on nested objects or array-of-object items (only the top-level wrapper did). This is the path consumers monkey-patch when feeding MCP tool schemas to OpenAI strict mode.
3. **`arcade_core/catalog.py::extract_properties`** extracted `required_keys`/`nullable` for **TypedDict** params but not for **Pydantic** models, so every converter received imprecise data for Pydantic object params (no `required`, no nullability).

## Changes

- **OpenAI + Anthropic converters** (commit 1): expand every object recursively at any depth (top-level param, nested object, array-of-object items). OpenAI strict closes every object (`additionalProperties: false`), lists all keys in `required`, and renders optionals as nullable unions; Anthropic uses standard JSON Schema (only required keys).
- **MCP converter** (commit 2): set `additionalProperties: false` on objects with a known shape, at every depth. Freeform `dict` params (no properties) stay open, since closing them would reject all keys.
- **Root cause** (commit 2): `extract_properties` now populates `required_keys` for Pydantic models via `FieldInfo.is_required()` and marks `nullable` on `Optional` fields, matching the existing TypedDict handling. All three converters get accurate metadata from one place.

---

> [!NOTE]
> **Medium Risk**
> Changes shared schema extraction and all three tool-schema emitters, which can alter API/MCP validation behavior for nested Pydantic/TypedDict params; mitigated by extensive new tests but still user-visible for existing tools.
> 
> **Overview**
> Fixes **incomplete nested object JSON Schema** for tool inputs so OpenAI strict, Anthropic, and MCP paths stay aligned and OpenAI no longer 400s on invalid tool lists.
> 
> **`arcade_core/catalog`:** Pydantic params now get **`required_keys`** and **`nullable`** like TypedDict; known shapes use **`({}, [])`** instead of collapsing to `None`, so empty/zero-field models differ from freeform **`dict`**. List and **`ValueSchema`** conversion propagate nested shapes when **`properties` / `inner_properties` is not `None`**.
> 
> **OpenAI & Anthropic converters:** Shared **`_build_object_schema`** recursively expands **`json`** and **`list[object]`** items (no more bare **`{"type": "object"}`**). OpenAI strict closes objects, lists all keys in **`required`**, and uses **`null`** unions (including enums); Anthropic lists only real required keys and unions **`null`** on nullable nested fields.
> 
> **MCP `convert`:** Known-shape objects and array-of-object items get **`additionalProperties: false`** at every depth; unknown **`dict`** shapes stay open.
> 
> Bumps **`arcade-core` 4.7.3** and **`arcade-mcp-server` 1.22.2** with broad converter/catalog tests plus a cross-dialect consistency test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46dee1d08e0421e1fe06e375ba300b4a50d57594. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-core/arcade_core/catalog.py
+++ b/libs/arcade-core/arcade_core/catalog.py
@@ -867,6 +867,7 @@ def extract_properties(
 
     # Handle Pydantic BaseModel
     if isinstance(type_to_check, type) and issubclass(type_to_check, BaseModel):
+        pydantic_required_keys: list[str] = []
         for field_name, field_info in type_to_check.model_fields.items():
             # Get the field type
             field_type = field_info.annotation
@@ -874,16 +875,23 @@ def extract_properties(
                 continue
 
             # Handle Optional types (Union[T, None])
-            if is_strict_optional(field_type):
+            is_nullable = is_strict_optional(field_type)
+            if is_nullable:
                 # Extract the non-None type from Optional
                 field_type = next(arg for arg in get_args(field_type) if arg is not type(None))
 
             # Get wire type info recursively
             # field_type cannot be None here due to the check above
             wire_info = get_wire_type_info(field_type)
+            if is_nullable:
+                wire_info.nullable = True
             properties[field_name] = wire_info
 
-        return (properties or None, None)
+            # A Pydantic field is required when it has no default (default or factory).
+            if field_info.is_required():
+                pydantic_required_keys.append(field_name)
+
+        return (properties or None, sorted(pydantic_required_keys) or None)
 
     # Handle TypedDict
     elif is_typeddict(type_to_check):
--- a/libs/arcade-core/arcade_core/catalog.py
+++ b/libs/arcade-core/arcade_core/catalog.py
@@ -860,8 +860,11 @@ def extract_properties(
     """
     Extract properties from TypedDict, Pydantic models, or other structured types.
 
-    Returns (properties, required_keys). required_keys is a sorted list of required
-    property names for TypedDict types, or None for other types.
+    Returns (properties, required_keys). For a known structured type (TypedDict or
+    Pydantic model) required_keys is a sorted list of required property names, which is
+    empty when every field is optional. For a type with no known shape (e.g. a freeform
+    ``dict``) required_keys is None, signaling that requiredness is unknown rather than
+    known-empty.
     """
     properties = {}
 
@@ -891,7 +894,7 @@ def extract_properties(
             if field_info.is_required():
                 pydantic_required_keys.append(field_name)
 
-        return (properties or None, sorted(pydantic_required_keys) or None)
+        return (properties or None, sorted(pydantic_required_keys))
 
     # Handle TypedDict
     elif is_typeddict(type_to_check):
@@ -917,8 +920,7 @@ def extract_properties(
 
             properties[field_name] = wire_info
 
-        req = sorted(getattr(type_to_check, "__required_keys__", frozenset()))
-        required_keys = req or None  # normalize empty → None
+        required_keys = sorted(getattr(type_to_check, "__required_keys__", frozenset()))
         return (properties or None, required_keys)
 
     # Handle regular dict with type annotations (e.g., dict[str, Any])
--- a/libs/arcade-core/arcade_core/catalog.py
+++ b/libs/arcade-core/arcade_core/catalog.py
@@ -775,12 +775,13 @@ def get_wire_type_info(_type: type) -> WireTypeInfo:
         inner_info = get_wire_type_info(inner_type)
         inner_wire_type = cast(InnerWireType, inner_info.wire_type)
 
-        # If inner type has properties (it's a complex object), propagate them
-        if inner_info.properties:
+        # If inner type has a known object shape (possibly empty), propagate it. A known-empty
+        # shape ({}) is distinct from None (unknown shape) so converters can close it.
+        if inner_info.properties is not None:
             inner_properties = inner_info.properties
             inner_required_keys = inner_info.required_keys
         # If inner type is array (nested arrays), propagate inner_properties
-        elif inner_info.inner_properties:
+        elif inner_info.inner_properties is not None:
             inner_properties = inner_info.inner_properties
             inner_required_keys = inner_info.inner_required_keys
     else:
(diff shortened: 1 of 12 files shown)


jottakka:
## Code review — verified against the checked-out branch

Reviewed across correctness / API-contract / maintainability / testing / adversarial lenses, then **reproduced every claim live** (built a venv from `libs/arcade-core`, `arcade-tdk`, `arcade-mcp-server` and ran the real `@tool` → catalog → converter path). OpenAI strict-mode rules were also re-confirmed against OpenAI's docs: at **every** object level you must set `additionalProperties: false`, list **all** properties in `required`, and express optionals as `["type", "null"]` unions (no omission) — missing any of these returns a hard 400.

**The core fix is correct and well-tested.** `to_openai` on a Pydantic model with optional/defaulted fields now emits a fully strict-valid schema. The notes below are one introduced inaccuracy plus a residual gap that still affects the bug this PR targets.

### Evidence (same `Recipient` model, two converters)

```python
class Recipient(BaseModel):
    email: str
    name: Optional[str] = None   # optional + nullable
    org: str = "acme"            # defaulted, NOT Optional
```

`to_openai(...)` — ✅ strict-valid:
```json
{
  "type": "object",
  "properties": {
    "email": {"type": "string"},
    "name":  {"type": ["string", "null"]},
    "org":   {"type": ["string", "null"]}
  },
  "required": ["email", "name", "org"],
  "additionalProperties": false
}
```

`create_mcp_tool(...).inputSchema` — ❌ strict-INVALID:
```json
{
  "type": "object",
  "properties": {
    "email": {"type": "string"},
    "name":  {"type": ["string", "null"]},
    "org":   {"type": "string"}
  },
  "additionalProperties": false,
  "required": ["email"]
}
```

`name` and `org` are absent from `required`, and `org` is not nullable — so this schema is rejected under OpenAI strict mode.

---

### Must fix

1. **`convert.py` (and `test_convert.py`) claim the MCP schema is "valid under OpenAI strict mode" — it isn't.** The comment at `convert.py:284-286` and the `test_input_schema_nested_object_is_closed` docstring both state that adding `additionalProperties: false` makes the schema "valid when an MCP host forwards it to OpenAI strict mode." As the evidence above shows, that is false for any object with an optional/defaulted field: MCP keeps standard `required` (only actually-required keys) and leaves non-`Optional` defaulted fields non-nullable. Strict mode requires *all* keys in `required` + nullable optionals. **Fix:** reword the claim (MCP emits standard JSON Schema with closed known objects; it is **not** directly forwardable to strict as-is), or make the MCP path mirror the OpenAI strict dialect if strict-forwarding is an actual requirement.

2. **Freeform `dict` / `list[dict]` params still 400 under strict.** `to_openai` hardcodes `strict: True`, but a freeform `dict` emits `{"type": "object"}` (no `additionalProperties`) and `list[dict]` emits `{"type": "array", "items": {"type": "object"}}` — both strict-invalid. This is pre-existing, but it's the exact failure mode this PR sets out to eliminate, and (unlike `convert.py`) the OpenAI converter doesn't acknowledge it. A clean schema-level fix isn't possible (strict can't express an open object), so this needs a **policy decision** (disable strict for such tools / reject at build time / emit a closed stand-in), an in-code acknowledgement in `openai.py`, and an `xfail`/regression test so the residual 400 is pinned rather than silent.

### Nice to fix

- **De-duplicate the null-union logic.** `_add_null_to_type` is byte-identical in `openai.py` and `anthropic.py`, and `convert.py::_apply_nullable` performs the same two mutations — three copies with no dialect difference, which re-seeds the "three places drifted into the same bug" pattern this PR fixes. A shared `converters/utils.py` helper collapses it.
- **Document the `None`-vs-`{}` contract on the data model.** The known-empty `({}, [])` vs unknown `(None, None)` distinction is now load-bearing across 4 files but lives only in scattered comments. Put it on `ValueSchema.properties` / `required_keys` / `WireTypeInfo` so the next editor doesn't reintroduce an `if properties:` truthiness check. (`convert.py` already mixes `is not None` for objects with a truthy guard for `required_keys`.)
- **Close the cross-converter guard's blind spot.** `_assert_openai_strict` in `test_schema_consistency.py` only asserts when `"properties" in schema`, so a future regression to a bare `{"type": "object"}` would be silently skipped by the recursive sweep. Add coverage for nested-freeform-dict and `list[dict]` under strict, and for an MCP partial-required object forwarded to strict.

### Nit

- Nested all-optional objects emit `required: []` while the top-level builder deletes an empty `required` — cosmetic, both strict-valid.
- Public exported TypedDict `items` narrowed `dict[str, Any]` → recursive type: fine, worth a changelog note for external static consumers.
- `arcade-core` 4.7.2 → 4.7.3 patch bump for an output-behavior change (defensible as a bug fix); cross-converter test lives under `arcade_mcp_server/` though it spans core converters.

jottakka (commented):
My agent left some comments, there are some things it found that were there before the PR but i beleive that it is worth to check.

jottakka (approved):

PR #784, opened 2026-02-26
Remove toolkits

EricGustin:
These are being moved to monorepo. This repository already has plenty of examples at `examples/mcp_servers/`

Sister PR: https://github.com/ArcadeAI/monorepo/pull/449

---

> [!NOTE]
> **Medium Risk**
> Primarily repo/CI/build wiring changes with large deletions; risk is mainly broken workflows or missing packaging/build steps due to removed toolkit-specific logic.
> 
> **Overview**
> Removes the `toolkits/` packages (code, tests, evals, and per-toolkit configs) as they’re being moved to a separate monorepo.
> 
> Simplifies automation accordingly: the `setup-uv-env` composite action no longer has toolkit/contrib/lib modes or working-directory handling and always runs a single `uv sync`, `release-on-version-change` stops passing toolkit flags and no longer skips toolkit tests, and the root `Makefile` drops toolkit install/check/test/build/dist/clean targets plus collapses GHCR publishing to a single flow.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ca9e4aa6cb54dd8d4c47dc898876d510241407e6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #864, opened 2026-06-05
ci: remove dead poetry action, add Dependabot for github-actions (TOO-1024)

samkujovich:
## Summary

CI housekeeping in `arcade-mcp`, split out of **TOO-938** (whose security goal — SHA-pinning the live third-party actions — is already met and verified). Two small, non-security changes a reviewer can confirm without running anything:

1. **Delete `.github/actions/setup-poetry-env/`.** The repo migrated to `setup-uv-env`; this composite action is referenced by zero workflows and is non-functional anyway (it `cd arcade && poetry ...`, but the `arcade/` dir + `poetry.lock` were removed in #784). Its last real change was Jan 2025 (#193). It's also the last `snok/install-poetry@v1` mutable-tag reference flagged in TOO-938 — deleting dead code is the correct resolution rather than pinning something that never runs.
2. **Add `.github/dependabot.yml`** for the `github-actions` ecosystem (weekly, grouped) so action version pins in workflows are kept current automatically.

## Design decisions

- **Delete, don't pin, the poetry action.** Pinning a non-referenced, non-runnable action adds maintenance surface for zero benefit; removing it eliminates the TOO-938 finding outright.
- **`directory: "/"`** scans the workflow files in `.github/workflows/`. **Known limitation:** Dependabot does *not* update `uses:` inside local composite actions (`.github/actions/*/action.yml`) — [dependabot-core#6704](https://github.com/dependabot/dependabot-core/issues/6704). See Coverage below.
- **Single `groups` entry (`patterns: ["*"]`)** — bundle all action bumps into one weekly PR instead of one-per-action, to avoid PR spam given the shared `actions/*` deps. Annotated inline.
- **`commit-message.prefix: "ci"`** to match the repo's conventional-commit style.

## Coverage (what Dependabot will / won't bump)

- **Covered** (referenced in workflow files): `actions/checkout`, `actions/cache`, `actions/setup-python`, `actions/upload-artifact`, `actions/download-artifact`, `actions/stale`, `codecov/codecov-action`, `tj-actions/changed-files`, `pypa/gh-action-pypi-publish`, `slackapi/slack-github-action`, and the `astral-sh/setup-uv` used directly in `pylon-sync.yml`.
- **Not covered** (composite-action limitation): the `astral-sh/setup-uv` pin inside `.github/actions/setup-uv-env/action.yml`. Bumping that copy stays manual — which lines up with the optional `setup-uv` v6→v8 item on TOO-1024.

## Scope

- **In:** delete one dead composite action; add one Dependabot config file.
- **Out (intentionally):** bumping `astral-sh/setup-uv` v6.8.0 → v8.2.0. Already immutably SHA-pinned (no security urgency), crosses two majors with breaking changes — its own PR with a CI run. Tracked on TOO-1024.
- No workflow logic changes; nothing that runs in CI today is modified.

## Risk

Touches the `.github/` workflow path, so per our sensitive-path convention:

- **Blast radius: CI configuration only.** No product/runtime code, no published-package behavior.
- **Deleting the action is safe** — verified zero references repo-wide (`grep -rn setup-poetry-env`), and it could not have run anyway (missing `poetry.lock`). No workflow can break from its removal.
- **Dependabot is read-only** — it only opens PRs, which remain gated by normal review + CI. Worst case is an unwanted bump PR, which is closed, not merged.

## Test plan

- [x] `.github/dependabot.yml` parses and has the expected keys (`version: 2`, `github-actions`, `weekly`).
- [x] Repo-wide `grep -rn 'setup-poetry-env'` returns no references after deletion.
- [x] `git diff main...HEAD` is exactly the two intended files (+12 / −38), no stray changes.
- [x] Dependabot coverage verified against the documented behavior (composite-action limitation noted above).
- [x] CI green (re-run after a flaky `codecov-action@v4.0.1` upload; `Test libs` passed on all matrix entries).
- [ ] After merge: Dependabot's first `github-actions` PR appears within the weekly window.

Resolves: TOO-1024


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "ci"
+    groups:
```
samkujovich, line 9: Author note: grouping all `github-actions` updates under one group (`patterns: ["*"]`) intentionally bundles every action bump into a single weekly PR. Without this, Dependabot opens one PR per action — noisy here given the shared `actions/*` deps across workflows. If we later want a specific action (e.g. the OIDC `pypa/gh-action-pypi-publish`) to get its own isolated PR for closer review, we can split it into a second named group.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
```
samkujovich, line 4: Author note (scope caveat): `directory: "/"` covers `uses:` in **workflow files** only. Dependabot does **not** bump `uses:` inside local composite actions — [dependabot-core#6704](https://github.com/dependabot/dependabot-core/issues/6704) — so the `astral-sh/setup-uv` pin in `.github/actions/setup-uv-env/action.yml` stays manual. Every other action lives in a workflow and is covered. Flagging so a reviewer doesn't expect setup-uv bumps from this config.

jottakka (approved):

EricGustin (approved):

PR #881, opened 2026-06-21
feat(arcade-core): add CONTACTS service domain (TOO-1322)

samkujovich:
## What

Adds a `CONTACTS` member to `ServiceDomain` (`arcade_core.metadata`, re-exported via `arcade_mcp_server.metadata`):

```
CONTACTS = "contacts"
"""Address books and people directories — managing personal or organizational contact records."""
```

## Why

The `google_contacts` toolkit is the only Google toolkit in the monorepo with **no** `ServiceDomain` — every sibling sets one (gmail→EMAIL, calendar→CALENDAR, docs→DOCUMENTS, drive→CLOUD_STORAGE, sheets→SPREADSHEETS, slides→PRESENTATIONS). No existing value fits an address book:

- `CRM` fails the 10/10 test — Google Contacts is a personal/org address book, not a sales-pipeline CRM (Salesforce/HubSpot/Attio).
- `SALES_INTELLIGENCE` is prospecting/lead databases (Apollo), not a personal/org directory.

Per the metadata convention (the `Datadog → OBSERVABILITY` / `Apollo → SALES_INTELLIGENCE` precedent), when no `ServiceDomain` fits a recognized external service, add one upstream rather than force a wrong category.

## Changes

- `ServiceDomain.CONTACTS = "contacts"` with a docstring, placed next to `CRM` so the distinction is explicit.
- A test asserting the enum value and a classified-tool round-trip (mirrors `test_decorator_accepts_sales_intelligence_domain`).

No version bump (consistent with the prior service-domain additions; handled at release).

## Test plan

- [x] `uv run pytest -W ignore libs/tests/tool/test_tool_metadata.py libs/tests/arcade_mcp_server/test_tool_metadata_serialization.py` — 61 passed (incl. new `test_decorator_accepts_contacts_domain`)

## Downstream

Unblocks **TOO-1322** in the monorepo: once released, bump `arcade-mcp-server` in `google_contacts` and classify its 5 tools with `ServiceDomain.CONTACTS`.

Refs: TOO-1322

---

🤖 Generated with [Claude Code](https://claude.com/claude-code); author is accountable for every line.

---

> [!NOTE]
> **Low Risk**
> Additive enum and test only; no changes to auth, runtime behavior, or existing tool metadata rules.
> 
> **Overview**
[...]

PR #880, opened 2026-06-20
feat(metadata): add SURVEY ServiceDomain

samkujovich:
## Summary

Adds a `SURVEY` value to the `ServiceDomain` enum in `arcade_core.metadata`.

Survey / form / questionnaire builders (Google Forms, Typeform, SurveyMonkey, Qualtrics, etc.) have no fitting `ServiceDomain` today — none of the existing values passes the 10/10 test for "is this a survey tool?". Per the tool-metadata convention, a missing-but-warranted category should be added upstream (the same path `OBSERVABILITY` took for Datadog) rather than forcing tools into the wrong domain or defaulting to `None`.

```python
SURVEY = "survey"
"""Survey, form, and questionnaire builders for collecting and analyzing responses."""
```

## First consumer

The Google Forms worker toolkit (ArcadeAI/monorepo#939). Its 21 tools currently ship without a `Classification` because no domain fit; once this releases, that toolkit will bump its `arcade-mcp-server` minimum and classify all tools as `SURVEY`.

## Test plan

- Adds `test_decorator_accepts_survey_domain` mirroring the existing per-domain acceptance tests.
- New enum member only; no behavior change to existing domains.

---

> [!NOTE]
> **Low Risk**
> Additive enum member and version bump only; no changes to validation logic or existing ServiceDomain values.
> 
> **Overview**
> Adds **`ServiceDomain.SURVEY`** (`"survey"`) to `arcade_core.metadata` so survey, form, and questionnaire integrations (e.g. Google Forms) can be classified for discovery and search boosting instead of leaving `Classification` unset or mis-tagging another domain.
> 
> **`arcade-core`** is bumped **4.9.0 → 4.10.0** for the release. A decorator acceptance test **`test_decorator_accepts_survey_domain`** follows the same pattern as other per-domain tests; existing domains and validation rules are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1377fab0ed5700ec092da199c4cf3674a3804e1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #882, opened 2026-06-22
chore(arcade-core): bump to 4.10.0 for CONTACTS service domain (TOO-1322)

samkujovich:
## What

Bumps `arcade-core` `4.9.0` → `4.10.0`.

## Why

#881 added `ServiceDomain.CONTACTS` but was merged **without** a version bump. The prior service-domain addition (#866, `SALES_INTELLIGENCE`) shipped the enum + a minor `arcade-core` bump in the same PR (`4.7.3` → `4.8.0`); #881 missed the bump. This follow-up marks the release that includes `CONTACTS` so downstream consumers can pin it.

## Changes

- `libs/arcade-core/pyproject.toml`: `version = "4.10.0"` (minor — new enum value, backward-compatible).

## Test plan

- [x] No code change; `CONTACTS` already on `main` via #881. Bump only.

Refs: TOO-1322

---

🤖 Generated with [Claude Code](https://claude.com/claude-code); author is accountable for every line.

### DIFF
--- a/libs/arcade-core/pyproject.toml
+++ b/libs/arcade-core/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "arcade-core"
-version = "4.9.0"
+version = "4.10.0"
 description = "Arcade Core - Core library for Arcade platform"
 readme = "README.md"
 license = { text = "MIT" }


samkujovich:
Closing — consolidating the bump into #880 instead. #880 needs a 4.9.0 → 4.10.0 bump anyway (SURVEY), and since CONTACTS (#881) is already on main without a bump, doing it once in #880 ships both new domains together as 4.10.0. Avoids two PRs racing to bump the same version line. (Reopen if an arcade-core release is cut before #880 merges — then CONTACTS would need its own immediate bump to avoid shipping mislabeled.)

PR #889, opened 2026-07-06
fix(arcade-mcp-server): forward custom auth provider id, stop sending "None"

jottakka:
**Linear:** [TOO-1392](https://linear.app/arcadedev/issue/TOO-1392/arcade-mcp-server-check-authorization-sends-provider-idnone-and-drops)

## Problem

`MCPServer._check_authorization` ([`server.py`](libs/arcade-mcp-server/arcade_mcp_server/server.py)) built the wire-model `AuthRequirement` with:

```python
provider_id = str(getattr(req, "provider_id", ""))
auth_req = AuthRequirement(provider_id=provider_id, provider_type=..., oauth2=...)
```

Two defects:

1. **`str(None)` → `"None"`** — when a requirement has no `provider_id` (the normal case for **custom OAuth2 providers**, which identify via `id`), the literal string `"None"` was sent to the Arcade Engine as a bogus provider alias.
2. **`id` dropped** — `req.id` was never forwarded, so the Engine could not resolve a custom provider.

Together these broke authorization for every custom-provider toolkit going through the MCP path, forcing a subclass-with-hardcoded-`provider_id` workaround. Built-in providers (Google, GitHub, …) were unaffected — they set a class-level `provider_id` and leave `id` unset, so they route by `provider_id`.

The definition-building path (`catalog.create_auth_requirement`) already preserved `id`; only the server's `authorize()` call was lossy.

## Fix

The wire `AuthRequirement` TypedDict has all-optional keys, so build it conditionally — only include `provider_id` / `id` when actually set. No more stringified `None`; `id` is forwarded for custom providers.

## Reproduction (before)

For `@tool(requires_auth=OAuth2(id="my-custom-provider", scopes=["read"]))`:

| | `provider_id` | `id` |
|---|---|---|
| Definition (correct) | `None` | `"my-custom-provider"` |
| **Wire payload sent to Engine (buggy)** | `"None"` ❌ | *dropped* ❌ |
| **Wire payload (fixed)** | *omitted* ✅ | `"my-custom-provider"` ✅ |

## Tests

Adds two regression tests asserting the exact payload handed to `arcade.auth.authorize`:
- custom provider (`provider_id=None, id=...`) → forwards `id`, never sends `provider_id`
- built-in provider (`provider_id="google", id=None`) → forwards `provider_id`, never sends a bogus `id`

Confirmed red on the buggy code, green after the fix. `test_server.py` (188 tests) passes; mypy clean.

Bumps `arcade-mcp-server` `1.24.1 → 1.24.2` (patch).

---

> [!NOTE]
> **Medium Risk**
> Touches auth wire payloads for all OAuth tool calls through MCP; scope is small and well-tested but incorrect forwarding would break authorization for custom providers.
> 
> **Overview**
> **Fixes OAuth authorization for custom providers on the MCP server path** by correcting how `MCPServer._check_authorization` builds the `AuthRequirement` passed to `arcade.auth.authorize`.
> 
> Previously, `provider_id` was always stringified (so missing values became the literal `"None"`) and `req.id` was never forwarded—breaking Engine resolution for custom OAuth2 toolkits while built-in providers (e.g. Google) kept working via `provider_id`.
> 
> The change **only adds `provider_id` and `id` to the wire payload when they are actually set**, matching the optional-key shape of the TypedDict and aligning runtime behavior with `create_auth_requirement` in the catalog.
> 
> **Adds regression tests** for custom-only `id`, built-in-only `provider_id`, and both fields together, and **bumps** `arcade-mcp-server` **1.24.1 → 1.24.2**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68a1ec8f80422b9fec5039bc656e59e9f8098fc3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-mcp-server/arcade_mcp_server/server.py
+++ b/libs/arcade-mcp-server/arcade_mcp_server/server.py
@@ -1877,7 +1877,7 @@ class MCPServer:
             )
 
         req = tool.definition.requirements.authorization
-        provider_type = str(getattr(req, "provider_type", "") or "")
+        provider_type = str(getattr(req, "provider_type", ""))
         # TypedDict requires concrete type; supply empty scopes if absent when oauth2 provider
         oauth2_req = (
             AuthRequirementOauth2(
--- a/libs/arcade-mcp-server/arcade_mcp_server/server.py
+++ b/libs/arcade-mcp-server/arcade_mcp_server/server.py
@@ -1889,9 +1889,7 @@ class MCPServer:
         auth_req = AuthRequirement(provider_type=provider_type, oauth2=oauth2_req)
         # provider_id and id are both optional on the requirement. Built-in providers
         # identify via provider_id; custom providers identify via id (with no
-        # provider_id). Forward each only when set: coercing an absent value with
-        # str() would send the literal string "None" as the provider alias, and
-        # dropping `id` would leave the Engine unable to resolve a custom provider.
+        # provider_id). Forward each only when set.
         provider_id = getattr(req, "provider_id", None)
         if provider_id:
             auth_req["provider_id"] = str(provider_id)
--- a/libs/arcade-mcp-server/arcade_mcp_server/server.py
+++ b/libs/arcade-mcp-server/arcade_mcp_server/server.py
@@ -1877,8 +1877,7 @@ class MCPServer:
             )
 
         req = tool.definition.requirements.authorization
-        provider_id = str(getattr(req, "provider_id", ""))
-        provider_type = str(getattr(req, "provider_type", ""))
+        provider_type = str(getattr(req, "provider_type", "") or "")
         # TypedDict requires concrete type; supply empty scopes if absent when oauth2 provider
         oauth2_req = (
             AuthRequirementOauth2(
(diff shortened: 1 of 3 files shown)


libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -1887,11 +1886,18 @@ async def _check_authorization(
...
-            provider_id=provider_id,
-            provider_type=provider_type,
-            oauth2=oauth2_req,
-        )
+        auth_req = AuthRequirement(provider_type=provider_type, oauth2=oauth2_req)
+        # provider_id and id are both optional on the requirement. Built-in providers
+        # identify via provider_id; custom providers identify via id (with no
+        # provider_id). Forward each only when set: coercing an absent value with
+        # str() would send the literal string "None" as the provider alias, and
+        # dropping `id` would leave the Engine unable to resolve a custom provider.
```
EricGustin:
```suggestion
        # provider_id and id are both optional on the requirement. Built-in providers
        # identify via provider_id; custom providers identify via id (with no
        # provider_id). Forward each only when set.
```

EricGustin (approved):
looks good. Thanks!

PR #890, opened 2026-07-08
feat(auth): add MicrosoftPowerBI OAuth provider

jottakka:
## What

Adds a strongly-typed `MicrosoftPowerBI` OAuth2 provider class to the framework, matching the existing provider classes (`Microsoft`, `GitHub`, etc.), so a toolkit can eventually do:

```python
from arcade_mcp_server.auth import MicrosoftPowerBI
```

instead of defining the class locally.

## Why

Power BI Service resource scopes (`https://analysis.windows.net/powerbi/api/*`) cannot be requested in the same authorize request as Microsoft Graph scopes — Entra rejects the mix with `AADSTS70011`. A dedicated provider (`provider_id = "microsoft-powerbi"`) authorizes on a separate connection while still reusing the Microsoft Entra endpoints.

## Changes

- **`libs/arcade-core/arcade_core/auth.py`** — new `MicrosoftPowerBI(Microsoft)` class placed next to `Microsoft`. It subclasses `Microsoft` (rather than `OAuth2`) so it inherits the Microsoft Entra endpoints and the `MicrosoftGraphErrorAdapter` mapping in `get_adapter_for_auth_provider` (which does `isinstance(auth_provider, Microsoft)`), overriding only `provider_id = "microsoft-powerbi"`. Keyword-only `__init__` matches the sibling classes.
- **`libs/arcade-mcp-server/arcade_mcp_server/auth/__init__.py`** — re-export added to the import block and `__all__`, alphabetically adjacent to `Microsoft`.
- **`libs/arcade-tdk/arcade_tdk/auth/__init__.py`** — same re-export (this file mirrors the mcp-server one; kept in sync).
- **`libs/tests/sdk/test_tool_decorator.py`** — added `MicrosoftPowerBI` cases to the parametrized `test_tool_decorator_with_auth_success`, plus dedicated tests asserting `provider_id == "microsoft-powerbi"`, `id`/`scopes` default to `None`, `isinstance(..., Microsoft)`, and importability from `arcade_mcp_server.auth`.

Note: `arcade_core.auth` has no module-level `__all__`, so there is nothing to update there.

## Verification

- `uv sync --extra all --extra dev`
- `uv run python -c "from arcade_mcp_server.auth import MicrosoftPowerBI; print(MicrosoftPowerBI().provider_id)"` → `microsoft-powerbi`
- Confirmed error-adapter inheritance: `get_adapter_for_auth_provider(MicrosoftPowerBI())` → `MicrosoftGraphErrorAdapter`
- `uv run pytest libs/tests/sdk/test_tool_decorator.py` → 20 passed
- `uv run ruff check` + `ruff format --check` on all changed files → clean
- `uv run mypy .` in `libs/arcade-core` → no issues

---

## Update (post-initial-review)

- Caught the branch up to `main` (it had fallen 31 commits behind).
- Folded in a related arcade-core addition: `ServiceDomain.BUSINESS_INTELLIGENCE` (Power BI / BI-platform tool classification) with a decorator test.
- Added the version bumps this branch was missing (additive public API): **arcade-core 4.10.0 → 4.11.0**, **arcade-tdk 3.9.0 → 3.10.0** (floor `arcade-core>=4.11.0`), **arcade-mcp-server 1.24.2 → 1.25.0** (floors `arcade-core>=4.11.0`, `arcade-tdk>=3.10.0`). Verified a clean isolated wheel install resolves these floors (and that an old `arcade-core==4.10.0` is correctly rejected).
- Added a symmetric importability test for `arcade_tdk.auth` (the re-export lives in both packages).

### Error-adapter scope caveat

`MicrosoftPowerBI` subclasses `Microsoft`, so it inherits the `MicrosoftGraphErrorAdapter` mapping. That adapter is **Microsoft Graph SDK-specific** — it only translates exceptions originating from `msgraph`/`kiota` and no-ops otherwise. A Power BI toolkit calling the Power BI Service REST API (`https://analysis.windows.net/powerbi/api/*`) over plain `httpx` will **not** trigger it and falls through to the generic HTTP adapter. The inheritance is still correct (right Entra endpoints, no worse than the HTTP fallback), but it does **not** provide Power BI-specific error mapping. See the comment thread re: whether a dedicated Power BI adapter is warranted.

---

> [!NOTE]
> **Low Risk**
> Additive public API and metadata enum value with tests; no changes to existing auth flows beyond new optional provider usage.
> 
> **Overview**
> Introduces **`MicrosoftPowerBI`**, a typed OAuth2 provider (`provider_id` **`microsoft-powerbi`**) that subclasses **`Microsoft`** so Entra endpoints and Graph-oriented error-adapter routing stay the same, while Power BI resource scopes can use a **separate connection** and avoid mixing with Graph scopes (Entra **AADSTS70011**).
> 
> The class is re-exported from **`arcade_mcp_server.auth`** and **`arcade_tdk.auth`**. **`ServiceDomain.BUSINESS_INTELLIGENCE`** is added for BI / data-viz tool metadata. Package versions bump (**arcade-core** 4.11.0, **arcade-tdk** 3.10.0, **arcade-mcp-server** 1.25.0) with matching dependency floors. Tests cover decorator auth wiring, defaults, and import paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ec5133e5d6f04097f3dc3e69ba07e42b9221fbb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jottakka:
@EricGustin — a question on the error-adapter side before we lock this in, since you've built most of the recent provider adapters.

`MicrosoftPowerBI` subclasses `Microsoft`, so it inherits `MicrosoftGraphErrorAdapter` via the `isinstance(auth_provider, Microsoft)` check in `get_adapter_for_auth_provider`. But that adapter is **Graph-SDK-specific** — it only maps exceptions coming from `msgraph`/`kiota` and no-ops on anything else. The Power BI toolkit hits the Power BI Service REST API (`https://analysis.windows.net/powerbi/api/*`) over plain `httpx`, so in practice those calls fall through to the generic HTTP adapter, not the Graph mapping.

Do you think it's worth adding a **dedicated Power BI error adapter** (mapping the Power BI REST error shapes / common statuses like 401/403/429 with retry hints) instead of leaning on the Graph one? Or is the generic HTTP fallback good enough for now and a separate adapter isn't worth it yet?

For context, this PR keeps `MicrosoftPowerBI` as a `Microsoft` subclass mainly to reuse the Entra endpoints + provider routing; the error-adapter inheritance is more of a side effect than a real benefit here. Happy to split a dedicated adapter into a follow-up if you think it's worthwhile.

byrro (approved):

PR #892, opened 2026-07-10
fix(arcade-mcp-server): derive engine URL from arcade login coordinator

EricGustin:
## Summary

A local server running a tool that requires auth loads its OAuth token from the coordinator recorded by `arcade login`, but the engine `base_url` used for authorization always defaulted to the production `api.arcade.dev`. If you logged into a non-production environment (e.g. a staging domain), the server sent your staging token to the production engine and authorization failed with `401 invalid_authorization` / "Invalid API credentials". Login succeeds (it hits the coordinator, which is staging), so the failure is confusing.

This resolves the engine URL from the stored `coordinator_url` when the credentials come from `arcade login`, so the token and the engine target the same environment.

Resolves: 

## Design decisions

Resolution precedence in `_resolve_arcade_api_url`:

1. An explicitly configured URL wins: `arcade_api_url` constructor arg, `ARCADE_API_URL` env, or a non-default `MCPSettings` value.
2. Otherwise, when the token was loaded from `arcade login`, derive the engine URL from `config.coordinator_url` by swapping the leading host label (`cloud.<domain>` to `api.<domain>`), preserving scheme and port. This mirrors the existing `PROD_COORDINATOR_HOST` / `PROD_ENGINE_HOST` convention.
3. Fall back to the production engine.

Deliberate guardrails:

- **Derivation only fires for `arcade login` credentials.** If an explicit `ARCADE_API_KEY` is set, the caller owns the key/URL pairing and a stale login coordinator must not silently redirect them.
- **Only maps recognized `cloud.*` hosts.** `localhost` and custom single-host deployments return `None` and fall back to the production default rather than targeting a guessed URL.
- **Production is unchanged.** `cloud.arcade.dev` derives `api.arcade.dev`, identical to today's default (and no log line is emitted in that case).

Alternative considered and rejected: deriving the default at the `ArcadeSettings` layer. That would push config-file IO into an import-time settings singleton and affect the HTTP/stdio startup banners; keeping the logic in the runtime auth client localizes it next to where the token and org/project context are already loaded from the same config.

resolves https://linear.app/arcadedev/issue/TOO-1456/auto-detect-non-prod-domain-when-running-local-arcade-mcp-server

---

> [!NOTE]
> **Medium Risk**
> Changes runtime Arcade client `base_url` selection for login-based local dev; behavior is guarded by explicit URL/key overrides and extensive tests, but mis-derivation could still break auth in edge deployments.
> 
> **Overview**
> Fixes **401 invalid authorization** when running a local MCP server after `arcade login` against a non-production coordinator: the OAuth token came from staging but tool auth always hit production `api.arcade.dev`.
> 
> **Engine URL resolution** now runs in `MCPServer` when credentials are loaded from login (not when an explicit `arc_*` service key is set). Explicit `arcade_api_url`, `ARCADE_API_URL`, or a non-default settings `api_url` still win; otherwise the server reads `config.coordinator_url` and maps `cloud.<domain>` → `api.<domain>` (scheme/port preserved). Unrecognized hosts (`localhost`, non-`cloud.*`) fall back to the production engine. Default settings `api_url` is treated as unset so derivation can apply.
> 
> Adds `_engine_url_from_coordinator`, `_coordinator_url_from_config`, and `_resolve_arcade_api_url`, plus tests in `test_arcade_client_url.py`. Patch release **1.25.1**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6940cd283ae8ff3c3d68c69521df5a06c42b9170. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-mcp-server/arcade_mcp_server/server.py
+++ b/libs/arcade-mcp-server/arcade_mcp_server/server.py
@@ -23,6 +23,7 @@ from urllib.parse import quote, urlparse, urlunparse
 
 from arcade_core.auth_tokens import get_valid_access_token
 from arcade_core.catalog import MaterializedTool, ToolCatalog
+from arcade_core.constants import PROD_COORDINATOR_HOST, PROD_ENGINE_HOST
 from arcade_core.errors import ErrorKind, ToolInputError
 from arcade_core.executor import ToolExecutor
 from arcade_core.log_extras import build_tool_error_log_extra, build_tool_error_span_attributes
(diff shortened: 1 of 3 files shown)


libs/arcade-mcp-server/arcade_mcp_server/server.py
```diff
@@ -189,6 +190,41 @@ def _extract_scopes(claims: dict[str, Any]) -> set[str]:
...
+
+    parsed = urlparse(coordinator_url)
+    host = parsed.hostname
+    if not host:
+        return None
+
+    labels = host.split(".")
+    if labels[0] != coordinator_label:
+        return None
+
```
jottakka, line 220: `parsed.port` raises `ValueError` for a coordinator URL with a non-numeric/out-of-range port (for example `https://cloud.staging.example.com:bad`). `Config.coordinator_url` is only typed as `str`, so a malformed value in `credentials.yaml` makes `MCPServer` construction fail instead of taking the documented fallback path. Please catch `ValueError` while reading the port (or validate the parsed URL) and return `None` so this remains a non-fatal configuration issue.

jottakka (approved):
Cursor added a comment. But I will let it pre approved.

PR #893, opened 2026-07-14
fix(core): preserve typed tool inputs through the executor

EricGustin:
## Problem

Any tool whose input parameter is a Pydantic `BaseModel` (or `list[BaseModel]`) fails at execution with:

```
FatalToolError: An unhandled AttributeError was raised by the tool
  AttributeError: 'dict' object has no attribute '...'
```

## Root cause

`ToolExecutor.run` built the tool's call arguments with `inputs.model_dump()`:

```python
func_args = inputs.model_dump()   # recursively converts nested models to dicts
...
results = await func(**func_args)
```

`model_dump()` recursively converts every nested `BaseModel` to a plain `dict`. So a parameter declared as `person: Person` reaches the tool body as a `dict`, and `person.contact.address.city` raises `AttributeError`. The catalog builds the input model with the real annotated types (`create_func_models`), validation produces proper instances, and then `model_dump()` throws that structure away right before the call.

## Fix

Build the call arguments per field from the validated model, preserving Python-typed objects:

- **`BaseModel` params** are passed as model instances, so attribute access works.
- **`TypedDict` params** are re-emitted as plain dicts. The catalog wraps a `TypedDict` param in a Pydantic model for validation (`_wrap_typeddicts_as_models`, so `extra='forbid'` applies), but the tool body indexes it as a dict (`node["name"]`). A naive "keep everything as-is" fix would hand these tools the wrapper model and break subscripting, so a new `is_typeddict_model` predicate re-dumps exactly those (and lists of them).
- Scalars, `list[...]`, and `dict` params are unchanged.

## Tests

`libs/tests/core/test_executor.py` gains three cases through the full `ToolExecutor.run` path:

- nested `BaseModel` param (attribute access) — failed before, passes now
- `list[BaseModel]` param (attribute access on elements) — failed before, passes now
- `TypedDict` param (subscript access) — passed before and still passes (regression guard)

## Verification

- `libs/tests/core/` full suite: 442 passed
- `ruff check` / `ruff format --check` on changed source: clean
- `mypy` (lib-level, `--exclude tests`): no issues in 33 files

arcade-core bumped `4.9.0` → `4.9.1`. No public API change; behavior fix only.

---

> [!NOTE]
> **Medium Risk**
> Changes core executor argument shaping for every tool with structured inputs; fixes a widespread runtime bug but alters how `BaseModel` params reach tool code (from dicts to instances).
> 
> **Overview**
> **Tool execution** no longer passes validated inputs to tool functions via `inputs.model_dump()`. Instead, `ToolExecutor` builds kwargs per field so declared types survive the call boundary.
> 
> **`BaseModel` and `list[BaseModel]` parameters** stay as Pydantic instances (nested attribute access works). **`TypedDict` parameters** are still validated through strict wrapper models but are re-emitted as plain dicts via new `is_typeddict_model` and recursive `_to_func_arg` handling.
> 
> **Tests** cover nested `BaseModel`, `list[BaseModel]`, and `TypedDict` inputs through full `ToolExecutor.run`. **arcade-core** is patched **4.9.0 → 4.9.1** (behavior fix, no public API change).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50f430bf517f27db676bc0d2e9de9d17eb3eaa5f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-core/arcade_core/catalog.py
+++ b/libs/arcade-core/arcade_core/catalog.py
@@ -1336,6 +1336,17 @@ class _StrictTypedDictBaseModel(_TypedDictBaseModel):
     model_config = ConfigDict(extra="forbid")
 
 
+def is_typeddict_model(value: Any) -> bool:
+    """Return True when value is a Pydantic model synthesized from a TypedDict.
+
+    A parameter annotated as a TypedDict (or list[TypedDict]) is wrapped in a
+    Pydantic model for validation, but the tool function still expects a plain
+    dict. The executor uses this to re-emit such arguments as dicts while leaving
+    genuine BaseModel arguments as model instances.
+    """
+    return isinstance(value, _TypedDictBaseModel)
+
+
 def create_model_from_typeddict(
     typeddict_class: type, model_name: str, strict: bool = False
 ) -> type[BaseModel]:
--- a/libs/arcade-core/arcade_core/executor.py
+++ b/libs/arcade-core/arcade_core/executor.py
@@ -5,6 +5,7 @@ from typing import Any
 
 from pydantic import BaseModel, ValidationError
 
+from arcade_core.catalog import is_typeddict_model
 from arcade_core.errors import (
     ToolInputError,
     ToolOutputError,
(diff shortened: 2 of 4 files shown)


EricGustin:
bugbot run

PR #897, opened 2026-07-20
[PLT-2528] fix(arcade-serve): move OTel deps to the 1.38 generation

josebarrueta:
## What

Move `arcade-serve`'s pinned OpenTelemetry dependencies from the **1.28.2 / 0.49b2** generation to **1.38** (`>=1.38.0` exporters, `>=0.59b0` instrumentation).

```diff
-    "opentelemetry-instrumentation-fastapi==0.49b2"
-    "opentelemetry-exporter-otlp-proto-http==1.28.2"
-    "opentelemetry-exporter-otlp-proto-common==1.28.2"
-    "opentelemetry-instrumentation-httpx==0.49b2"
-    "opentelemetry-instrumentation-aiohttp-client==0.49b2"
-    "opentelemetry-instrumentation-requests==0.49b2"
+    "opentelemetry-instrumentation-fastapi>=0.59b0"
+    "opentelemetry-exporter-otlp-proto-http>=1.38.0"
+    "opentelemetry-exporter-otlp-proto-common>=1.38.0"
+    "opentelemetry-instrumentation-httpx>=0.59b0"
+    "opentelemetry-instrumentation-aiohttp-client>=0.59b0"
+    "opentelemetry-instrumentation-requests>=0.59b0"
```

## Why

The exact 1.28.2 pin makes `arcade-serve` **unsatisfiable** in the Arcade worker *supervisor* image (one isolated venv per toolkit version) for two independent reasons — the image cannot build at all:

1. **protobuf CVE floor** — the worker floors `protobuf>=6.31.1` (CVE-2025-4565), but `arcade-serve` → `opentelemetry-proto==1.28.2` → `protobuf>=5.0,<6.0`. Unsatisfiable.
2. **arcade-telemetry** — `arcade-telemetry` (and the Arcade coordinator) require `opentelemetry-exporter-otlp-proto-http>=1.38.0`, installed alongside `arcade_mcp_server` in the supervisor venv.

Both resolve once `arcade-serve` moves to the 1.38 generation (which supports protobuf 6/7 and matches `arcade-telemetry`).

## Safety

Version bump plus **type-only** adjustments in `arcade_serve/fastapi/telemetry.py` — no runtime behavior change (verified against `opentelemetry` 1.43 / instrumentation 0.64b). The newer stubs tightened `AioHttpClientInstrumentor._instrument` (now `**kwargs: Unpack[InstrumentKwargs]`, a `TypedDict`; httpx/requests still use `**kwargs: Any`), which surfaced three pre-existing type mismatches. Each is fixed properly — **no `# type: ignore`**:

- **Omit `tracer_provider`** to use the global provider (the API's documented default when the key is absent; runtime-identical to the previous explicit `None`, which the impl coerced via `kwargs.get(...)`).
- **Narrow the Optional** `self._tracer_provider` with an explicit `None`-check before instrumenting (`_init_tracer` sets it just above).
- **Narrow the ping span** with `isinstance(..., ReadableSpan)` — with the SDK provider active, `start_span` returns an SDK span (verified `True` at runtime), which is what the exporter consumes.

The OTel APIs `arcade-serve` imports are otherwise stable across 1.28→1.38 (`_log_exporter.OTLPLogExporter`, `.metric_exporter`, `.trace_exporter`; `sdk._logs`/`.metrics`/`.trace`/`.resources`; `_logs`, `semconv.*`, `instrumentation.*`). `mypy` + `ruff` clean.

## Validation

Built the full **115-toolkit** Arcade worker supervisor image with this change applied:
- **Builds green** — the supervisor venv resolves (previously failed here), all toolkit venvs build.
- **Boots and serves** — `GET /worker/health → 200`, `GET /worker/tools → 200` under the resolved coordinated set **OpenTelemetry 1.44 / instrumentation 0.64b0 / protobuf 7.x** (protobuf 7 satisfies the CVE floor).
- arcade-serve's telemetry setup imports resolve and run under 1.44.

Tracked in PLT-2528; unblocks the Arcade worker supervisor (PLT-2525).

---

> [!NOTE]
> **Low Risk**
> Dependency floor bump plus typing guards in telemetry setup; behavior is documented as unchanged and validated on a full supervisor image build.
> 
> **Overview**
> **Bumps `arcade-serve` OpenTelemetry packages** from pinned **1.28.2 / 0.49b2** to **`>=1.38.0` exporters** and **`>=0.59b0` instrumentation**, so the library resolves alongside `arcade-telemetry` and protobuf 6+ in the worker supervisor image.
> 
> In `telemetry.py`, **HTTP client instrumentors** on the arcade-telemetry path now call `_instrument()` **without** passing `tracer_provider` (same global-provider behavior as before). On the in-house OTLP path, the code **narrows** `self._tracer_provider` before passing it to instrumentors and **validates** the connectivity-probe span as `ReadableSpan` before export—changes driven by stricter OTel 1.38+ typing, not new behavior.
> 
> Tests expect `_instrument()` with no kwargs when arcade-telemetry is active.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 485e4b89a46587540e0b14a91ad16317ad83ebbc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-serve/arcade_serve/fastapi/telemetry.py
+++ b/libs/arcade-serve/arcade_serve/fastapi/telemetry.py
@@ -29,7 +29,7 @@ from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
 from opentelemetry.sdk.metrics import MeterProvider
 from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
 from opentelemetry.sdk.resources import Resource
-from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace import ReadableSpan, TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
 from opentelemetry.semconv._incubating.attributes import deployment_attributes
 from opentelemetry.semconv.attributes import service_attributes
@@ -84,12 +84,11 @@ class OTELHandler:
                 FastAPIInstrumentor().instrument_app(
                     app, excluded_urls=EXCLUDED_URLS, exclude_spans=EXCLUDED_SPANS
                 )
-                # Pass tracer_provider=None so instrumentors pick up the global
-                # provider set by arcade-telemetry.
-                HTTPXClientInstrumentor()._instrument(tracer_provider=None)
-                # aiohttp's instrumentor stub types tracer_provider as non-optional.
-                AioHttpClientInstrumentor()._instrument(tracer_provider=None)  # type: ignore[arg-type]
-                RequestsInstrumentor()._instrument(tracer_provider=None)
+                # Omit tracer_provider so the instrumentors resolve the global
+                # provider set by arcade-telemetry (the API's documented default).
+                HTTPXClientInstrumentor()._instrument()
+                AioHttpClientInstrumentor()._instrument()
+                RequestsInstrumentor()._instrument()
                 return
             # init_providers returned None (arcade-telemetry import race or
             # internal opt-out) — fall through to the in-house OTLP setup so
@@ -112,11 +111,14 @@ class OTELHandler:
         FastAPIInstrumentor().instrument_app(
             app, excluded_urls=EXCLUDED_URLS, exclude_spans=EXCLUDED_SPANS
         )
-        HTTPXClientInstrumentor()._instrument(tracer_provider=self._tracer_provider)
-        # aiohttp's instrumentor stub wants a non-optional api TracerProvider; the SDK
-        # provider set just above is compatible at runtime.
-        AioHttpClientInstrumentor()._instrument(tracer_provider=self._tracer_provider)  # type: ignore[arg-type]
-        RequestsInstrumentor()._instrument(tracer_provider=self._tracer_provider)
+        # _init_tracer set this just above; narrow away the Optional so the typed
+        # aiohttp instrumentor kwargs accept it.
+        tracer_provider = self._tracer_provider
+        if tracer_provider is None:
+            raise RuntimeError("tracer provider was not initialized")
+        HTTPXClientInstrumentor()._instrument(tracer_provider=tracer_provider)
+        AioHttpClientInstrumentor()._instrument(tracer_provider=tracer_provider)
+        RequestsInstrumentor()._instrument(tracer_provider=tracer_provider)
 
     def _init_tracer(self) -> None:
         self._tracer_provider = TracerProvider(resource=self.resource)
@@ -125,10 +127,13 @@ class OTELHandler:
         # Create an OTLP exporter
         self._tracer_span_exporter = OTLPSpanExporter()
 
+        # An SDK tracer provider is active (set just above), so start_span returns an
+        # SDK span; narrow to the ReadableSpan the exporter consumes.
+        ping_span = trace.get_tracer(__name__).start_span("ping")
+        if not isinstance(ping_span, ReadableSpan):
+            raise TypeError("expected an SDK ReadableSpan for the connectivity probe")
         try:
-            # start_span is typed as the API Span; at runtime it is an SDK span, which is
-            # the ReadableSpan the exporter consumes.
-            self._tracer_span_exporter.export([trace.get_tracer(__name__).start_span("ping")])  # type: ignore[list-item]
+            self._tracer_span_exporter.export([ping_span])
         except Exception as e:
             raise ConnectionError(
                 f"Could not connect to OpenTelemetry Tracer endpoint. Check OpenTelemetry configuration or disable: {e}"
--- a/libs/arcade-serve/arcade_serve/fastapi/telemetry.py
+++ b/libs/arcade-serve/arcade_serve/fastapi/telemetry.py
@@ -87,7 +87,8 @@ class OTELHandler:
                 # Pass tracer_provider=None so instrumentors pick up the global
                 # provider set by arcade-telemetry.
                 HTTPXClientInstrumentor()._instrument(tracer_provider=None)
-                AioHttpClientInstrumentor()._instrument(tracer_provider=None)
+                # aiohttp's instrumentor stub types tracer_provider as non-optional.
+                AioHttpClientInstrumentor()._instrument(tracer_provider=None)  # type: ignore[arg-type]
                 RequestsInstrumentor()._instrument(tracer_provider=None)
                 return
             # init_providers returned None (arcade-telemetry import race or
@@ -112,7 +113,9 @@ class OTELHandler:
             app, excluded_urls=EXCLUDED_URLS, exclude_spans=EXCLUDED_SPANS
         )
         HTTPXClientInstrumentor()._instrument(tracer_provider=self._tracer_provider)
-        AioHttpClientInstrumentor()._instrument(tracer_provider=self._tracer_provider)
+        # aiohttp's instrumentor stub wants a non-optional api TracerProvider; the SDK
+        # provider set just above is compatible at runtime.
+        AioHttpClientInstrumentor()._instrument(tracer_provider=self._tracer_provider)  # type: ignore[arg-type]
         RequestsInstrumentor()._instrument(tracer_provider=self._tracer_provider)
 
     def _init_tracer(self) -> None:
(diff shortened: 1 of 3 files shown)


libs/arcade-serve/arcade_serve/fastapi/telemetry.py
```diff
@@ -87,7 +87,8 @@ def instrument_app(self, app: FastAPI) -> None:
                 # Pass tracer_provider=None so instrumentors pick up the global
                 # provider set by arcade-telemetry.
                 HTTPXClientInstrumentor()._instrument(tracer_provider=None)
-                AioHttpClientInstrumentor()._instrument(tracer_provider=None)
+                # aiohttp's instrumentor stub types tracer_provider as non-optional.
+                AioHttpClientInstrumentor()._instrument(tracer_provider=None)  # type: ignore[arg-type]
```
sdreyer: why do we need to type ignore?

josebarrueta: With the upgrade mypy failed with these errors:

```
🔍 Type checking libs/arcade-serve/
arcade_serve/fastapi/telemetry.py:90: error: Argument "tracer_provider" to "_instrument" of "AioHttpClientInstrumentor" has incompatible type "None"; expected "TracerProvider"  [arg-type]
arcade_serve/fastapi/telemetry.py:115: error: Argument "tracer_provider" to "_instrument" of "AioHttpClientInstrumentor" has incompatible type "opentelemetry.sdk.trace.TracerProvider | None"; expected "opentelemetry.trace.TracerProvider"  [arg-type]
arcade_serve/fastapi/telemetry.py:126: error: List item 0 has incompatible type "Span"; expected "ReadableSpan"  [list-item]
Found 3 errors in 1 file (checked 13 source files)
🔍 Type checking libs/arcade-tdk/
Success: no issues found in 20 source files
❌ mypy failed in: libs/arcade-serve
```

josebarrueta:
Good call — I dug into each of the three mypy errors against the actual upgraded runtime (`opentelemetry` 1.43 / instrumentation 0.64b) to make sure the `# type: ignore`s weren't hiding a real behavior change. They weren't: **the upgrade only tightened types, runtime behavior is unchanged.** I've replaced all three ignores with proper fixes (pushed 38d8bcf).

**Root cause of the errors:** `AioHttpClientInstrumentor._instrument` now takes `**kwargs: Unpack[InstrumentKwargs]` (a `TypedDict`), where `tracer_provider` is `NotRequired[TracerProvider]`. `HTTPXClientInstrumentor` / `RequestsInstrumentor` still use `**kwargs: Any`, which is why only the aiohttp calls were flagged. Nothing in the call semantics changed.

**Each site — verified and fixed properly:**

1. **`_instrument(tracer_provider=None)` (arcade-telemetry branch).** The TypedDict types the key as `NotRequired[TracerProvider]`, so `None` is technically off-contract; the documented default when the key is *omitted* is the global provider — exactly the original intent. The impl does `kwargs.get("tracer_provider")`, so passing `None` and omitting are runtime-identical (both → global). Fix: **omit** the arg. Verified `_instrument()` instruments against the global provider.

2. **`_instrument(tracer_provider=self._tracer_provider)` (in-house branch).** Only issue was the declared `Optional` (`_init_tracer` sets it just above; the SDK `TracerProvider` is a subclass of the API one, so that part was already fine). Fix: **narrow** via an explicit `None`-check before the calls.

3. **`export([start_span("ping")])`.** `start_span` is typed as the API `Span`, but with the SDK provider active it returns an SDK span. Verified at runtime: `isinstance(start_span("ping"), ReadableSpan) is True`. Fix: **narrow with `isinstance(..., ReadableSpan)`** so the exporter's `ReadableSpan` type is satisfied.

No `# type: ignore` remain; used explicit `if … raise` narrowing (not `assert`, per the repo's `S101`). `mypy` + `ruff` clean, module imports and the narrowed paths exercise correctly. End-to-end, the full 116-toolkit worker image still builds, boots, and serves under this OTel set.

libs/arcade-serve/arcade_serve/fastapi/telemetry.py
josebarrueta: Good catch — we don't. I removed all three ignores in 38d8bcf. I'd verified against the upgraded runtime (opentelemetry 1.43 / 0.64b) that the errors were only tightened *types* (aiohttp's `_instrument` gained a typed `Unpack[InstrumentKwargs]`), not a behavior change, and replaced them with proper narrowing: omit `tracer_provider` to use the global default, an explicit `None`-check for the Optional, and `isinstance(..., ReadableSpan)` for the ping span (verified `True` at runtime). No `type: ignore` remain; mypy + ruff clean. Details in the PR comment above.

libs/arcade-serve/arcade_serve/fastapi/telemetry.py
```diff
@@ -84,11 +84,11 @@ def instrument_app(self, app: FastAPI) -> None:
...
                     app, excluded_urls=EXCLUDED_URLS, exclude_spans=EXCLUDED_SPANS
                 )
-                # Pass tracer_provider=None so instrumentors pick up the global
-                # provider set by arcade-telemetry.
-                HTTPXClientInstrumentor()._instrument(tracer_provider=None)
-                AioHttpClientInstrumentor()._instrument(tracer_provider=None)
-                RequestsInstrumentor()._instrument(tracer_provider=None)
+                # Omit tracer_provider so the instrumentors resolve the global
+                # provider set by arcade-telemetry (the API's documented default).
+                HTTPXClientInstrumentor()._instrument()
```
sdreyer, line 89: I believe I have set it to None previously because not setting it made it try to initialize again. Just want to check if if remembering it correctly

josebarrueta: Verified that omitting `tracer_provider` here does **not** trigger any re-initialization — it resolves the exact same global provider that `arcade-telemetry` set, byte-for-byte identical to passing `tracer_provider=None`. Proof and validations below.

**1. Omitted and `None` collapse to the same value.** All three instrumentors read the arg via `kwargs.get("tracer_provider")`, which returns `None` whether the key is passed as `None` or omitted entirely (`opentelemetry-instrumentation` 0.64b0):

```python
# httpx / requests
tracer_provider = kwargs.get("tracer_provider")
tracer = get_tracer(__name__, __version__, tracer_provider=tracer_provider)
# aiohttp
tracer_provider=kwargs.get("tracer_provider")
```

**2. `get_tracer(..., tracer_provider=None)` reuses the existing global — it never initializes a provider.** From `opentelemetry.trace` (api 1.43.0):

```python
def get_tracer(..., tracer_provider=None, ...):
    if tracer_provider is None:
        tracer_provider = get_tracer_provider()   # returns the already-set global
    return tracer_provider.get_tracer(...)

def get_tracer_provider() -> TracerProvider:
    if _TRACER_PROVIDER is None:
        if OTEL_PYTHON_TRACER_PROVIDER not in os.environ:
            return _PROXY_TRACER_PROVIDER          # returns proxy, does NOT set/create
        tracer_provider = _load_provider(...)
        _set_tracer_provider(tracer_provider, log=False)
    return cast("TracerProvider", _TRACER_PROVIDER)
```

`get_tracer_provider()` only ever assigns `_TRACER_PROVIDER` when the `OTEL_PYTHON_TRACER_PROVIDER` env var is set. In our path `arcade-telemetry`'s `init_providers()` runs **before** these `_instrument()` calls (telemetry.py:76 → 89-91), so `_TRACER_PROVIDER` is already populated and is simply returned. There is no initialization branch on this path.

**3. Empirical assertion.** With a global provider set (mimicking `arcade-telemetry`), both call forms pass `None` into `get_tracer`, and the global provider object is unchanged afterward — no re-init:

```
omit -> tracer_provider passed to get_tracer: None
None -> tracer_provider passed to get_tracer: None
IDENTICAL: True
global provider still the same object: True
```

**4. Test coverage.** `test_otel_handler_delegates_when_arcade_telemetry_present` asserts the exact call shape on this branch:

```python
mock_httpx.return_value._instrument.assert_called_once_with()
mock_aiohttp.return_value._instrument.assert_called_once_with()
mock_requests.return_value._instrument.assert_called_once_with()
```

`mypy` + `ruff` clean on `libs/arcade-serve/`; the telemetry-bridge suite passes (8/8).

sdreyer (approved):

PR #902, opened 2026-08-06
fix(arcade-tdk): make GraphQL adapter loader version-tolerant across gql 3.5.x/4.0.x (TOO-1338)

david-arcade-dev:
## TOO-1338 — Linear toolkit masks GraphQL errors as generic FatalToolError

Root-caused to **this** repo's GraphQL error adapter, not the downstream toolkit.

### The bug
`_load_gql_transport_errors()` (`libs/arcade-tdk/arcade_tdk/providers/graphql/error_adapter.py`) eagerly read `module.TransportConnectionFailed`, guarded **only** by `except ImportError`.

`TransportConnectionFailed` was added in **gql 4.0.0** and does **not** exist on the stable **3.5.x** line (the gql stable line jumps `3.5.3 -> 4.0.0`; no stable 3.6.x was ever released). Any consumer pinned to `gql>=3.5,<4.0` resolves to **3.5.3**, where that attribute access raises `AttributeError` — which is *not* an `ImportError`, so it escaped the guard.

Downstream, `_raise_as_arcade_error` in `tool.py` iterates the adapter chain inside a broad `except Exception` that logs and `continue`s. So the GraphQL adapter was **silently skipped**, the HTTP fallback didn't recognize a gql `TransportQueryError`, and the call fell through to `FatalToolError: An unhandled TransportQueryError was raised by the tool.` The real GraphQL `errors[]` message ("Entity not found" / "not authorized") was dropped from the agent-facing `message` (surviving only in `developer_message` / logs).

**Reproduced empirically** on real gql 3.5.3 (raised `AttributeError`) and confirmed fixed on both 3.5.3 and 4.0.0 (now → `UpstreamError`, status 404, message `Upstream GraphQL error: Entity not found`).

### The fix
Resolve each gql exception class with `getattr(module, name, _MissingGqlError)`, where `_MissingGqlError` is a private, never-raised sentinel. The adapter's `isinstance()` checks stay valid but simply never match a class the installed gql doesn't define — and every class a partial version *does* define still maps correctly (verified: `TransportProtocolError` still works on 3.5.x).

### Secondary bug (fixed in this PR)
`_handle_query_error` seeded `status` at the 422 floor and only replaced it when `mapped > status`, so specific 4xx codes below 422 — **NOT_FOUND (404), FORBIDDEN (403), UNAUTHENTICATED (401), BAD_USER_INPUT (400)** — never won and were reported as 422. Now 422 is only the fallback when *no* recognized code is present; when codes map, the **highest mapped status wins**. This preserves the existing "5xx wins over 4xx in a multi-error response" contract (existing test: FORBIDDEN + INTERNAL_SERVER_ERROR → 500) while fixing lone-4xx masking. Fixed here because it's a one-block change in the same function on the same error path.

### Supported gql-version policy (decision)
arcade-tdk supports **gql 3.5.x AND 4.0.x**. gql is an optional/soft dependency (toolkits bring their own; it is *not* an arcade-tdk or test dependency). Rather than add a heavyweight multi-install CI matrix, regression coverage is **version-agnostic**: tests build a fake `gql.transport.exceptions` module holding exactly the class inventory of each version profile (3.5.x omits `TransportConnectionFailed`; 4.0.x includes it) and drive the real loader through it. This is deterministic, needs no gql install, and covers both lines. gql 4.0's other additions (`TransportConnectionClosed`/`TransportClosed`, `TransportAlreadyConnected`) are **not** referenced by the adapter, so nothing else is version-fragile.

### Tests
- Loader degrades (no `AttributeError`) on both profiles; sentinel occupies the missing slot on 3.5.x.
- Core regression: `TransportQueryError` → `UpstreamError` with the real message even when `TransportConnectionFailed` is absent.
- `TransportProtocolError` still maps on the 3.5.x profile (sentinel doesn't swallow it).
- 4xx status-selection: lone NOT_FOUND → 404; FORBIDDEN/UNAUTHENTICATED/NOT_FOUND/BAD_USER_INPUT each reported as themselves; unknown code still → 422.
- Full `libs/tests/sdk` + `libs/tests/tool` suites green (736 passed).

## Rollout — versions consumers must bump to
Downstream toolkits depend on the **published** packages, so the fix reaches them only after release. Bump minimums to:

- **arcade-tdk `>=3.10.1`** (bumped 3.10.0 → 3.10.1) — required for anyone depending on arcade-tdk directly.
- **arcade-mcp-server `>=1.25.2`** (bumped 1.25.1 → 1.25.2; its arcade-tdk floor raised to `>=3.10.1`) — required for anyone getting arcade-tdk transitively via arcade-mcp-server, so a fresh resolve is forced onto the fixed arcade-tdk.

### DIFF
--- a/libs/arcade-mcp-server/pyproject.toml
+++ b/libs/arcade-mcp-server/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "arcade-mcp-server"
-version = "1.25.1"
+version = "1.25.2"
 description = "Model Context Protocol (MCP) server framework for Arcade.dev"
 readme = "README.md"
 authors = [{ name = "Arcade.dev" }]
@@ -23,7 +23,7 @@ requires-python = ">=3.10"
 dependencies = [
     "arcade-core>=4.11.0,<5.0.0",
     "arcade-serve>=3.4.0,<4.0.0",
-    "arcade-tdk>=3.10.0,<4.0.0",
+    "arcade-tdk>=3.10.1,<4.0.0",
     "arcadepy>=1.5.0",
     "pydantic>=2.0.0",
     "fastapi>=0.100.0",
--- a/libs/arcade-tdk/arcade_tdk/providers/graphql/error_adapter.py
+++ b/libs/arcade-tdk/arcade_tdk/providers/graphql/error_adapter.py
@@ -29,11 +29,32 @@ _GQL_CODE_TO_STATUS = {
 }
 
 
+class _MissingGqlError(Exception):
+    """Sentinel for a gql exception class absent on the installed gql version.
+
+    Substituted for a class the installed gql does not define so the adapter's
+    ``isinstance()`` checks stay valid but simply never match. Never raised or
+    instantiated.
+    """
+
+
 @lru_cache(maxsize=1)
 def _load_gql_transport_errors() -> (
     tuple[type[Any], type[Any], type[Any], type[Any], type[Any]] | None
 ):
-    """Import gql transport exceptions lazily and cache the result."""
+    """Import gql transport exceptions lazily and cache the result.
+
+    Each class is resolved with ``getattr`` and a never-raised sentinel so a gql
+    version that omits one of these names degrades gracefully instead of raising
+    ``AttributeError``. In particular ``TransportConnectionFailed`` was added in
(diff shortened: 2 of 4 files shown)


david-arcade-dev:
Rolling back this quick-fix PR — re-approaching TOO-1338 via a proper research-spike (reproduce with evidence, then develop fix options interactively) before committing anything.

PR #912, opened 2026-08-22
fix(arcade-cli): resolve org/project coordinator URL from arcade login

### LINKED ISSUE #896: [Bug] CLI: org/project commands ignore saved coordinator host and default to Arcade Cloud
## Summary

`arcade login -h <coordinator-host>` saves the coordinator URL with the credentials, but the `org` and `project` command groups build their coordinator URL from `PROD_COORDINATOR_HOST` instead of the saved value. After a successful login against a dedicated instance, `arcade project list` calls `https://cloud.arcade.dev/api/v1/orgs/<org-id>/projects` and fails with `401 Unauthorized`, because the stored token belongs to the coordinator the user logged in to, not to Arcade Cloud.

Hit by a customer on a dedicated instance today. Login shows the correct identity and active project, then the very next suggested command (`arcade project list`, suggested by the login success message itself) fails.

## Repro

1. `arcade logout`
2. `arcade login -h coordinator.<tenant-domain>` (succeeds, shows "Logged in as ..." and the active project)
3. `arcade project list`

Result:

```
Failed to list projects: Client error '401 Unauthorized' for url
'https://cloud.arcade.dev/api/v1/orgs/<org-id>/projects'
```

## Root cause

`libs/arcade-cli/arcade_cli/project.py` and `org.py` each define a module-level `state["coordinator_url"]` computed from `PROD_COORDINATOR_HOST`, overridden only by a `--host/-h` option on the command group callback. Neither consults the `coordinator_url` saved by `arcade login`. Note `utils.get_auth_headers` already falls back to `config.coordinator_url`, so the token lookup and the request URL can disagree, which is exactly what produces the 401.

## Workaround

Pass the host on the command group (flag position matters, it belongs to the group, not the subcommand):

```
arcade project -h coordinator.<tenant-domain> list
arcade org -h coordinator.<tenant-domain> list
```

## Suggested fix

When no `-h` is passed, resolve the group's coordinator URL as `config.coordinator_url` first, then `PROD_COORDINATOR_HOST`, mirroring the resolution precedence introduced for the engine URL in #892. Worth auditing the other command groups (e.g. `secret`) for the same hardcoded default.

HBX814:
## Summary

`arcade login -h <coordinator-host>` saves the coordinator it authenticated against, and the token it saves is only valid for that coordinator. The `org` and `project` command groups, however, built their request URL from `PROD_COORDINATOR_HOST`, while `get_auth_headers` already fell back to the saved `config.coordinator_url` for the token. On a dedicated instance the two halves of the request disagreed: the call went to `https://cloud.arcade.dev` carrying a token minted by the tenant coordinator, and came back `401 Unauthorized`.

That made the failure land in a bad spot â€” `arcade project list` is the command the login success message itself suggests, so the first thing a user does after a successful login fails.

This resolves the group URL as **explicit flags â†’ coordinator saved by `arcade login` â†’ production**, so the token and the request URL always name the same environment.

Resolves: #896

## Design decisions

**Resolution lives in one shared helper.** `resolve_coordinator_url` in `arcade_cli/utils.py` sits next to `get_auth_headers`, which already does the token half of the same resolution. Keeping them adjacent is the point: these two must agree, and the bug was precisely that they didn't.

**Explicit flags stay authoritative, and that includes `--port`/`--tls`/`--no-tls`, not just `--host`.** The saved coordinator is a full URL, so honoring it *while* applying a partial override would mean guessing which half wins (does the saved `:8443` survive a bare `--tls`?). Rather than invent that precedence, any explicit connection flag means the caller is describing the target and the URL is built from the flags alone. This also keeps behavior byte-identical for anyone currently passing flags, including the `arcade project -h <host> list` workaround from the issue.

**The saved coordinator is used verbatim**, preserving its scheme and port, rather than being decomposed into a host and rebuilt.

**`--host` now defaults to `None` instead of `PROD_COORDINATOR_HOST`.** With the old default, a user explicitly typing `-h cloud.arcade.dev` was indistinguishable from not passing the flag, so "was this overridden?" could not be answered. `None` makes that distinction real.

**Reads go through a small `_coordinator_url()` accessor.** The module-level `state` dict previously held a hardcoded production URL that the group callback overwrote on every real invocation. Seeding it with production meant that if the callback were ever bypassed, the fallback would be the exact silent-401 behavior being fixed here. The accessor re-resolves instead, so the correct behavior is also the fallback.

**Failure to read the config degrades to `None`, not an exception.** Importing `arcade_core.config` loads the credentials file at import time and raises when logged out, so `get_saved_coordinator_url` swallows that and lets callers fall back to production. A logged-out user gets today's behavior.

## Scope

In scope:
- `org` and `project` command groups (the groups reported in #896).

Not in scope (deliberately):
- **`secret.py`, which the issue flags for the same audit.** It has the same shape of bug, but it is engine-scoped (`PROD_ENGINE_HOST`), not coordinator-scoped, so fixing it needs the `cloud.*` â†’ `api.*` derivation added in #892 â€” which currently lives in `arcade_mcp_server/server.py` as `_engine_url_from_coordinator`. Doing it properly means promoting that helper into `arcade-core` so the CLI and the MCP server share one implementation instead of duplicating it. That is a wider refactor across two packages and reads better as its own PR. Happy to follow up with it if you want it.
- `connect.py` already resolves `coordinator_url or f"https://{PROD_COORDINATOR_HOST}"`, so it is unaffected.

## Test plan

New `libs/tests/cli/test_coordinator_url.py` (20 tests) covering the resolver, the config read, both command groups end-to-end through Typer (`list` and `set`), and the accessor fallback.

- [x] **Confirmed the new tests actually reproduce the bug.** Reverting only `project.py`/`org.py` to the original code fails 4 of them with `Actual: fetch_projects('https://cloud.arcade.dev', 'org_1')` â€” the exact URL from the issue. They pass with the fix.
- [x] **Exercised the real end-user path with no mocks**, via a genuine post-login `credentials.yaml` (written through `Config.save_to_file`) and `ARCADE_WORK_DIR`:

  | | `project list` targets | `org list` targets |
  |---|---|---|
  | before | `https://cloud.arcade.dev` (â†’ 401) | `https://cloud.arcade.dev` (â†’ 401) |
  | after | `https://coordinator.tenant.example.com` | `https://coordinator.tenant.example.com` |

  And `arcade project -h override.example.com list` still targets `https://override.example.com`, so the issue's workaround keeps working.
- [x] Logged-out / unreadable-config case falls back to `https://cloud.arcade.dev` without raising.
- [x] `libs/tests/cli/` green: **806 passed, 45 skipped** (786 before this change, +20 new).
- [x] Root `pyproject.toml` bumped 1.15.2 → 1.15.3 (the version `arcade-cli` ships under). `test_dependency_alignment.py` passes.
- [x] Codecov flagged 2 uncovered patch lines on the first run: the `_coordinator_url()` call sites inside `arcade project set` / `arcade org set`, which the original tests did not reach. Those subcommands resolve the Coordinator the same way `list` does, so they are now covered too â€” a real gap, not a coverage-chasing edit.
- [x] `ruff check`, `ruff format --check`, and `pre-commit run --files <changed>` clean on the changed files; `mypy` clean on `arcade-cli` (29 files).
- [x] Full `libs/tests/` run: 3420 passed, 532 skipped (3400 on `main` before this change). The 2 failures in `arcade_mcp_server/integration/test_end_to_end.py` (`test_stdio_e2e`, `test_http_e2e`) are **pre-existing on `main`** â€” verified by stashing this change and re-running â€” and are a local server-spawn/port issue on Windows, unrelated to `arcade-cli`.

No version bump: behavior-only fix, no public API or dependency change.

## Rebased onto current `main` (Sep 16)

The branch had gone stale while waiting on workflow approval, so it is rebased onto `main` (`7246e9d6`). The rebase was clean — no conflicts; none of the files here were touched by the 10 commits `main` gained.

One thing the original push missed: this repo bumps the **root** `pyproject.toml` for `arcade-cli` changes (`arcade-cli` has no package `pyproject.toml` of its own), which 5 of the last 6 `libs/arcade-cli` commits on `main` do. Added: root `1.15.2` → `1.15.3`, a patch bump for a behavior-only fix.

## Risk note

Behavior changes only for users who pass **no** connection flags to `arcade org` / `arcade project`:

- Logged in to production â†’ unchanged (`cloud.arcade.dev` either way).
- Logged in to a dedicated/staging coordinator â†’ now targets that coordinator instead of returning 401. This is the fix.
- Logged out, or credentials unreadable â†’ unchanged (production).

Anyone passing `--host`/`--port`/`--tls`/`--no-tls` is unaffected by construction. The one way this could surprise someone is a **stale** `coordinator_url` in `credentials.yaml` from an old login, which would now be honored where production was used before; `arcade logout && arcade login`, or an explicit `-h`, resolves that.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to users on dedicated Coordinators who omit connection flags; explicit flags and logged-out production fallback stay the same.
> 
> **Overview**
> Fixes **401 Unauthorized** on `arcade org` / `arcade project` after logging into a non-production Coordinator: those groups used to always build the API base URL from production while auth tokens came from the Coordinator saved by `arcade login`.
> 
> **Coordinator URL resolution** is centralized in `resolve_coordinator_url` (with `get_saved_coordinator_url`), aligned with `get_auth_headers`. Precedence is **explicit `--host` / `--port` / `--tls` / `--no-tls` → saved login URL (verbatim) → production**. `--host` now defaults to `None` so “no flag” can be distinguished from an explicit override.
> 
> `org` and `project` stop seeding module `state` with a hardcoded production URL; the group callback sets URL via the resolver, and `_coordinator_url()` re-resolves when state is empty. New `libs/tests/cli/test_coordinator_url.py` covers the resolver and Typer paths; root package version **1.15.2 → 1.15.3**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc8909a4af06bfd7d2d7fb31a49881cfe096ada9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


EricGustin:
superseded by https://github.com/ArcadeAI/arcade-mcp/pull/934

Issue #703, opened 2025-12-02
Audit error messages for actionable fix instructions


TheMostlyGreat:
## Summary

Error messages should tell users **how to fix** the problem, not just **what** the problem is.

## Current State

Example - missing secret error:
```
Tool 'tool_name' cannot be executed because it requires the following secrets that are not available: API_KEY
```

This tells the user *what's wrong* but not *how to fix it*.

## Proposed State

```
✗ Missing secret 'API_KEY'

  Tool 'fetch_data' requires this secret but it's not configured.

  To fix, either:
  1. Add to .env file:     API_KEY=your_key_here
  2. Set environment var:  export API_KEY=your_key_here

  Then restart the server.
```

## Why This Matters

- **Reduces friction** - Users don't have to search docs or guess
- **Better DX** - Errors become self-service documentation
- **Fewer support questions** - The answer is in the error

## Inspiration

Elm and Rust compilers are famous for helpful error messages. Every error should answer: "What do I do now?"

## Scope

Audit all user-facing errors in `arcade-mcp-server`:

1. **Missing secrets** - Show how to set them
2. **Invalid tool input** - Show expected shape vs received
3. **Authorization failures** - Explain the auth flow
4. **Transport errors** - Suggest common fixes (port in use, etc.)
5. **Configuration errors** - Show valid options

Each error should include:
- Clear problem statement
- Specific context (which tool, which field, what value)
- Concrete fix instructions
- Next step (e.g., "Then restart the server")

## Related

This pattern is being implemented in the TypeScript version for consistency across implementations.

evantahler:
Linked to https://linear.app/arcadedev/issue/TOO-199/audit-error-messages-for-actionable-fix-instructions

PR #913, opened 2026-08-22
fix(arcade-mcp-server): make invalid tool input errors actionable and stop leaking internals

HBX814:
## Summary

Two changes to the invalid-tool-input path, found while auditing #703.

**1. The legacy branch shipped error internals to the client.** Input validation failures are version-gated: 2025-11-25 clients get a `CallToolResult`, 2025-06-18 clients get a JSON-RPC `-32602`. The legacy branch built that response with `str(error)` â€” which renders the whole `ToolCallError` pydantic model, so `developer_message` and `stacktrace` went out as `field=value` pairs.

That matters more than cosmetics. For an input-validation error the stacktrace is a Pydantic traceback, and Pydantic embeds `input_value=` in it, so **the rejected argument was echoed back verbatim**. `_serialize_input` carries an explicit comment that rejected values must stay out of the surfaced fields because they may hold secrets or PII, and `_debug_exposure` exists so stacktraces reach a client only behind an opt-in env flag. This branch bypassed both.

**2. Invalid-input errors now say what the tool expects** â€” the "Invalid tool input â€” show expected shape" bullet from #703.

Before:

```
Invalid input: units: Input should be 'c' or 'f'; days: Input should be a valid integer, unable to parse string as an integer
```

After:

```
Invalid input: units: Input should be 'c' or 'f'; days: Input should be a valid integer, unable to parse string as an integer

Expected:
  - units (string, optional): Temperature units [allowed values: c, f]
  - days (integer, optional): Number of forecast days

Fix these arguments and call the tool again.
```

Refs: #703 â€” deliberately not `Resolves`: this covers the one error category of that issue
that had not already been converted. See Scope for what remains.

## Design decisions

**The guidance is built from the `ToolDefinition`, never from the submitted values.** #703 phrases this category as "show expected shape vs received", but echoing *received* values is exactly what `_serialize_input` is written to avoid â€” a naive reading of the issue would reintroduce the leak that change 1 above removes. So only the declared schema is rendered. Two tests pin this by passing a sentinel value and asserting it appears nowhere.

**Only the rejected parameters are described.** The caller already received the full schema from `tools/list`; repeating all of it on every failure buries the actionable part and grows unbounded with the tool's arity. The block is bounded by the number of errors instead.

**The leading `Invalid input: <field>: <reason>` summary is unchanged, and guidance is appended after it.** Existing tests (and any caller) match on that prefix, so it stays byte-identical; nothing is spliced into it.

**`_serialize_input`'s new `definition` parameter is positional-only (`/`).** `**kwargs` holds caller-supplied tool arguments, and a tool is free to declare a parameter named `definition` â€” or `input_model`. Positional-only placement keeps such an argument in `kwargs` instead of raising "got multiple values for argument". This also closes the same latent collision that already existed for `input_model`.

**`definition` is optional and the helper returns `""` when it cannot describe anything.** Validation still works without a definition, so this is enrichment rather than a new requirement, and callers can append unconditionally.

**Change 1 keeps `additional_prompt_content` and routes internals through `augment_error_message_for_debug`** rather than dropping them, so the documented debug escape hatch still works on the legacy path â€” it just becomes the *only* way to expose internals there. Two tests assert the flags still surface `[DEBUG] stacktrace:` / `[DEBUG] developer_message:`.

## Scope

In scope: the "invalid tool input" category of #703, plus the internals leak found on the same path.

Not in scope:
- **Most of #703 is already done.** Missing secrets, reserved secrets, authorization failures, transport restrictions, and the generic tool-execution failure already emit the `âœ— â€¦ To fix:` shape the issue asked for. I only touched the category that hadn't been converted. Worth a maintainer confirming whether #703 should stay open for the rest.
- **The startup missing-secret warning** (`_check_and_warn_missing_secrets`) still says only *"declares secret(s) 'X' which is/are not set. It will return an error if called."* â€” no fix instructions, while the runtime error for the same condition has them. Small consistency gap, deliberately left out to keep this diff reviewable.
- **Transport errors ("port in use")**, item 4 of #703, are unhandled â€” `uvicorn.run()` surfaces the raw `OSError`. That needs a different area of the code and reads better as its own PR.
- **The legacy branch also skips `_log_tool_call_error` / `_record_tool_error_span_attributes`**, which the 2025-11-25 branch calls. That looks like an oversight, but changing logging/telemetry behavior is outside a message fix, so I left it alone rather than widen the blast radius.

## Test plan

Two new files, 25 tests, written before the implementation, per the repo's TDD requirement.

- [x] **Confirmed the tests fail on the current code first.** The pre-fix run reproduced the leak exactly â€” the client message was `message="[TOOL_RUNTIME_BAD_INPUT_VALUE] ToolInputError â€¦ Invalid input: tagsâ€¦=str]\n For further information visit https://errors.pydantic.dev/2.13/v/list_type\n' status_code=400 extra=None`, i.e. the full model repr with the Pydantic error inside.
- [x] Verified the leak directly at the source: for a bad argument, the sentinel value is absent from `message` and `developer_message` but **present in `stacktrace`**, and therefore present in `str(error)` â€” which is what the legacy branch sent.
- [x] `libs/tests/arcade_mcp_server/test_invalid_input_legacy_protocol.py` (8): rejected value not echoed; no `Traceback`; no `kind=` / `developer_message=` / `can_retry=` / `status_code=` model-repr fields; message still actionable; both debug flags still work; modern path unchanged.
- [x] `libs/tests/core/test_invalid_input_guidance.py` (17): expected shape, description, enum allowed-values, array element type, only-rejected-fields, next-step line, two no-leak tests, three backward-compatibility tests, and five covering the helper's quiet-degradation paths (no definition, no rejected fields, a definition without `parameters`, an unrecognized field, and a model-level error with no field location).
- [x] The legacy test resolves the tool name from the catalog rather than hardcoding it. An early version hardcoded `needs_a_list`, but Arcade derives `NeedsAList`, so the server answered "Unknown tool" and three absence-based assertions passed vacuously. `_legacy_error_text` now asserts the call actually reached input validation, so that failure mode cannot recur.
- [x] Full `libs/tests/` run: **3425 passed, 532 skipped** (3400 on `main` before, +25 new). No regressions â€” notably `test_executor.py`'s `startswith("â€¦ Invalid input: inp:")` assertions and `test_input_validation_error_does_not_leak_input_values` still pass.
- [x] The 2 failures in `arcade_mcp_server/integration/test_end_to_end.py` (`test_stdio_e2e`, `test_http_e2e`) are **pre-existing on `main`** â€” verified by stashing this change and re-running â€” a local server-spawn/port issue on Windows.
- [x] `pre-commit run -a` (what CI's `quality` job runs) fully clean, including the `check-debug-leak-flags` guard; `ruff check` and `ruff format --check` clean; `mypy` clean on `arcade-core` (33 files) and `arcade-mcp-server` (51 files).
- [x] Versions bumped once on this branch: `arcade-core` 4.18.0 â†’ 4.19.0 (additive behavior change), `arcade-mcp-server` 1.30.1 â†’ 1.30.2 (defect fix). `test_dependency_alignment.py` passes; the root constraints (`arcade-core>=4.9.0`, `arcade-mcp-server>=1.23.0`) already admit both, and no dependency floor needed raising since neither change is breaking.

## Note on the first CI run

The first push failed `quality` and `Debug leak flag guard`, both from one mistake of mine: the legacy test hardcoded the debug-flag activation acknowledgement string, which `scripts/check_debug_leak_flags_off.py` forbids outside a small allowlist. It passed locally only because that guard scans `git ls-files` and the file was still untracked when I ran the hooks.

Rather than widen the allowlist, the test now reads `_DEBUG_LEAK_MAGIC`, `_ENV_EXPOSE_STACKTRACE`, and `_ENV_EXPOSE_DEVELOPER_MESSAGE` from `_debug_exposure`, so the ack string stays confined to the files already permitted to contain it â€” and the constants have a single source of truth. Verified by running the guard script directly with the file staged, and by `pre-commit run -a`.

Codecov also flagged 4 uncovered patch lines. All four were defensive branches in the new helper plus the `additional_prompt_content` branch; they are now covered by the added tests. The lines still uncovered in `executor.py` (111, 115, 140-141) and `server.py` (1627) are pre-existing, outside this diff.

## Rebased onto current `main` (Sep 16)

The branch had gone stale while waiting on workflow approval, so it is rebased onto `main` (`7246e9d6`). Two conflicts, both version bumps only — `main` had moved `arcade-core` 4.11.0 → 4.18.0 and `arcade-mcp-server` 1.26.0 → 1.30.1, leaving the original bumps below `main`. Re-bumped from the current numbers; the code changes themselves merged without conflict.

Re-checked that the fix is still needed rather than assuming: `str(error)` is still on the legacy bad-input branch on `main` today, and the merged result places the replacement correctly.

## Risk note

Change 1 touches how tool errors are surfaced, which is reachable from `context.get_secret()`-adjacent flows, so worth stating precisely:

- **2025-06-18 clients, invalid input:** previously received the model repr including the rejected value; now receive the curated message. Any consumer that was parsing `field=value` pairs out of that string would break â€” but that string was never a documented contract, and the sibling protocol version never emitted it.
- **2025-11-25 clients:** unchanged.
- **Debug flags:** unchanged. Internals still reach the client when `ARCADE_DEBUG_EXPOSE_*` is set to the ack value, on both paths.
- **Change 2** only appends to a message. It is skipped entirely when no `ToolDefinition` is available, and no rejected value can enter it by construction.
- MCP stdio channel is untouched: no new stdout/stderr writes, per the repo's stdio-channel rule.

---

> [!NOTE]
> **Medium Risk**
> Changes how invalid-input errors are surfaced to MCP clients (security-sensitive: stops echoing secrets/PII on the legacy path) and extends user-visible error text; modern protocol behavior is unchanged aside from richer messages from core.
> 
> **Overview**
> Fixes two problems on the invalid-tool-input path: **legacy MCP clients were getting full error dumps (including rejected argument values via Pydantic stacktraces)**, and **validation failures only said what was wrong, not what the tool expects**.
> 
> In **arcade-core**, validation errors still start with `Invalid input: …` but now append an **Expected:** block for **only the rejected parameters**, built from the tool’s `ToolDefinition` (types, required/optional, descriptions, enums)—never from submitted values. `_serialize_input` takes an optional positional-only `definition` so tool args named `definition`/`input_model` don’t collide with `**kwargs`.
> 
> In **arcade-mcp-server**, the 2025-06-18 JSON-RPC `-32602` branch no longer uses `str(error)`; it sends the curated `error.message` (plus `additional_prompt_content`) and routes stacktrace/developer details through **`augment_error_message_for_debug`**, matching the newer protocol path. Package versions bump to **arcade-core 4.19.0** and **arcade-mcp-server 1.30.2**, with new tests covering guidance, no-leak behavior, renamed wire params, and legacy debug flags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 214c6e285f0f26ffd5e8511682bbd46b61e043f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


HBX814:
Confirmed and fixed in 214c6e28 — thanks, this was a real miss.

Verified the mechanism rather than taking it on faith. `extract_field_info` sets `param_info.name = str_annotations[0]` for a two-string `Annotated`, while `create_func_models` keys `input_fields` on the name from `inspect.signature(...).parameters.items()`. So the two genuinely diverge, and since validation errors report the Pydantic name, the lookup missed. A probe on a renamed parameter:

```
declared InputParameter.name : ['wire_name']
pydantic input_model fields  : ['py_param']

Invalid input: py_param: Field required
>>> Expected block present: False
```

Worth noting it hit the most common case — a missing required argument — not just an edge case.

The fix resolves both names. `create_input_definition` and `create_func_models` walk the same `inspect.signature` in the same order and skip `ToolContext` identically, so the declared parameters line up positionally with the model fields; Python names are added with `setdefault` so a wire name is never shadowed by another parameter's Python name. `input_model` stays optional, so the declared-name path still works without it.

One thing that surfaced while fixing it: calling with the wire name rejects the same parameter under *both* names at once (`py_param: Field required; wire_name: Extra inputs are not permitted`), which made the first version print the parameter twice. Parameters are now de-duplicated and described once. The guidance still shows the wire name, since that is what the caller knows the parameter as from `tools/list`.

Five tests added. Three of them fail against the previous commit (the two behavioural ones plus the helper), so the regression is actually pinned rather than passing vacuously. `mypy`, `ruff`, and `pre-commit run -a` are clean; full suite 3430 passed.

Separately, the probe showed that calling a renamed parameter by its advertised wire name fails validation outright (`Extra inputs are not permitted`) — the executor never maps wire names back to Python names. That looks pre-existing and unrelated to this PR, so I have left it alone, but flagging it in case it is news.

HBX814:
Rebased onto current `main` (e0f09dfe). The only conflicts were the version bumps, which are now `arcade-core` 4.21.0 → 4.22.0 and `arcade-mcp-server` 1.32.0 → 1.32.1. The code changes are unchanged.

Re-verified after the rebase:
- Full `libs/tests`: 4157 passed, 5 skipped. `integration/test_end_to_end.py` is excluded; it already fails on `main` on Windows.
- `pre-commit run -a` is clean.

This is ready for review whenever someone has time. The startup-warning gap from the Scope section is now a separate PR: #946.

PR #918, opened 2026-08-25
feat(core): add a resource registry beside the tool catalog [TOO-1932]

EricGustin:
## Summary

Adds a URI-to-resource registry to `arcade-core`, so a worker has somewhere to keep the resources a toolkit ships. Its own type, its own file.

Resources on this path are static. Contents resolve once when a toolkit registers and are held from then on, so nothing executes while a request is in flight and the endpoints that read this registry stay a thin serialization layer.

Part 2 of 6. Stacked on #923.

Resolves: https://linear.app/arcadedev/issue/TOO-1932

## Design decisions

**The registry hangs off `ToolCatalog`.** That is the one object the server factory hands to the worker (`worker.catalog = catalog`), so anything living elsewhere needs new plumbing to reach it. The registry is still a separate type in a separate file; the catalog just carries it. It is a `PrivateAttr` with a `default_factory`, so two catalogs can never end up sharing one registry. There is a test for that, because the bare-mutable-class-default version of this bug is easy to write and hard to see.

**Listing is ordered by URI, and the order is kept on the write.** A worker can run as several processes behind one address, and a cursor issued by one process has to mean the same thing when the next page lands on a different one. Insertion order gives you no such guarantee.

`add` keeps a URI index sorted with `bisect.insort` and `list` slices it, so the read is `O(page_size)` rather than sorting the whole set each time. A worker registers once at startup and lists on every request, so the ordering cost belongs on the write. Re-registering a URI replaces in place without inserting a second time.

`page_size` validates on assignment rather than only in `__init__`, because it is a public attribute callers set directly. A non-positive value makes cursor paging non-terminating: zero hands back an empty page and the same cursor forever, and a negative one walks the offset backwards into a cursor this registry then rejects as malformed.

Text and blob stay separate types. Keeping them apart by type, instead of by which field happens to be empty, means a zero-length document and a zero-length blob remain distinguishable all the way out to the caller.

## Test plan

- [x] 14 new tests in `libs/tests/core/test_resource_registry.py`: the text/blob split, cursor round-tripping, rejection of a cursor this registry did not issue, per-catalog isolation, that replacing a URI leaves exactly one entry in the ordered index, and that a non-positive `page_size` is refused at construction and at assignment

## Author checklist

---

> [!NOTE]
> **Low Risk**
> Additive library API with no changes to tool execution or auth; risk is limited to future callers of the new registry and list cursors.
> 
> **Overview**
> Introduces **`ResourceRegistry`** in `arcade-core` so workers can hold toolkit **static** resources (resolved once at registration, served without running code on the request path). Registration maps each URI to **`TextResourceContents`** or **`BlobResourceContents`** (including empty str vs empty bytes), with typed errors for missing URIs and invalid inputs.
> 
> **`list`** pages resources in **URI sort order** using opaque cursors that encode the **last URI on the page** (not a numeric offset), so pagination stays consistent across multi-process replicas and mid-rollout catalog differences. **`ToolCatalog`** exposes a per-instance registry via **`catalog.resources`** (`PrivateAttr` + `default_factory` so catalogs do not share state).
> 
> Bumps **`arcade-core`** to **4.13.0** and adds **`test_resource_registry.py`** covering paging, cursor edge cases, replacement, and catalog isolation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit efde42055bfbd7d660b6b821f8cbcfea77833b72. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-core/arcade_core/resources.py
+++ b/libs/arcade-core/arcade_core/resources.py
@@ -1,0 +1,159 @@
+"""The resource registry: what a worker can serve for ``resources/*``.
+
+Resources on this path are static. Contents are resolved once, when a toolkit
+is registered, and held. Nothing is executed on the request path, so the
+endpoints that read this registry stay a serialization shim.
+"""
+
+from __future__ import annotations
+
+import base64
+import binascii
+import inspect
+from bisect import insort
+from dataclasses import dataclass
+
+from arcade_core.resource_schema import (
+    BlobResourceContents,
+    Resource,
+    TextResourceContents,
+)
+
+DEFAULT_PAGE_SIZE = 250
+
+_CURSOR_PREFIX = "offset:"
+
+
+class InvalidCursorError(ValueError):
+    """Raised when a caller sends a cursor this registry did not issue."""
+
+
+class ResourceNotFoundError(KeyError):
+    """Raised when no resource is registered at the requested URI."""
+
+
+def encode_cursor(offset: int) -> str:
+    """Cursors are opaque to callers, so the encoding can change freely."""
+    raw = f"{_CURSOR_PREFIX}{offset}".encode()
+    return base64.urlsafe_b64encode(raw).decode("ascii").rstrip("=")
+
+
+def decode_cursor(cursor: str) -> int:
+    padding = "=" * (-len(cursor) % 4)
+    try:
+        raw = base64.urlsafe_b64decode(cursor + padding).decode("utf-8")
+    except (binascii.Error, UnicodeDecodeError, ValueError) as exc:
+        raise InvalidCursorError(f"malformed cursor: {cursor!r}") from exc
+    if not raw.startswith(_CURSOR_PREFIX):
+        raise InvalidCursorError(f"malformed cursor: {cursor!r}")
+    try:
+        offset = int(raw[len(_CURSOR_PREFIX) :])
+    except ValueError as exc:
+        raise InvalidCursorError(f"malformed cursor: {cursor!r}") from exc
+    if offset < 0:
+        raise InvalidCursorError(f"malformed cursor: {cursor!r}")
+    return offset
+
+
+@dataclass(frozen=True)
+class RegisteredResource:
+    """A listing entry and the bytes it resolves to."""
+
+    resource: Resource
+    contents: TextResourceContents | BlobResourceContents
+
+
+class ResourceRegistry:
+    """URI to resource, beside the tool catalog and independent of it."""
+
+    def __init__(self, page_size: int = DEFAULT_PAGE_SIZE) -> None:
+        self._resources: dict[str, RegisteredResource] = {}
+        self._uris: list[str] = []
+        self.page_size = page_size
+
+    @property
+    def page_size(self) -> int:
+        return self._page_size
+
+    @page_size.setter
+    def page_size(self, value: int) -> None:
+        # Validated on assignment, not just in __init__, because this is a public
+        # attribute callers set directly. A non-positive size makes cursor paging
+        # non-terminating: zero hands back an empty page and the same cursor
+        # forever, and a negative one walks the offset backwards into a cursor
+        # this registry then rejects as malformed.
+        if value <= 0:
+            raise ValueError(f"page_size must be positive, got {value}")
+        self._page_size = value
+
+    def __len__(self) -> int:
+        return len(self._resources)
+
+    def __contains__(self, uri: object) -> bool:
+        return uri in self._resources
+
+    def add(self, resource: Resource, contents: str | bytes) -> RegisteredResource:
+        """Register a resource at its own URI, replacing any resource already there.
+
+        Text and binary are kept apart by type rather than by emptiness, so a
+        zero-length document and a zero-length blob stay distinguishable all the
+        way to the client.
+        """
+        if not isinstance(contents, (str, bytes)):
+            # Checked on the value rather than on the function, because any check
+            # against the function object is defeatable by a wrapper: a synchronous
+            # def returning a coroutine passes iscoroutinefunction and still lands
+            # here holding one.
+            hint = (
+                " A resource is resolved once at registration on a synchronous path,"
+                " so it cannot return a coroutine."
+                if inspect.iscoroutine(contents)
+                else ""
+            )
+            raise TypeError(
+                f"resource contents for {resource.uri} must be str or bytes, "
+                f"got {type(contents).__name__}.{hint}"
+            )
+
+        if isinstance(contents, bytes):
+            body: TextResourceContents | BlobResourceContents = BlobResourceContents(
+                uri=resource.uri,
+                mimeType=resource.mimeType,
+                blob=base64.b64encode(contents).decode("ascii"),
+            )
+        else:
+            body = TextResourceContents(
+                uri=resource.uri,
+                mimeType=resource.mimeType,
+                text=contents,
+            )
+
+        registered = RegisteredResource(resource=resource, contents=body)
+        if resource.uri not in self._resources:
+            insort(self._uris, resource.uri)
+        self._resources[resource.uri] = registered
+        return registered
+
+    def get(self, uri: str) -> RegisteredResource:
+        try:
+            return self._resources[uri]
+        except KeyError:
+            raise ResourceNotFoundError(uri) from None
+
+    def list(self, cursor: str | None = None) -> tuple[list[Resource], str | None]:
+        """Return one page of resources and the cursor for the next, if any.
+
+        A worker registers once at startup and lists on every request, so the
+        ordering cost sits on the write. ``add`` keeps ``_uris`` sorted and this
+        is a slice.
+
+        Ordering is by URI so a cursor keeps its meaning when the next page is
+        served by a different process. A worker can run as several processes
+        behind one address, and none of them shares insertion order.
+        """
+        offset = decode_cursor(cursor) if cursor else 0
+        window = self._uris[offset : offset + self.page_size]
+        page = [self._resources[uri].resource for uri in window]
+        next_offset = offset + self.page_size
+        next_cursor = encode_cursor(next_offset) if next_offset < len(self._uris) else None
+        return page, next_cursor
--- a/libs/arcade-core/arcade_core/resources.py
+++ b/libs/arcade-core/arcade_core/resources.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 import base64
 import binascii
 import inspect
-from bisect import insort
+from bisect import bisect_right, insort
 from dataclasses import dataclass
 
 from arcade_core.resource_schema import (
@@ -21,7 +21,7 @@ from arcade_core.resource_schema import (
 
 DEFAULT_PAGE_SIZE = 250
 
-_CURSOR_PREFIX = "offset:"
+_CURSOR_PREFIX = "after:"
 
 
 class InvalidCursorError(ValueError):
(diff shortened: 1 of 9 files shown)


libs/arcade-core/arcade_core/resources.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def list(self, cursor: str | None = None) -> tuple[list[Resource], str | None]:
+        """Return one page of resources and the cursor for the next, if any.
+
+        Ordering is by URI rather than by insertion, because the hosted worker
+        runs several processes from one image and a cursor issued by one of them
+        has to mean the same thing to the next.
+        """
+        offset = decode_cursor(cursor) if cursor else 0
+        uris = sorted(self._resources)
```
EricGustin: Since the engine will not store the list itself, it will be reaching out to the worker often. Therefore, listing will be much more common than adding. So, instead of sorting on every single call to `list`, we should insert at the correct location.

EricGustin: Valid, applied.

You are right about the access pattern. Nothing stores the list, so the engine refetches it per request, while `add` runs once per toolkit at startup. Paying `O(n log n)` on the read to keep the write `O(1)` is backwards.

`add` now keeps a `_uris` list ordered with `bisect.insort`, and `list` slices it:

```python
if resource.uri not in self._resources:
    insort(self._uris, resource.uri)
self._resources[resource.uri] = registered
```

```python
offset = decode_cursor(cursor) if cursor else 0
window = self._uris[offset : offset + self.page_size]
page = [self._resources[uri].resource for uri in window]
```

Read drops from `O(n log n)` to `O(page_size)`. Write goes from `O(1)` to `O(n)` for the memmove, on the path that runs at boot.

The `not in` guard is the part that needed a test. Re-registering a URI replaces in place and must not insert a second time, and `len(registry)` reads the dict, so it would have stayed `1` with a duplicate sitting in the index. `test_re_registering_a_uri_replaces_it` now also asserts the listing comes back with one entry, which is the only place that duplicate would surface.

libs/arcade-core/arcade_core/resources.py
```diff
@@ -0,0 +1,127 @@
...
+    contents: TextResourceContents | BlobResourceContents
+
+
+class ResourceRegistry:
+    """URI to resource, beside the tool catalog and independent of it."""
+
+    def __init__(self, page_size: int = DEFAULT_PAGE_SIZE) -> None:
+        self._resources: dict[str, RegisteredResource] = {}
+        self._uris: list[str] = []
+        self.page_size = page_size
```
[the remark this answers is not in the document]
EricGustin: Valid, and reproduced. Both failure modes are real:

```
page_size=0  -> page=0 cursor='b2Zmc2V0OjA'; next round page=0 cursor='b2Zmc2V0OjA'
page_size=-1 -> page=1 cursor='b2Zmc2V0Oi0x'
               feeding that cursor straight back: InvalidCursorError: malformed cursor
```

The negative case is a little worse than described. The registry does not just walk backwards, it emits a cursor that it then rejects as malformed, because `decode_cursor` refuses a negative offset.

I did not take the suggested patch as written, because validating only in `__init__` leaves the hole open. `page_size` is a public attribute and callers assign to it directly (the worker tests do exactly that to force pagination), so `registry.page_size = 0` would still get through. It is a validating property instead, so construction and assignment share one rule:

```python
@page_size.setter
def page_size(self, value: int) -> None:
    if value <= 0:
        raise ValueError(f"page_size must be positive, got {value}")
    self._page_size = value
```

The public name is unchanged, so nothing that sets `page_size` needs to move. Test covers both entry points, parameterized over `0` and `-1`.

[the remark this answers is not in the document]
EricGustin: Already on the branch, and as the property setter rather than an `__init__` check: `page_size` is a public attribute callers assign directly, so validating only in the constructor would leave the direct-assignment path open. `resources.py:78-87` raises `ValueError` on any non-positive value from either entry point, and `test_a_nonpositive_page_size_is_refused` is parameterized over `0` and `-1` against both.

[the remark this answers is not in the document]
EricGustin: Agreed, and that is what shipped. While here I also closed the last coverage gaps in `resources.py`: a cursor whose offset is not an integer, `__contains__`, and `add()` rejecting contents that are neither `str` nor `bytes` including the coroutine case. The file is at 100% (was 93%, 5 lines missing).

libs/arcade-core/arcade_core/resources.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Ordering is by URI so a cursor keeps its meaning when the next page is
+        served by a different process. A worker can run as several processes
+        behind one address, and none of them shares insertion order.
+        """
+        offset = decode_cursor(cursor) if cursor else 0
+        window = self._uris[offset : offset + self.page_size]
+        page = [self._resources[uri].resource for uri in window]
+        next_offset = offset + self.page_size
+        next_cursor = encode_cursor(next_offset) if next_offset < len(self._uris) else None
```
byrro: Ordering by URI is supposed to make this cursor mean the same thing on another process, but we encode an index into `_uris`, not a URI. That only works if both lists are identical. Once URIs include the toolkit version, a rolling deploy gives one replica `ui://Math/1.1.0/...` and another `ui://Math/1.0.0/...`, so the same offset can skip, duplicate, or return `([], None)` as if we were done.

I wonder if we should put the last served URI in the cursor and resume with `bisect_right`, so we don't skip or repeat an entry both replicas hold.

Although it seems worth doing, honestly that still can't give a consistent snapshot across disjoint version sets: a cursor after `ui://Math/1.1.0/b` on a 1.0.0 replica is past the end and looks finished. If we want that to error instead, the cursor needs a registry identity we reject on mismatch.

libs/arcade-core/arcade_core/resources.py
```diff
@@ -0,0 +1,159 @@
...
+                blob=base64.b64encode(contents).decode("ascii"),
+            )
+        else:
+            body = TextResourceContents(
+                uri=resource.uri,
+                mimeType=resource.mimeType,
+                text=contents,
+            )
+
+        registered = RegisteredResource(resource=resource, contents=body)
```
byrro, line 139: `add` keeps the caller's `Resource` by reference, and `list`/`get`/`add`'s returned `RegisteredResource` all expose that same object. Mutating `.uri` on it desyncs the index from the dict: `list` advertises the new URI, `get` of that URI raises `ResourceNotFoundError`, and `get` of the old key returns the lie.

I wonder if we should use `model_copy(deep=True)` on the way in and out so the registry owns what it serves

libs/arcade-core/arcade_core/resources.py
```diff
@@ -0,0 +1,159 @@
...
+                uri=resource.uri,
+                mimeType=resource.mimeType,
+                blob=base64.b64encode(contents).decode("ascii"),
+            )
+        else:
+            body = TextResourceContents(
+                uri=resource.uri,
+                mimeType=resource.mimeType,
+                text=contents,
+            )
```
byrro, line 137: `add` builds the `resources/read` body from `uri`, `mimeType`, and the payload only, so `contents.meta` is always `None`. Listing `Resource.meta` is a different slot; MCP Apps puts `ui.csp` / `ui.domain` / `ui.permissions` on `contents[]`.

The following PRs `#921` and `#922` serialize what we stored, so there is no later fill-in, and a UI that fetches or loads remote assets will read fine while the host default-deny CSP blocks the calls.

I think we should take content `_meta` as an explicit `add` input (a `contents_meta` kwarg, or a prebuilt `TextResourceContents` / `BlobResourceContents`) and not copy it from the listing object.

byrro (approved):
three low-priority comments; I'm approving, but happy to re-approve if you deem worth addressing the comments in this PR

libs/arcade-core/arcade_core/resources.py
EricGustin: Addressing this will be a part of M2 (TOO-1943)

libs/arcade-core/arcade_core/resources.py
EricGustin: Addressed in `efde4205`. You were right that the docstring promised something the code did not do, and I took the fix you suggested.

The cursor now names the last URI served and the resume point is a `bisect_right`. A URI is the same position in any replica that has it, and a resume point in any replica that does not, which is why the anchor is searched for rather than required.

`task_manager.py` already does `_encode_cursor(page[-1])`, so this is the house pattern rather than a new one. It differs deliberately on the miss: the task manager raises when a cursor's anchor is gone, because a task that no longer exists really is an error, while a resource URI absent from one replica is the ordinary mid-rollout state and has to resume rather than fail.

Three tests for what an offset could not do, each verified to fail against the old encoding:

- two replicas with different catalogs handing a page between them (an offset of 2 into the second one re-served `ui://B` and never reached `ui://C`)
- a cursor naming a URI the serving replica does not hold
- an entry inserted ahead of the cursor

Your last paragraph is the part I have not done, and I think it needs a decision rather than a patch: a cursor sorting past everything a replica holds still reads as a finished listing, and telling that apart from a genuine last page needs a registry identity in the cursor that a mismatch rejects. Filing it rather than guessing at the identity.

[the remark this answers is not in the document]
EricGustin: Confirmed, and fixed in `efde4205`. Same finding as byrro's on this line, and the two together are what moved it up the list.

`list` now pages by the last URI served rather than an index into the local `_uris`, resolved with `bisect_right`, so a replica holding a different set resumes after where that URI would sort instead of at an unrelated position. Three tests cover it, each verified to fail against the offset encoding.

libs/arcade-core/arcade_core/resources.py
EricGustin: Reproduced exactly as you describe: mutate a returned `.uri` and `list()` advertises a URI `get()` raises `ResourceNotFoundError` for, while `get()` of the old key returns the object now claiming the new one.

Not fixing it here, and I want to give the reasoning rather than just defer.

There is no reachable trigger. On this PR there are **zero** production callers of `add()`, and once #919 lands `declare()` builds the `Resource` itself, so the toolkit path never hands the registry an object anyone else is holding. The realistic direction is the other one, a consumer mutating what it got back, and that is already governed by an explicit convention in the only consumer: `_project_for_version` in `server.py` says "NEVER MUTATE cached DTOs — always dump first" and dumps.

`model_copy(deep=True)` on the way out also puts a copy of the **contents body**, which is the whole document, on every read, to defend a property the codebase already forbids by convention.

I think the better question is whether to freeze the model so mutation raises rather than copy so it survives, and that is a decision rather than a patch. Filing it with your analysis so it gets one.

libs/arcade-core/arcade_core/resources.py
EricGustin: Confirmed: a resource declared with `_meta={"ui": {...}}` registers with `Resource.meta` set and `contents.meta` `None`. You are right that MCP Apps reads `ui.csp` off `contents[]`, and right that #921 and #922 serialise what was stored so there is no later fill-in.

Not doing it in this PR, for a reason I want to separate from the cursor one above, because on the surface they look alike.

A cursor is a wire format the engine round-trips opaquely, so changing its meaning after M1 ships is a compatibility event during exactly the rolling deploy it exists to survive. That is why it had to happen now. Adding `_meta` to `contents[]` later is **additive**: no reader breaks and there is no migration. The cost of waiting is bounded in a way the cursor's was not.

And the shape of the input, a `contents_meta` kwarg versus accepting a prebuilt `TextResourceContents`, is the authoring-surface question that TOO-1943 owns ("let a resource declare its own `_meta.ui`", already scheduled for M2). Picking it here in the registry PR pre-empts that ticket for no compatibility gain.

I have added your analysis to TOO-1943 so whoever picks it up starts from the registry half rather than rediscovering it.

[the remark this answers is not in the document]
EricGustin: Confirmed, and the same call as on byrro's comment on this line: real, and deferred to TOO-1943 rather than done here.

The short version: adding `_meta` to `contents[]` later is additive on the wire, so unlike the cursor on line 158 there is no compatibility event in waiting. And the shape of the input is the authoring-surface decision TOO-1943 owns. Your analysis is on that ticket.

PR #919, opened 2026-08-25
feat(core): qualify resource URIs by toolkit and version [TOO-1933]

EricGustin:
## Summary

A resource a toolkit declares gets registered at `ui://<Toolkit>/<version>/<path>`. The author writes the path. The URI is derived at registration and never typed by hand.

Part 3 of 6. Stacked on #918.

Resolves: https://linear.app/arcadedev/issue/TOO-1933

## Design decisions

Two things can collide, and each segment closes one of them.

One worker can host many toolkits, so two of them could otherwise both claim `ui://draft-review.html`. The toolkit segment settles that, and with it in place no cross-toolkit collision is possible at all.

The version segment handles a subtler case. The same toolkit can be installed at two versions across two workers. Strip the version and a v8 tool and a v7 tool both resolve to `ui://Gmail/draft.html`, so a gateway that picks the wrong worker renders a v7 interface for a v8 tool and nothing anywhere reports an error. That failure is silent, and it surfaces to whoever hits it as a rendering bug with no obvious cause. Worth a path segment.

The scheme is carried through and never replaced. A host that renders a tool's interface requires `ui://` and throws on anything else, so a qualifier that swaps the prefix (`arcade://source/ui://...`) breaks rendering outright. There is a test pinning that.

Qualification happens inside the registry, since that is the only place a declaration and the toolkit's identity are both in scope. Traversing paths (`../`) get refused instead of normalized away.

## Scope

In scope: making a URI unique within one worker.

Not in scope: making it unique across backing servers. That belongs to whatever serves these onward. One note for whoever builds it: a qualifier at that level has to apply unconditionally rather than skipping URIs that already look qualified, because a remote MCP server emits this same authority-segment shape and no amount of inspection tells the two apart. Two qualifiers at two levels nest without trouble.

## Test plan

- [x] 20 new tests in `libs/tests/core/test_resource_uri_qualification.py`

## Author checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how resource URIs are formed and validated at registration; mistakes could break UI resolution or caching, though existing `add()` behavior is unchanged and invalid paths fail closed.
> 
> **Overview**
> Toolkit static resources no longer register at author-supplied URIs alone. Authors declare a **path** (and metadata); **`ResourceRegistry.declare`** derives **`ui://<Toolkit>/<version>/<path>`** via new **`qualify`**, using toolkit name/version only at registration time.
> 
> **`qualify`** validates scheme, toolkit, and version (rejecting values that would hijack authority or collide across segments), normalizes slashy paths, percent-encodes path segments, blocks `.`/`..` traversal and control characters, and preserves the scheme (default **`UI_SCHEME`** = `ui`) without prefix mangling. **`ResourceDeclaration`** captures what authors write; **`InvalidResourcePathError`** surfaces bad inputs.
> 
> **`arcade-core`** bumps to **4.14.0**. New tests in **`test_resource_uri_qualification.py`** cover parsing, collisions, PEP 440-style versions, and registry **`declare`** wiring so listing and content URIs match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbd3fe521a5ee0579e31fdc2546b677afc128483. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-core/arcade_core/resources.py
+++ b/libs/arcade-core/arcade_core/resources.py
@@ -89,6 +89,21 @@ def qualify(toolkit_name: str, toolkit_version: str, path: str, scheme: str = UI
     if any(segment in (".", "..") for segment in segments):
         raise InvalidResourcePathError(f"a resource path may not traverse: {path!r}")
 
+    # A path may not carry a character that changes what the URI means or stops
+    # it being one. "?" opens a query and "#" opens a fragment, so either would
+    # leave the resource registered under a string whose path component is
+    # shorter than the author wrote; a control character makes the URI fail to
+    # parse at all. Both fail at the point of reading, far from the declaration.
+    for char in path:
+        if char in "?#":
+            raise InvalidResourcePathError(
+                f"a resource path may not contain {char!r}, which starts a new URI component: {path!r}"
+            )
+        if not char.isprintable():
+            raise InvalidResourcePathError(
+                f"a resource path may not contain a control character: {path!r}"
+            )
+
     return f"{scheme}://{toolkit_name}/{toolkit_version}/{'/'.join(segments)}"
 
 
--- a/libs/arcade-core/arcade_core/resources.py
+++ b/libs/arcade-core/arcade_core/resources.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 import base64
 import binascii
 import inspect
-from bisect import insort
+from bisect import bisect_right, insort
 from dataclasses import dataclass
 
 from arcade_core.resource_schema import (
@@ -21,7 +21,7 @@ from arcade_core.resource_schema import (
 
 DEFAULT_PAGE_SIZE = 250
 
-_CURSOR_PREFIX = "offset:"
+_CURSOR_PREFIX = "after:"
 
 #: The scheme a tool's user interface is addressed under. Hosts that render an
 #: interface require it and refuse anything else.
@@ -36,13 +36,24 @@ class ResourceNotFoundError(KeyError):
     """Raised when no resource is registered at the requested URI."""
 
 
-def encode_cursor(offset: int) -> str:
-    """Cursors are opaque to callers, so the encoding can change freely."""
-    raw = f"{_CURSOR_PREFIX}{offset}".encode()
+def encode_cursor(last_uri: str) -> str:
+    """A cursor names the last URI served, not how many were served before it.
+
+    A worker runs as several processes and a page can be served by a different
+    one than issued the cursor. An index is only the same position in both when
+    both hold the same URIs, and they do not: a toolkit version lives in the
+    URI, so mid-rollout one replica holds ``ui://Math/1.1.0/x`` where another
+    holds ``ui://Math/1.0.0/x``. The same offset then skips, repeats, or ends
+    the listing early. A URI is the same position in any replica that has it,
+    and a resume point in any replica that does not.
+
+    Opaque to callers, so the encoding can still change freely.
+    """
+    raw = f"{_CURSOR_PREFIX}{last_uri}".encode()
     return base64.urlsafe_b64encode(raw).decode("ascii").rstrip("=")
 
 
-def decode_cursor(cursor: str) -> int:
+def decode_cursor(cursor: str) -> str:
     padding = "=" * (-len(cursor) % 4)
     try:
         raw = base64.urlsafe_b64decode(cursor + padding).decode("utf-8")
@@ -50,13 +61,10 @@ def decode_cursor(cursor: str) -> int:
         raise InvalidCursorError(f"malformed cursor: {cursor!r}") from exc
     if not raw.startswith(_CURSOR_PREFIX):
         raise InvalidCursorError(f"malformed cursor: {cursor!r}")
-    try:
-        offset = int(raw[len(_CURSOR_PREFIX) :])
-    except ValueError as exc:
-        raise InvalidCursorError(f"malformed cursor: {cursor!r}") from exc
-    if offset < 0:
+    last_uri = raw[len(_CURSOR_PREFIX) :]
+    if not last_uri:
         raise InvalidCursorError(f"malformed cursor: {cursor!r}")
-    return offset
+    return last_uri
 
 
 class InvalidResourcePathError(ValueError):
@@ -237,10 +245,14 @@ class ResourceRegistry:
         Ordering is by URI so a cursor keeps its meaning when the next page is
         served by a different process. A worker can run as several processes
         behind one address, and none of them shares insertion order.
+
+        The cursor names a URI and the resume point is found with a binary
+        search, so a replica that does not hold that exact URI still resumes
+        after where it would sort rather than at some unrelated index.
         """
-        offset = decode_cursor(cursor) if cursor else 0
-        window = self._uris[offset : offset + self.page_size]
+        start = bisect_right(self._uris, decode_cursor(cursor)) if cursor else 0
+        end = start + self.page_size
+        window = self._uris[start:end]
         page = [self._resources[uri].resource for uri in window]
-        next_offset = offset + self.page_size
-        next_cursor = encode_cursor(next_offset) if next_offset < len(self._uris) else None
+        next_cursor = encode_cursor(window[-1]) if window and end < len(self._uris) else None
         return page, next_cursor
--- a/libs/arcade-core/arcade_core/resources.py
+++ b/libs/arcade-core/arcade_core/resources.py
@@ -23,6 +23,10 @@ DEFAULT_PAGE_SIZE = 250
 
 _CURSOR_PREFIX = "offset:"
 
+#: The scheme a tool's user interface is addressed under. Hosts that render an
+#: interface require it and refuse anything else.
+UI_SCHEME = "ui"
+
 
 class InvalidCursorError(ValueError):
     """Raised when a caller sends a cursor this registry did not issue."""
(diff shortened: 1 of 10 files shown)


libs/arcade-core/arcade_core/resources.py
```diff
@@ -55,6 +59,66 @@ def decode_cursor(cursor: str) -> int:
...
+    if not toolkit_name:
+        raise InvalidResourcePathError("a resource URI needs a toolkit name")
+    if not toolkit_version:
+        raise InvalidResourcePathError("a resource URI needs a toolkit version")
+
+    segments = [segment for segment in path.split("/") if segment]
+    if not segments:
+        raise InvalidResourcePathError(f"a resource needs a path: {path!r}")
+    if any(segment in (".", "..") for segment in segments):
+        raise InvalidResourcePathError(f"a resource path may not traverse: {path!r}")
```
byrro, line 123: This check is literal, so `%2e%2e` is not `..` and we accept it.

`qualify("Gmail", "8.1.0", "%2e%2e/draft.html")` becomes `ui://Gmail/8.1.0/%2e%2e/draft.html`; a WHATWG host collapses that to `ui://Gmail/draft.html`, the same URI at any version. Three `%2e%2e` also climb out of the toolkit segment.

Although the path is toolkit-author input and unlikely to contain this problematic format, for the sake of completeness / correctness, I think it'd be better encode each segment with `quote(segment, safe="")`, or reject `%` outright. If we encode, the `endswith(path)` tests need to assert an `unquote` round trip instead.

libs/arcade-core/arcade_core/resources.py
```diff
@@ -55,6 +59,66 @@ def decode_cursor(cursor: str) -> int:
...
+    # parse at all. Both fail at the point of reading, far from the declaration.
+    for char in path:
+        if char in "?#":
+            raise InvalidResourcePathError(
+                f"a resource path may not contain {char!r}, which starts a new URI component: {path!r}"
+            )
+        if not char.isprintable():
+            raise InvalidResourcePathError(
+                f"a resource path may not contain a control character: {path!r}"
+            )
```
byrro: This loop refuses `?`, `#`, and controls, but not `%`, space, or non-ASCII. `qualify(..., "50%.html")` is a string Go `net/url` rejects (`invalid URL escape "%.h"`), so the engine cannot parse a URI the worker just listed. `draft review.html` and `café.html` (btw, my favorite morning drink) parse, then become `draft%20review.html` / `caf%C3%A9.html` on any normalize.

If that normalized URI is sent back, exact `ResourceRegistry.get` returns `ResourceNotFoundError`. `#921` reads through that lookup. `urlparse` in `test_a_qualified_uri_parses_as_a_uri` never rejects or re-encodes, so it cannot catch this.

Per-segment `quote(segment, safe="")` (same change as the `%2e%2e` case) makes all three byte-stable. The `endswith(path)` / `café.html` tests should assert an `unquote` round trip, not that the raw path is still in the URI.

libs/arcade-core/arcade_core/resources.py
```diff
@@ -55,6 +59,66 @@ def decode_cursor(cursor: str) -> int:
...
+    tool's interface requires ``ui://`` and throws on anything else, so a
+    prefix-replacing qualifier breaks rendering outright.
+    """
+    scheme = scheme.rstrip(":/")
+    if not scheme:
+        raise InvalidResourcePathError("a resource URI needs a scheme")
+    if not toolkit_name:
+        raise InvalidResourcePathError("a resource URI needs a toolkit name")
+    if not toolkit_version:
+        raise InvalidResourcePathError("a resource URI needs a toolkit version")
```
byrro: Name and version are only checked for empty, then spliced in: name as the authority, version as the first path segment. It definitely looks like a name that we shouldn't expect, but `My Toolkit` makes Go and WHATWG refuse the URI (`invalid character " " in host name` / `Invalid URL`). `Gmail?v2` drops the path. `qualify("Gmail", "8.1.0/x", "a.html")` equals `qualify("Gmail", "8.1.0", "x/a.html")`. 
PR #920  will call this with `Toolkit.name` from `from_entrypoint`, which is the raw `entry.value`.

Just for peace of mind, I think it'd be better to reject a toolkit name that is not on a narrow ASCII unreserved allowlist at boot, and encode the version the same way as the path segments.

libs/arcade-core/arcade_core/resources.py
```diff
@@ -55,6 +59,66 @@ def decode_cursor(cursor: str) -> int:
...
+    same toolkit installed at two versions across two workers, which is what
+    keeps a tool and its interface in agreement.
+
+    The scheme is carried through and never replaced. A host that renders a
+    tool's interface requires ``ui://`` and throws on anything else, so a
+    prefix-replacing qualifier breaks rendering outright.
+    """
+    scheme = scheme.rstrip(":/")
+    if not scheme:
+        raise InvalidResourcePathError("a resource URI needs a scheme")
```
byrro: `scheme` is only `rstrip(":/")` then placed in front of `://`.

The string `scheme="ui://Slack/9.0.0"` parses as host `Slack` in both Go and WHATWG, so a Gmail declaration is served under another toolkit's authority.

Malformed strings like `"u?i"`, `"1ui"` / `"ht tp"` will make `new URL()` throw; `test_the_scheme_is_carried_through_and_never_replaced` only checks that the `"ui"` default still starts with `ui://`.

We should reject anything that is not an RFC 3986 scheme (`[A-Za-z][A-Za-z0-9+.-]*` after the rstrip), or drop `scheme` from `ResourceDeclaration` and keep a small closed set (`ui`, `docs`).

byrro (approved):
Left a few comments as optional follow-ups. They're real correctness gaps, but there's arguably a low likelihood that they'll ever manifest, so up to you whether it's worth addressing them here

libs/arcade-core/arcade_core/resources.py
EricGustin: Confirmed and fixed in `f94d73b3`. Reproduced first, and Go agrees with your reading:

```
qualify("Gmail", "8.1.0", "draft.html", scheme="ui://Slack/9.0.0")
  -> ui://Slack/9.0.0://Gmail/8.1.0/draft.html
net/url                                        -> host="Slack"
```

`scheme` is now RFC 3986, `[A-Za-z][A-Za-z0-9+.-]*` after the rstrip, which also refuses `1ui` and `ht tp`. I kept it a pattern rather than the closed `{ui, docs}` set you offered as the alternative, because D-level decisions here treat the scheme as carried through and never replaced, and a closed set would quietly become the place new schemes have to be registered.

libs/arcade-core/arcade_core/resources.py
EricGustin: Confirmed, all three, and the version one is the sharpest:

```
qualify("My Toolkit", ...)  -> net/url: invalid character " " in host name
qualify("Gmail?v2", ...)    -> host="Gmail", path=""        the whole path is gone
qualify("Gmail", "8.1.0/x", "a.html") == qualify("Gmail", "8.1.0", "x/a.html")
```

Both are now held to RFC 3986 unreserved, which is what `normalize_toolkit_name` already produces for every real toolkit. `test_a_version_cannot_eat_a_path_segment` pins the collision as the equality it used to satisfy.

Refused rather than encoded, deliberately, and that is the split that resolved all four of your comments on this PR: a **path** is this resource's own name, so it is encoded; the **scheme, toolkit and version** are the URI's identity, so they are refused. Encoding an identity would answer under a name nobody asked for.

libs/arcade-core/arcade_core/resources.py
EricGustin: Confirmed. Go decodes it exactly as you said:

```
ui://Gmail/8.1.0/%2e%2e/draft.html  ->  path "/8.1.0/../draft.html"
```

Took the encode option. Each segment now goes through `quote(segment, safe="")`, so a literal `%2e%2e` becomes `%252e%252e`, which decodes back to a segment *named* `%2e%2e` rather than to `..`. The traversal check still runs on the raw path first and still catches a literal `..`, so both spellings are covered.

libs/arcade-core/arcade_core/resources.py
EricGustin: Right on all three, and you were right that the loop I had added was the wrong shape. Fixed in `f94d73b3` with the per-segment `quote(segment, safe="")` you suggested.

```
50%.html            -> ui://Gmail/8.1.0/50%25.html
a b.html            -> ui://Gmail/8.1.0/a%20b.html
café.html           -> ui://Gmail/8.1.0/caf%C3%A9.html
report?v=2.html     -> ui://Gmail/8.1.0/report%3Fv%3D2.html
```

All eight forms now parse in Go, keep host `Gmail`, and **round-trip byte for byte**, decoding back to the path the author declared.

That subsumes the `?`/`#`/control-character loop, so it is gone. Enumerating the characters that break a URI was the wrong shape twice over: `quote` covers every one of them and cannot miss the next. The one enumeration left is control characters, kept because they encode cleanly but are never what an author meant, so refusing them catches corruption rather than a preference.

And the test changed the way you said it would need to. It now asserts `urlparse`/`urlunparse` leaves the URI unchanged and `unquote` gives back the declared path. The old `endswith(path)` assertion passed happily on `50%.html`, which is a URI no parser accepts.

libs/arcade-core/arcade_core/resources.py
```diff
@@ -63,6 +69,75 @@ def decode_cursor(cursor: str) -> str:
...
 
 
+#: RFC 3986 scheme. Without this, ``scheme="ui://Slack/9.0.0"`` parses as the
+#: host ``Slack`` and one toolkit answers under another toolkit's authority.
+_SCHEME = re.compile(r"[A-Za-z][A-Za-z0-9+.-]*\Z")
+
+#: RFC 3986 unreserved. The toolkit and version are spliced in as the authority
+#: and the first path segment, so a "/" in a version makes it eat a path
+#: segment and two different declarations collide on one URI.
+_IDENTITY = re.compile(r"[A-Za-z0-9._~-]+\Z")
```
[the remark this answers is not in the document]
EricGustin: Correct, and this was a regression I introduced with the identity allowlist minutes earlier. Verified before fixing:

```
ToolkitDefinition(version="1.0.0+build.5")  -> accepted
qualify("Gmail", "1.0.0+build.5", "a.html") -> InvalidResourcePathError
```

So a toolkit shipping semver build metadata or a PEP 440 local version would have loaded its tools and dropped every resource it declares. That is the same failure mode this stack has been closing everywhere else, reintroduced by a narrower allowlist than the grammar needs.

Fixed in `bbd3fe52`. The pattern is now unreserved plus `+` and `!`, the two sub-delims a real version can carry: `+` for semver build metadata and PEP 440 local versions, `!` for a PEP 440 epoch. Checked against `net/url` first, because the engine is what parses these: every shape those two grammars produce parses, keeps its host, and round-trips byte for byte.

```
1.0.0a1  1.0.0rc1  1.0.0.post1  1.0.0.dev0
1.0.0+build.5  1.0.0+local_abc  1.0.0-rc.1  1!1.0.0
```

Left literal rather than encoded, since `1.0.0%2Bbuild.5` and `1.0.0+build.5` both round-trip and decode alike, so encoding would give one resource two spellings.

The new parametrize is that list of versions, so the next narrowing of this pattern has to explain itself against them. `8.1.0/x` is still refused, which is the collision the allowlist exists for.

PR #922, opened 2026-08-25
feat(serve): serve POST /worker/resources/list and /worker/resources/read [TOO-1936]

EricGustin:
## Summary

The two resource endpoints, `POST /worker/resources/list` and `POST /worker/resources/read`, registered in the default component set. One rule governs both: the path suffix names the operation, the request body is that operation's params object, and the response is its result object, all spelled the way they go on the wire so neither end has to translate anything.

Part 6 of 6. Stacked on #921. This is the whole protocol addition.

Resolves: https://linear.app/arcadedev/issue/TOO-1936

### Both are POST

Both requests carry params: an optional `cursor` on list, a required `uri` on read. And `RequestData` carries only `path`, `method` and `body`, with `path` coming from `request.url.path`, which Starlette has already stripped of the query string. A worker component genuinely cannot read a query parameter. A first-page list sends an empty body, which the router already coerces to `{}`.

### `response_model_exclude_none=True` gets passed through

FastAPI defaults it to `False`, so without it a resource with no annotations goes out carrying `"annotations": null, "size": null, "icons": null`. Those keys should be absent.

### An unknown URI is 404 with code `-32002`

Which overlaps awkwardly with reading a 404 as "this worker has no resource endpoints at all", so the tolerance rule is per-endpoint: 404 and 405 mean an empty catalog on the list endpoint only. On read, a 404 is a typed not-found. That leaves one ambiguous case, an old worker 404ing a read, and it is harmless, because the caller does not get the resource either way.

### A malformed body is 400, not 500

The body was parsed and `RequestData` built *above* the block that maps errors, so a body that is not a JSON object escaped as an unhandled 500. `RequestData.body_json` is `dict[str, Any] | None`, so `[]`, `false`, `0`, `""` and truncated JSON all took that path. A caller sending nonsense was told the worker had broken, which matters because the engine reads a 500 as a worker fault worth retrying and a 400 as its own bug. The parse moved inside the mapped block.

This one is wider than the rest of the PR and worth a reviewer's attention: `_wrap_handler` is shared, so `/worker/tools` and `/worker/tools/invoke` move from 500 to 400 on a malformed body too. That behavior is pre-existing and identical on `main`. It is fixed here because this PR is what adds the `ValidationError` to 400 mapping in this file, and that mapping was dead for the most likely error. A literal `null` body is left reading as empty params, since the router already maps an absent body to `{}`.

The HTTP mapping lives in the FastAPI binding, which keeps `core` framework-agnostic. Components raise typed errors and the binding turns them into an error object carried by a status code.

## Test plan

Exercised against a real uvicorn worker with a real toolkit installed, nothing in front of it:

```
$ curl -sS -X POST localhost:8931/worker/resources/list -H "Authorization: Bearer $TOKEN" -d "{}"
{"resources":[{"name":"draft_review","uri":"ui://demokit/3.2.1/draft-review.html","mimeType":"text/html;profile=example"}]}

$ curl -sS -X POST localhost:8931/worker/resources/read -H "Authorization: Bearer $TOKEN" -d "{\"uri\":\"ui://demokit/3.2.1/draft-review.html\"}"
{"contents":[{"uri":"ui://demokit/3.2.1/draft-review.html","mimeType":"text/html;profile=example","text":"<!DOCTYPE html>..."}]}

# unknown URI
404 {"code":-32002,"message":"Resource not found: ui://demokit/3.2.1/nope.html"}

# no Authorization header
401
```

The mime type is whatever the toolkit declared. These libraries carry it verbatim and never inspect it.

- [x] 19 contract tests asserting the wire bytes instead of the Python objects, including one checking that a parameterized mime type survives byte-identically. That one reads the raw response body, since a stray space after the semicolon survives JSON parsing and would still stop a host rendering.

## Risk note

Adds two routes under `/worker/*`, which this repo's PR template lists as sensitive. Both sit behind the same `validate_engine_request` dependency as the existing endpoints, and no auth path changes. A worker with no declared resources answers list with an empty array.

The malformed-body change above touches `_wrap_handler`, which every worker endpoint goes through. The blast radius is the status code a caller sees for a body that was already failing: 500 becomes 400 on `/worker/tools` and `/worker/tools/invoke` as well. No well-formed request changes behavior, and there is a test pinning each status.

## Reviewer note

The default-components test pinned the three existing routes by name and now pins five.

## Author checklist

---

> [!NOTE]
> **Medium Risk**
> New authenticated `/worker/*` routes and shared `_wrap_handler` behavior change malformed-body responses from 500 to 400 on existing tool endpoints; well-formed traffic is unchanged.
> 
> **Overview**
> Adds **POST `/worker/resources/list`** and **POST `/worker/resources/read`** to the default worker routes, wired through new `ListResourcesComponent` and `ReadResourceComponent` that use the catalog’s resource APIs with params/result shapes on the wire and **`response_model_exclude_none=True`** so optional fields stay absent instead of `null`.
> 
> Introduces **`InvalidRequestParamsError`** and **`_caller_params`** so bad caller bodies map to **400 / `-32602`**, while worker-internal validation failures stay **500**. The FastAPI **`_wrap_handler`** now parses JSON inside the error-mapping path (malformed/non-object bodies → 400 for **all** worker POST routes, including **`/worker/tools/invoke`**), maps unknown URIs to **404 / `-32002`**, and treats invalid cursors as invalid params. List/read spans record **`outcome`** without marking routine not-found or bad-cursor cases as failed traces.
> 
> Bumps **`arcade-serve` to 3.6.0** and **`arcade-mcp-server` to 1.28.1** (depends on `arcade-serve>=3.6.0`). Adds contract tests for wire bytes, auth, pagination, and span behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62fd6741fd43dd488a6942925d6b00fe0c83f6ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-serve/arcade_serve/core/components.py
+++ b/libs/arcade-serve/arcade_serve/core/components.py
@@ -1,3 +1,5 @@
+from typing import Any, TypeVar
+
 from arcade_core.log_extras import build_tool_error_span_attributes
 from arcade_core.resource_schema import (
     ListResourcesParams,
@@ -5,19 +7,33 @@ from arcade_core.resource_schema import (
     ReadResourceParams,
     ReadResourceResult,
 )
+from arcade_core.resources import InvalidCursorError, ResourceNotFoundError
 from arcade_core.schema import (
     ToolCallRequest,
     ToolCallResponse,
 )
 from opentelemetry import trace
+from opentelemetry.trace import Status, StatusCode
+from pydantic import BaseModel, ValidationError
 
 from arcade_serve.core.common import (
     CatalogResponse,
     HealthCheckResponse,
+    InvalidRequestParamsError,
     RequestData,
     Router,
     WorkerComponent,
 )
+
+_P = TypeVar("_P", bound=BaseModel)
+
+
+def _caller_params(model: type[_P], body: Any) -> _P:
+    """Build an endpoint's params from the caller's body."""
+    try:
+        return model.model_validate(body or {})
+    except ValidationError as e:
+        raise InvalidRequestParamsError(f"{e.error_count()} error(s)") from e
 
 
 class CatalogComponent(WorkerComponent):
@@ -68,7 +84,7 @@ class CallToolComponent(WorkerComponent):
         tracer = trace.get_tracer(__name__)
         with tracer.start_as_current_span("CallTool") as current_span:
             call_tool_request_data = request.body_json
-            call_tool_request = ToolCallRequest.model_validate(call_tool_request_data)
+            call_tool_request = _caller_params(ToolCallRequest, call_tool_request_data)
 
             current_span.set_attribute("tool_name", str(call_tool_request.tool.name))
             current_span.set_attribute("toolkit_version", str(call_tool_request.tool.version))
@@ -133,9 +149,23 @@ class ListResourcesComponent(WorkerComponent):
         Handle the request to list resources.
         """
         tracer = trace.get_tracer(__name__)
-        with tracer.start_as_current_span("ListResources"):
-            params = ListResourcesParams.model_validate(request.body_json or {})
-            return self.worker.list_resources(params.cursor)
+        # A refused cursor is an answer this endpoint gives, so it must not
+        # land as a failed span.
+        with tracer.start_as_current_span(
+            "ListResources", record_exception=False, set_status_on_exception=False
+        ) as current_span:
+            try:
+                params = _caller_params(ListResourcesParams, request.body_json)
+                result = self.worker.list_resources(params.cursor)
+            except (InvalidCursorError, InvalidRequestParamsError):
+                current_span.set_attribute("outcome", "invalid_request")
+                raise
+            except Exception as e:
+                current_span.record_exception(e)
+                current_span.set_status(Status(StatusCode.ERROR))
+                raise
+            current_span.set_attribute("outcome", "success")
+            return result
 
 
 class ReadResourceComponent(WorkerComponent):
@@ -160,7 +190,25 @@ class ReadResourceComponent(WorkerComponent):
         Handle the request to read a resource.
         """
         tracer = trace.get_tracer(__name__)
-        with tracer.start_as_current_span("ReadResource") as current_span:
-            params = ReadResourceParams.model_validate(request.body_json or {})
-            current_span.set_attribute("resource_uri", params.uri)
-            return self.worker.read_resource(params.uri)
+        # A URI this worker does not serve is a routine answer. Left to the
+        # default, every render of a stale interface arrives as an error span
+        # carrying a stack trace.
+        with tracer.start_as_current_span(
+            "ReadResource", record_exception=False, set_status_on_exception=False
+        ) as current_span:
+            try:
+                params = _caller_params(ReadResourceParams, request.body_json)
+                current_span.set_attribute("resource_uri", params.uri)
+                result = self.worker.read_resource(params.uri)
+            except ResourceNotFoundError:
+                current_span.set_attribute("outcome", "not_found")
+                raise
+            except InvalidRequestParamsError:
+                current_span.set_attribute("outcome", "invalid_request")
+                raise
+            except Exception as e:
+                current_span.record_exception(e)
+                current_span.set_status(Status(StatusCode.ERROR))
+                raise
+            current_span.set_attribute("outcome", "success")
+            return result
--- a/libs/arcade-serve/arcade_serve/core/components.py
+++ b/libs/arcade-serve/arcade_serve/core/components.py
@@ -1,4 +1,10 @@
 from arcade_core.log_extras import build_tool_error_span_attributes
+from arcade_core.resource_schema import (
+    ListResourcesParams,
+    ListResourcesResult,
+    ReadResourceParams,
+    ReadResourceResult,
+)
 from arcade_core.schema import (
     ToolCallRequest,
     ToolCallResponse,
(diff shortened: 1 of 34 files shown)


libs/arcade-serve/arcade_serve/core/components.py
```diff
@@ -99,3 +105,60 @@ async def __call__(self, request: RequestData) -> HealthCheckResponse:
...
+            response_model_exclude_none=True,
+        )
+
+    async def __call__(self, request: RequestData) -> ListResourcesResult:
+        """
+        Handle the request to list resources.
+        """
+        tracer = trace.get_tracer(__name__)
+        with tracer.start_as_current_span("ListResources"):
+            params = ListResourcesParams.model_validate(request.body_json or {})
```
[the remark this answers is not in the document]
EricGustin: Invalid as described, but running it down found a real bug next door, so thank you for it.

Those bodies do not return "the first resource page with HTTP 200". They return **500**. `RequestData.body_json` is typed `JSONResponse | None`, and `JSONResponse` is `dict[str, Any]` (`arcade_serve/core/common.py:10`), so a non-object body fails at `RequestData` construction, which happens *above* the component. The `or {}` you flagged is unreachable for every input you listed. Measured against a live worker:

```
--- resources/list ---
  body=[]           -> 500
  body=false        -> 500
  body=0            -> 500
  body=""           -> 500
  body=null         -> 200
  body={} valid     -> 200
  body=absent       -> 200
  body=malformed {  -> 500
```

So I have not taken the suggested patch: with both spellings the value reaching `model_validate` is `{}` either way, and the comment it would carry would describe a case that cannot occur.

The real defect is the 500. The body is parsed and `RequestData` is built *outside* the `try` that maps errors, so `json.JSONDecodeError` and that `ValidationError` escape unhandled. A caller sending nonsense is told the worker broke. That matters here because the engine reads a 500 as a worker fault worth retrying and alerting on, and a 400 as its own bug.

Fixed by moving the parse inside the mapped block:

```python
try:
    body_json = json.loads(body_str) if body_str else {}
    request_data = RequestData(...)
except (json.JSONDecodeError, ValidationError):
    return _error_response(400, INVALID_PARAMS, "Invalid params: malformed request body")
```

**Flagging the scope, because it is wider than this PR.** `_wrap_handler` is shared by every worker endpoint, so this also moves `/worker/tools` and `/worker/tools/invoke` from 500 to 400 on a malformed body. That behavior is pre-existing and identical on `main`:

```
--- tools/invoke (untouched by this stack) ---
  body=[]           -> 500
  body=malformed {  -> 500
```

I fixed it here rather than filing it separately because this PR is what adds the `ValidationError` to 400 mapping in this file, and that mapping was dead for the most likely error. Say the word if you would rather it came out and went into its own PR.

One case left alone deliberately: a literal `null` body still reads as empty params. The router already maps an absent body to `{}`, and `null` for "no params" is close enough to omitted.

Test parameterizes over `[]`, `false`, `0`, `""` and a truncated `{`, all asserting 400.

libs/arcade-serve/arcade_serve/fastapi/worker.py
```diff
@@ -105,16 +116,36 @@ async def wrapped_handler(
...
+                # 404 here means "no such URI". A 404 from the router means the
+                # endpoint itself is absent, which is a different thing to the
+                # caller and is why only the read path reports one.
+                return _error_response(404, RESOURCE_NOT_FOUND, f"Resource not found: {e.args[0]}")
+            except InvalidCursorError:
+                return _error_response(400, INVALID_PARAMS, "Invalid cursor")
+            except ValidationError as e:
+                return _error_response(
+                    400, INVALID_PARAMS, f"Invalid params: {e.error_count()} error(s)"
+                )
```
byrro: This `except` wraps the whole handler, not the parse of caller input. A `ValidationError` from building `ToolCallError` on `/worker/tools/invoke` then goes out as `400 -32602 Invalid params` with no log.

That happens when a tool raises `RetryableToolError` with `additional_prompt_content` as a list, or `ToolExecutionError` with `extra` as a list: the exception constructors do not check types, and [the call to `output_factory.fail`](https://github.com/ArcadeAI/arcade-mcp/blob/81aec75e39cfaff00a1bb5e2ed74d49d9a9b3432/libs/arcade-core/arcade_core/executor.py#L70-L82) sits outside the executor's `except Exception`. The agent never gets the retry guidance.

We should map `ValidationError` only at the `model_validate` calls for caller params and leave a worker-side one as a 500.

libs/arcade-serve/arcade_serve/core/components.py
```diff
@@ -99,3 +105,62 @@ async def __call__(self, request: RequestData) -> HealthCheckResponse:
...
+
+    async def __call__(self, request: RequestData) -> ReadResourceResult:
+        """
+        Handle the request to read a resource.
+        """
+        tracer = trace.get_tracer(__name__)
+        with tracer.start_as_current_span("ReadResource") as current_span:
+            params = ReadResourceParams.model_validate(request.body_json or {})
+            current_span.set_attribute("resource_uri", params.uri)
+            return self.worker.read_resource(params.uri)
```
byrro: `start_as_current_span` records the exception and marks this span `ERROR` when `ResourceNotFoundError` leaves the `with`.

[`_wrap_handler`](https://github.com/ArcadeAI/arcade-mcp/blob/81aec75e39cfaff00a1bb5e2ed74d49d9a9b3432/libs/arcade-serve/arcade_serve/fastapi/worker.py#L138-L144) then turns that into a 404, which this protocol treats as a normal answer.

Same thing on [`ListResources`](https://github.com/ArcadeAI/arcade-mcp/blob/81aec75e39cfaff00a1bb5e2ed74d49d9a9b3432/libs/arcade-serve/arcade_serve/core/components.py#L136-L138) for `InvalidCursorError` -> 400.

A stale or unknown `resourceUri` is the read that actually happens in practice, so every render of a missing UI becomes an error-status span with a full stacktrace.

Thinking about telemetry hygiene, I wonder if we should close the span before those typed errors propagate, or open it with `record_exception=False` and `set_status_on_exception=False`, so a 404 / invalid cursor remains non-error. Because the full stacktrace is unlikely to add anything relevant and can potentially clutter and accumulate in whatever db is storing these events.

byrro (approved):
LGTM; left two comments as optional improvements

libs/arcade-serve/arcade_serve/fastapi/worker.py
```diff
@@ -105,16 +116,36 @@ async def wrapped_handler(
...
-                return await handler(request_data)
-            else:
+            try:
+                body_json = json.loads(body_str) if body_str else {}
+                request_data = RequestData(
+                    path=request.url.path,
+                    method=request.method,
+                    body_json=body_json,
+                )
+            except (json.JSONDecodeError, ValidationError):
```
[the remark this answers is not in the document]
EricGustin: Confirmed and fixed in 62e3cf03.

`json.loads` on bytes does its own encoding detection, so undecodable input raises `UnicodeDecodeError` before any JSON parsing happens. It is a `ValueError` but not a `JSONDecodeError`, so it went straight past the tuple and out as a 500.

Added to the tuple, with a test that posts `{"uri": "\xff\xfe"}` and asserts 400 / -32602.

libs/arcade-serve/arcade_serve/fastapi/worker.py
EricGustin: Confirmed, and done the way you described in 62e3cf03.

I reproduced the escape you found:

```
constructor accepted a list: ['a', 'b']
output_factory.fail raised: ValidationError
```

`RetryableToolError.__init__` annotates `additional_prompt_content: str | None` but never checks it, and the `output_factory.fail` call sits inside the `except ToolRuntimeError` clause, so a `ValidationError` raised there is not caught by the sibling `except Exception` and leaves `ToolExecutor.call()` entirely.

Params now go through one helper that converts `ValidationError` into a typed `InvalidRequestParamsError`, and the wrapper answers 400 only for that. Everything else, including a worker-side model that fails to build, stays a 500. `CallTool` moved onto the same helper so it keeps its 400 for a genuinely malformed body.

Test asserts a valid read whose handler fails to build a model comes back 500, and it fails without the change.

One thing this does not do: the agent still gets no retry guidance in that scenario, it now just gets an honest 500 instead of a 400 blaming its request. The missing type check on the error constructors is a real defect and I would rather fix it in `arcade-core` on its own than widen this PR. Filing it separately.

libs/arcade-serve/arcade_serve/core/components.py
EricGustin: Agreed, and taken the second option you offered, in 62e3cf03.

Both spans now open with `record_exception=False, set_status_on_exception=False`. `ResourceNotFoundError` and `InvalidCursorError` set an `outcome` attribute (`not_found`, `invalid_request`) and leave the status unset. `InvalidRequestParamsError` joins them, since a malformed body is the same kind of answer. Anything else still calls `record_exception` and sets `ERROR` explicitly, so suppression covers only the answers these endpoints give.

I preferred this over closing the span early because closing early loses the attribute and makes the ordering easy to get wrong on the next edit.

Four tests: a 404 read, a refused cursor, a served read, and a handler that raises `RuntimeError`. The last one asserts the span is still `ERROR` with an exception event, which is what stops this turning into blanket suppression later. The first three fail without the change.

Your point about volume was the deciding one: a stale `resourceUri` is the read that actually happens, so this was going to be the dominant span shape.

PR #923, opened 2026-08-26
refactor(core): move the resource schema into arcade-core [TOO-1931]

EricGustin:
## Summary

`arcade-serve` needs the resource shapes so it can serve them, and it cannot import `arcade-mcp-server`, because the dependency runs the other way. This moves the result-shaped half of the resource models down into `arcade-core` and re-exports them from `arcade_mcp_server.types`, so both libraries read one definition instead of drifting apart.

Part 1 of 6 in a stack. Replaces #917, which GitHub auto-closed when I renamed the head branch. Same commit, same content.

Resolves: https://linear.app/arcadedev/issue/TOO-1931

## Design decisions

The models land in `arcade_core/resource_schema.py`, parallel to the existing `arcade_core/schema.py` that holds the tool schema. Moved: `Result`, `PaginatedResult`, `BaseMetadata`, `Icon`, `Annotations`, `Cursor`, `Role`, `Resource`, `ResourceTemplate`, `ResourceContents`, `TextResourceContents`, `BlobResourceContents`, `ListResourcesResult`, `ListResourceTemplatesResult`, `ReadResourceResult`, `ReadResourceParams`.

The three request envelopes stay where they are. `ListResourcesRequest` and its siblings extend `JSONRPCRequest`, so moving them would drag `JSONRPCMessage`, `Request` and `RequestId` into `arcade-core` for no gain. A transport that carries no JSON-RPC envelope only ever needs the params object.

`ListResourcesParams` is new. `arcade-mcp-server` never modelled it, since `PaginatedRequest.params` is a bare `dict[str, Any]`.

### A serialization bug came along with the move

`Resource`, `ResourceTemplate` and `ResourceContents` all declare `meta` with `alias="_meta"`, but their model config never set `populate_by_name`. Combine that with `extra="allow"` and `Resource(meta=...)` gets quietly accepted as an *extra* field, then serialized under the wrong key:

```
>>> Resource(uri="u", name="n", meta={"ui": {"x": 1}}).model_dump(by_alias=True, exclude_none=True)
{"name": "n", "uri": "u", "meta": {"ui": {"x": 1}}}     # and .meta reads None
```

Nothing reads `meta`, so anything a developer put in `_meta` would have disappeared without an error at any layer. `Result` already set `populate_by_name`, which is why the classes behaved differently and why nobody caught it. Fixed here with a regression test, since these are the classes being moved.

## Scope

In scope: relocating the models, adding the missing params object, keeping `arcade_mcp_server.types` importable exactly as before.

Not in scope: anything that reads or serves these models. That starts in #918.

## Test plan

- [x] `uv run mypy .` clean in each of `arcade-core`, `arcade-serve`, `arcade-tdk`, `arcade-mcp-server`
- [x] 10 new tests in `libs/tests/core/test_resource_schema.py` covering the field names, the `_meta` alias in both directions, and the text-versus-blob distinction
- [x] One re-export test asserting the moved names are the *same objects* under both import paths, which is what catches a later edit redefining a model instead of re-exporting it

## Author checklist

---

> [!NOTE]
> **Medium Risk**
> Shared wire-contract models and stricter list-result validation can change runtime serialization and validation for any code building or parsing resource payloads; the `_meta` fix corrects previously silent data loss.
> 
> **Overview**
> **Centralizes MCP resource list/read wire models in `arcade_core.resource_schema`** so `arcade-serve` can share the same Pydantic types without importing `arcade-mcp-server`. `arcade_mcp_server.types` **drops the local definitions** and **re-exports** the moved symbols so existing imports keep working; JSON-RPC request envelopes (`ListResourcesRequest`, etc.) stay in the MCP server package.
> 
> Adds **`ListResourcesParams`** (optional pagination cursor) for non–JSON-RPC callers. **`ListResourcesResult.resources` is required** (no default empty list), so malformed payloads cannot masquerade as an empty catalog.
> 
> Fixes a **`_meta` / `meta=` serialization bug** by enabling `populate_by_name` on `BaseMetadata` (and related content models): without it, `meta=` was stored as an extra field and serialized as `meta` instead of `_meta`. Bumps **`arcade-core` to 4.12.0** and **`arcade-mcp-server` to 1.26.1** with a matching core dependency floor.
> 
> New tests lock **wire serialization** (`_meta`, text vs blob, strict list results) and assert MCP types **re-export the same class objects** as core (not duplicate models).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5fccf58abb3ded2e3a924617dec9ead3b97ceac9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-core/arcade_core/resource_schema.py
+++ b/libs/arcade-core/arcade_core/resource_schema.py
@@ -74,9 +74,6 @@ class ListResourcesParams(BaseModel):
 
 
 class ListResourcesResult(PaginatedResult):
-    # Required, as the specification has it, so an empty page has to be sent as
-    # one. Defaulted, a payload whose key is absent or misspelled parses as a
-    # worker with no resources, and the caller cannot tell the two apart.
     resources: list[Resource]
 
 
--- a/libs/arcade-core/arcade_core/resource_schema.py
+++ b/libs/arcade-core/arcade_core/resource_schema.py
@@ -102,11 +102,6 @@ class ResourceContents(BaseModel):
     mimeType: str | None = None
     meta: dict[str, Any] | None = Field(alias="_meta", default=None)
 
-    # extra="allow" matches Result and BaseMetadata above, and the reason is the
-    # same: a key this file has not modelled yet is carried rather than dropped,
-    # so the schema lagging the specification costs nothing. populate_by_name
-    # lets a caller pass ``meta=`` for the ``_meta`` field; without it, and with
-    # extra="allow", ``meta=`` is silently accepted as an extra field instead.
     model_config = ConfigDict(extra="allow", populate_by_name=True)
 
 
--- a/libs/arcade-core/arcade_core/resource_schema.py
+++ b/libs/arcade-core/arcade_core/resource_schema.py
@@ -1,0 +1,114 @@
+"""Resource schema for the worker format.
+
+The shapes a worker returns when a caller lists or reads the resources a
+toolkit ships: a listing entry, its contents as text or as a base64 blob, and
+the paginated results that carry them.
+
+Field names are spelled exactly as they go on the wire, including the
+underscore on ``_meta``, so nothing has to be renamed at either end.
+"""
+
+from typing import Any, Literal
+
+from pydantic import BaseModel, ConfigDict, Field
+
+Cursor = str
+Role = Literal["user", "assistant"]
+
+
+class Result(BaseModel):
+    meta: dict[str, Any] | None = Field(alias="_meta", default=None)
+
+    model_config = ConfigDict(extra="allow", populate_by_name=True)
+
+
+class PaginatedResult(Result):
+    nextCursor: Cursor | None = None
+
+
+class BaseMetadata(BaseModel):
+    name: str
+    title: str | None = None
+
+    # populate_by_name lets callers pass ``meta=`` for the ``_meta`` field. Without
+    # it, and with extra="allow", ``meta=`` is silently accepted as an extra field
+    # and serialized as ``meta``, which no caller reads.
+    model_config = ConfigDict(extra="allow", populate_by_name=True)
+
+
+class Icon(BaseModel):
+    """Icon metadata."""
+
+    src: str
+    mimeType: str | None = None
+    sizes: list[str] | None = None
+    theme: Literal["light", "dark"] | None = None
+
+    model_config = ConfigDict(extra="allow")
+
+
+class Annotations(BaseModel):
+    audience: list[Role] | None = None
+    priority: float | None = None
+    lastModified: str | None = None
+
+    model_config = ConfigDict(extra="allow")
+
+
+class Resource(BaseMetadata):
+    uri: str
+    description: str | None = None
+    mimeType: str | None = None
+    annotations: Annotations | None = None
+    size: int | None = None
+    icons: list[Icon] | None = None
+    meta: dict[str, Any] | None = Field(alias="_meta", default=None)
+
+
+class ListResourcesParams(BaseModel):
+    """Params for a resource listing request: an optional pagination cursor."""
+
+    cursor: Cursor | None = None
+
+    model_config = ConfigDict(extra="allow", populate_by_name=True)
+
+
+class ListResourcesResult(PaginatedResult):
+    resources: list[Resource] = Field(default_factory=list)
+
+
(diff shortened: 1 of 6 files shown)


libs/tests/arcade_mcp_server/test_types_reexport.py
```diff
@@ -0,0 +1,41 @@
+"""The resource models live in arcade-core; arcade_mcp_server re-exports them.
```
EricGustin: What exactly is the point of this file? I can only see a weak argument for having this "test"

EricGustin: Valid in part. The file was doing two jobs and only one of them was worth doing.

`test_json_rpc_request_envelopes_stayed_behind` asserted `not hasattr` on three names. It pins a non-decision, it goes red the day someone legitimately moves them, and it tells a reader nothing. Deleted.

I kept the identity check, and here is the argument, which you can still reject. The premise of this PR is that `arcade-core` and `arcade-mcp-server` read one definition. If a later edit redefines `Resource` in `arcade_mcp_server.types` instead of re-exporting it, the schema forks: two import paths, two classes, two serializations, and every test in the suite still passes. That is the state this PR exists to undo, and nothing else would notice it coming back.

So the file is now 20 lines and one test, and the docstring states that failure rather than narrating the move. If you read it and still think it is ceremony, say so and I will delete it.

libs/arcade-core/arcade_core/resource_schema.py
```diff
@@ -0,0 +1,114 @@
...
+class ReadResourceParams(BaseModel):
+    uri: str
+
+
+class ResourceContents(BaseModel):
+    uri: str
+    mimeType: str | None = None
+    meta: dict[str, Any] | None = Field(alias="_meta", default=None)
+
+    model_config = ConfigDict(populate_by_name=True)
```
byrro: Should we include `extra="allow"` here, matching `Result` and `BaseMetadata` in this file? If not, a future key on `contents[]` is dropped on parse with no error.

libs/arcade-core/arcade_core/resource_schema.py
```diff
@@ -0,0 +1,114 @@
...
+class ListResourcesParams(BaseModel):
+    """Params for a resource listing request: an optional pagination cursor."""
+
+    cursor: Cursor | None = None
+
+    model_config = ConfigDict(extra="allow", populate_by_name=True)
+
+
+class ListResourcesResult(PaginatedResult):
+    resources: list[Resource] = Field(default_factory=list)
```
byrro: A missing or misspelled `resources` key becomes `[]` here, which looks like an empty catalog. The likelihood that this will actually manifest seems very low, but just to be safe, perhaps we should make the field required, matching MCP and `ReadResourceResult.contents`, so an empty page has to be explicit?

byrro (approved):
two small nits / low priority comments; leaving it pre-approved if you don't find worth addressing them

libs/arcade-core/arcade_core/resource_schema.py
```diff
@@ -0,0 +1,122 @@
...
+
+    cursor: Cursor | None = None
+
+    model_config = ConfigDict(extra="allow", populate_by_name=True)
+
+
+class ListResourcesResult(PaginatedResult):
+    # Required, as the specification has it, so an empty page has to be sent as
+    # one. Defaulted, a payload whose key is absent or misspelled parses as a
+    # worker with no resources, and the caller cannot tell the two apart.
```
EricGustin:
```suggestion
```

libs/arcade-core/arcade_core/resource_schema.py
```diff
@@ -0,0 +1,122 @@
...
+class ResourceContents(BaseModel):
+    uri: str
+    mimeType: str | None = None
+    meta: dict[str, Any] | None = Field(alias="_meta", default=None)
+
+    # extra="allow" matches Result and BaseMetadata above, and the reason is the
+    # same: a key this file has not modelled yet is carried rather than dropped,
+    # so the schema lagging the specification costs nothing. populate_by_name
+    # lets a caller pass ``meta=`` for the ``_meta`` field; without it, and with
+    # extra="allow", ``meta=`` is silently accepted as an extra field instead.
```
EricGustin:
```suggestion
```

PR #931, opened 2026-09-04
feat(core): let a tool name its user interface [TOO-1941, TOO-1942]

EricGustin:
Resolves [TOO-1941](https://linear.app/arcadedev/issue/TOO-1941).
Resolves [TOO-1942](https://linear.app/arcadedev/issue/TOO-1942).

## Before

As a tooling author, I want a tool in my toolkit to come with a user interface that a client can render when the tool is called. I have written the interface as an HTML document and declared it in my toolkit with `@resource`, so the worker serves it under a URI qualified by my toolkit's name and version. When I look for a way to say which tool the document belongs to, `@tool` has nothing for it, the tool definition has no field that could carry it, and the only place such a pointer can be written is on the server object my toolkit's entrypoint builds, which installed-toolkit discovery never imports. I can start my toolkit locally and point a client that renders tool interfaces at it, and nothing renders: the tool listing carries no pointer, the server's own endpoint knows nothing about the document, and if the client runs in a browser it cannot even read the server's responses, because the server names an allowed origin only on the preflight.

## After

As a tooling author, I want the same tool to come with the same interface. I declare the document with `@resource(file="sum-list.html")` on a stub function beside the file, and the decorator hands me back the declaration. In the tool's module I import that declaration and write `ui=sum_list_ui` on the tool. When the tool is registered, its definition carries the document's full URI under `_meta`, and the document is registered with it under that same URI, because one derivation qualifies both. I never write a URI or a media type: a document a tool names is served as `text/html;profile=mcp-app`, and if I had declared it as something else the toolkit fails to load with an error saying a client will not render it. If I pass anything but a declaration as `ui`, the module fails to import. I can start my toolkit locally and point the same client at it, and the interface renders: the tool listing carries the pointer, the server's own endpoint lists and reads the document, and a client running in a browser from an allowed origin reads every response, because the origin is echoed on all of them rather than on the preflight alone.

## Why `@resource` returns the declaration rather than the function

A tool has to name its interface without a string that must match something declared elsewhere. Returning the declaration makes the module attribute the thing a tool imports and passes, keeps it callable so a toolkit's tests can still run a body that produces contents, and leaves `ui` with one accepted type. arcade-core 4.16.0 returned the bare function; nothing consumes that yet.

## Why every tool definition now serializes `_meta`

The catalog endpoint serializes a definition with every field present, so a tool that names no interface emits `"_meta": null`, exactly as `deprecation_message` and `metadata` do today for tools that set neither. A consumer decoding into an optional field reads null as absent, and the engine's decoder is tested to do so.

Releases arcade-core 4.17.0, arcade-tdk 3.12.0 and arcade-mcp-server 1.29.0.

~ 🐙 Eriq, Eric's Agent

---

> [!NOTE]
> **Medium Risk**
> Touches catalog registration, resource URI identity, and tool listing serialization; mis-qualification or duplicate paths fail load, but behavior is largely additive behind new `ui=` usage.
> 
> **Overview**
> **Tools can declare a renderable UI** by passing the `ResourceDeclaration` returned from `@resource` into `@tool(ui=...)`. Registration registers that document once (shared across tools), forces the MCP-app HTML media type for interfaces, and puts `ui.resourceUri` on the tool definition as `_meta`.
> 
> **`@resource` is reworked**: it now returns a `ResourceDeclaration` (callable for tests) instead of tagging the function with `__arcade_resource__`. Declarations can use `file=` beside the module; `ResourceRegistry.declare` resolves file/function bodies internally. **UI URIs qualify by the installed package that ships the declaration** (`interface_uri`), not the composing server, so multi-toolkit MCP apps avoid path collisions.
> 
> **Wire/API plumbing**: `ToolDefinition` gains optional `_meta` (serialized with `populate_by_name`). MCP conversion merges existing `_meta` with `arcade` metadata; the MCP server loads catalog resources at startup and echoes **CORS allow-origin on every HTTP response**, not only OPTIONS.
> 
> Bumps **arcade-core 4.17.0**, **arcade-tdk 3.12.0**, **arcade-mcp-server 1.29.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da1a763ac871d97a022a4f7308166280e83dc7a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-core/arcade_core/catalog.py
+++ b/libs/arcade-core/arcade_core/catalog.py
@@ -39,8 +39,6 @@ from arcade_core.resources import (
     InvalidResourcePathError,
     ResourceDeclaration,
     ResourceRegistry,
-    as_interface,
-    declared_contents,
     qualify,
     ui_pointer,
 )
@@ -172,10 +170,6 @@ class ToolCatalog(BaseModel):
     # worker surface. They are a separate collection with separate types; the
     # catalog is the carrier, not the owner.
     _resources: ResourceRegistry = PrivateAttr(default_factory=ResourceRegistry)
-    # URI to the declaration registered there and what registered it, so a second
-    # declaration of one path is refused and the same declaration reached twice,
-    # from two tools or from the module scan after a tool, is not.
-    _declared: dict[str, tuple[ResourceDeclaration, str]] = PrivateAttr(default_factory=dict)
 
     def __init__(self, **data) -> None:  # type: ignore[no-untyped-def]
         super().__init__(**data)
@@ -281,22 +275,19 @@ class ToolCatalog(BaseModel):
             logger.info(f"Server '{toolkit_name!s}' is disabled and will not be cataloged.")
             return
 
-        interface = getattr(tool_func, "__tool_ui__", None)
-        if interface is not None:
+        ui = getattr(tool_func, "__tool_ui__", None)
+        if ui is not None:
             try:
-                self._register_resource(
-                    interface,
-                    str(fully_qualified_name),
+                self._resources.declare(
+                    ui,
                     toolkit_name=definition.toolkit.name,
                     toolkit_version=definition.toolkit.version,
-                    as_tool_interface=True,
+                    as_interface=True,
                 )
-            except ToolkitLoadError:
-                raise
             except Exception as e:
                 raise ToolDefinitionError(
-                    f"Tool '{definition.name}' names {interface.name!r} as its user interface, "
-                    f"which could not be registered. Reason: {e}"
+                    f"Tool '{definition.name}' names {ui.name!r} as its user interface, which "
+                    f"could not be registered. Reason: {e}"
                 ) from e
 
         self._tools[fully_qualified_name] = MaterializedTool(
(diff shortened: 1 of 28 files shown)


libs/arcade-core/arcade_core/catalog.py
```diff
@@ -269,6 +275,21 @@ def add_tool(
...
             return
 
+        ui = getattr(tool_func, "__tool_ui__", None)
+        if ui is not None:
+            try:
+                self._resources.declare(
+                    ui,
+                    toolkit_name=definition.toolkit.name,
+                    toolkit_version=definition.toolkit.version,
+                    as_interface=True,
```
byrro, line 285: We're qualifying the UI from `definition.toolkit` which `MCPApp.add_tool` and `add_tools_from_module` [set to the app](https://github.com/ArcadeAI/arcade-mcp/blob/763741bcc7e49faf658fe4222f5d6e501c112402/libs/arcade-mcp-server/arcade_mcp_server/mcp_app.py#L353-L380). Two imported packages that both declare `panel.html` both become `ui://Combined/1.0.0/panel.html`.

`declare` raises, we wrap that as `ToolDefinitionError`, and the server never reaches `app.run()`. We [document](https://docs.arcade.dev/en/build/create-tools/tool-basics/organize-mcp-tools) this compose path, and it fails whenever the imported UIs share a path. The user loses every tool on that server. `Toolkit.from_package` keeps the source identities and is fine.

I think we should qualify the resource and `_meta.ui.resourceUri` with the originating toolkit, and still reject two different declarations at the same path inside one source toolkit

byrro (approved):
Left only one comment that looks important to address; otherwise LGTM. Pre-approving for when you deem appropriate to merge.

PR #932, opened 2026-09-08
feat(core): let a resource declare its own _meta [TOO-1943, TOO-1944]

EricGustin:
Resolves [TOO-1943](https://linear.app/arcadedev/issue/TOO-1943).
Resolves [TOO-1944](https://linear.app/arcadedev/issue/TOO-1944).

## Before

As a tooling author, I ship an interface that has to reach an external origin, or that should not be wrapped in the host's own bordered card. The extension puts all of that on the resource's `_meta.ui`: `csp` allowlists naming which origins my scripts, styles and network calls may come from, `permissions` for things like clipboard access, `domain`, and `prefersBorder`. I declare my document with `@resource` and there is nowhere to put any of it, so my interface renders and then has every outbound call refused by the host's default-deny policy, and gets a card drawn around it whether it wants one or not.

If I am writing a standalone server instead, `MCPApp.add_resource(meta=...)` looks like the channel. It is not: it fills the listing entry, and a host reads the rendering contract off the read response. A host may skip the listing for UI resources entirely.

## After

As a tooling author, I write `@resource(file="calculator.html", meta={"ui": {"csp": {...}, "prefersBorder": False}})`. The object is carried through untouched, and registration puts it on both the listing entry and the read response from that one value, so the two cannot disagree about what my document asked for. A host reads it where the extension says to look, my calls reach the origins I declared, and my card is mine to draw or not.

`MCPApp.add_resource(meta=...)` keeps its name and its meaning and now reaches the read response too, so a standalone server that was already setting it starts working without changing a line.

## Why the read response is the one that matters

The extension defines the contract on the `resources/read` contents item, requires a host to fetch a tool's interface with a read, and allows a server to omit UI resources from `resources/list` altogether. A document whose `_meta.ui` rides only on the listing therefore reaches a host that never asked for the listing with nothing at all.

## Three paths built a read body, and all three dropped it

`ResourceRegistry._store` for the worker protocol, `ResourceManager._coerce_result` for the standalone endpoint, and `ResourceManager.load_from_catalog` for a toolkit's declarations served on that endpoint. All three now carry it. A handler that builds its own contents objects owns their `_meta` and is left alone.

Verified end to end through the engine's real HTTP client against a live local worker: a tool carries `_meta.ui.resourceUri`, the read at that URI returns the document with `text/html;profile=mcp-app` byte-identical, and the resource's own `_meta.ui` arrives with `csp`, `permissions` and `prefersBorder` intact. The engine needed no change; its resource DTOs are the official Go SDK types and those already carry `_meta`. Coverage for that side is in ArcadeAI/monorepo#3984.

## One authored value, one slot-filler

`declare` is the only thing that fills both slots, from the value an author declared. The registry's low-level `add` fills the listing only, so a resource can never end up with a read slot that disagrees with its listing, which is a state the standalone server has no way to represent.

Releases arcade-core 4.18.0, arcade-tdk 3.13.0 and arcade-mcp-server 1.30.0. arcade-tdk's floor moves with arcade-core because it re-exports `resource()`, as it has in every release that changed `arcade_core.resources`.

~ 🐙 Eriq, Eric's Agent

---

> [!NOTE]
> **Medium Risk**
> Touches resource list/read payloads used by hosts for UI sandboxing; behavior change is additive but incorrect propagation could break CSP or permissions enforcement.
> 
> **Overview**
> Tooling authors can attach **out-of-band rendering metadata** (e.g. `ui.csp`, `ui.permissions`, `ui.prefersBorder`) when declaring static UI documents via **`@resource(..., meta=...)`**. Registration copies that dict **verbatim** onto both the **resources/list** entry and the **resources/read** contents item, so hosts that only read never miss the contract.
> 
> **arcade-core** adds `meta` on `ResourceDeclaration` and threads it through `ResourceRegistry.declare` into stored `Resource` and `TextResourceContents` / `BlobResourceContents` (`contents_meta` on `_store`). **`registry.add`** still sets listing `_meta` only—not read contents—matching the intended split.
> 
> **arcade-mcp-server** extends `ResourceManager.read_resource` and `_coerce_result` so registered or template `_meta` is applied when handlers return primitives; handlers that return their own `ResourceContents` lists are unchanged. Handler-less resources still return empty text with `_meta` attached.
> 
> Bumps **arcade-core 4.18.0**, **arcade-mcp-server 1.30.0**, **arcade-tdk 3.13.0**, with tests across core registry, MCP manager, and worker HTTP endpoints (wire `_meta` alias).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2f491df2b5b6861cd4f7a75e721f5884258caba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/arcade-tdk/pyproject.toml
+++ b/libs/arcade-tdk/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "arcade-tdk"
-version = "3.12.0"
+version = "3.13.0"
 description = "Arcade TDK - Toolkit Development Kit for building Arcade tools"
 readme = "README.md"
 license = { text = "MIT" }
@@ -16,7 +16,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 requires-python = ">=3.10"
-dependencies = ["arcade-core>=4.17.0,<5.0.0", "pydantic>=2.7.0"]
+dependencies = ["arcade-core>=4.18.0,<5.0.0", "pydantic>=2.7.0"]
 
 [project.optional-dependencies]
 dev = [
--- a/libs/arcade-core/arcade_core/resources.py
+++ b/libs/arcade-core/arcade_core/resources.py
@@ -165,6 +165,10 @@ class ResourceDeclaration:
     file: Path | None = None
     #: Produces the contents when no file is declared. Called once, at registration.
     func: Callable[[], Any] | None = field(default=None, compare=False, repr=False)
+    #: Out-of-band data for the resource, carried through untouched. A host reads a
+    #: document's rendering contract from the read response, so registration puts
+    #: this on the contents as well as on the listing entry.
+    meta: dict[str, Any] | None = None
 
     def __call__(self) -> Any:
         """Run the declaring function, so a toolkit's own tests can call it directly."""
@@ -277,6 +281,7 @@ def resource(
     title: str | None = None,
     description: str | None = None,
     mime_type: str | None = None,
+    meta: dict[str, Any] | None = None,
     scheme: str = UI_SCHEME,
 ) -> Callable[[Callable[..., Any]], ResourceDeclaration]:
     """Declare a static resource a toolkit ships.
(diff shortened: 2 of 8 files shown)


libs/arcade-tdk/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "arcade-tdk"
-version = "3.12.0"
+version = "3.13.0"
```
EricGustin, line 3: Bumping TDK because it reexports `resource()` from `arcade_core`, so it needs the new `meta` arg.

byrro (approved):
LGTM

PR #934, opened 2026-09-14
feat(cli): discovery, named contexts, login by URL, CI mode, no-Cloud guard [DEP-238]

pmdroid:
## What & why

DEP-238. Teaches the `arcade` CLI to target a self-hosted Arcade installation instead of only Arcade Cloud.

- **Discovery + login by URL** — `arcade login --url <install-url>` (or `ARCADE_URL`) fetches the installation's discovery document at `GET /.well-known/arcade`, signs in against the discovered coordinator, picks a project, and saves a **named context**. Deployments-off installations report the discovery reason and exit non-zero; a URL with no discovery document is refused before anything is sent (the "installation predates discovery" path).
- **Named contexts** — credentials now hold multiple named contexts (engine/coordinator/dashboard URLs + a `cloud`/`self_hosted` kind + auth/org/project). The legacy single-context file migrates to a `default` context on load. `arcade context list|use|show` manages and switches them.
- **No-flag targeting** — `arcade deploy`, `secret`, `org`, and `project` resolve their target from the active context; no flags needed. Unauthenticated deploy names the login command.
- **CI mode** — `ARCADE_URL` + `ARCADE_API_KEY` complete a deploy with no terminal and no credentials file (bearer API key auth).
- **No-Cloud guard** — with a self-hosted context active, any attempt to contact an Arcade Cloud host (`api.arcade.dev`, `cloud.arcade.dev`, `*.arcade.dev`) is refused. The hardcoded cloud defaults in `deploy`/`secret`/`org`/`project` were routed through the active context.

Cloud users see no prompt and no new required flag.

## Tests

`libs/tests/cli/test_context.py`, `test_login_url.py`, `test_contexts_cmd.py`, `test_deploy_context.py`:
legacy→default migration + named-context round trip, discovery parsing + predates path, cloud-host detection, the no-Cloud guard, login by URL (incl. disabled-deployments and missing-discovery), the context command, unauthenticated deploy naming login, and a CI-env deploy driven without prompts. Full `libs/tests/cli` suite green; `ruff` + `mypy` clean on changed libs.

## Proof

Backend/CLI-only change with no dashboard surface. Verified on the CLI via the unit suite above (behavior asserted with mocked discovery + OAuth). The e2e kind-lane run with the recording proxy and the Cloud-context regression against local-stack live in the monorepo BDD harness (out of this repo's scope); the spec promotion / tag moves are tracked there.

---

> [!NOTE]
> **High Risk**
> Changes authentication, credentials file format, and which hosts receive bearer tokens—mistakes could leak credentials to Cloud or break login/deploy for existing users.
> 
> **Overview**
> The CLI can target **self-hosted Arcade installations** alongside Cloud by introducing **named contexts** (engine/coordinator/dashboard URLs, `cloud` vs `self_hosted` kind, auth and org/project) stored in `credentials.yaml`, with legacy single-context files migrating to `default` on load.
> 
> **Login and discovery:** `arcade login --url` (or pinned `ARCADE_URL` at import) fetches `/.well-known/arcade`, OAuth-signs in via the discovered coordinator, and saves a named context; `arcade context list|set|show|delete|add` manages them. Global `--context` / `ARCADE_CONTEXT` run one command against a saved context without changing the default. **CI mode** uses `ARCADE_URL` + `ARCADE_API_KEY` (pinned at startup so project `.env` cannot retarget mid-command). A **no-Cloud guard** blocks `*.arcade.dev` hosts when a self-hosted context is active.
> 
> **Command wiring:** Deploy, secret, org, project, connect, and dashboard default to the active context’s URLs instead of hardcoded Cloud hosts; logout removes one context (`--all` clears everything). Deploy adds optional `--provider`, treats **degraded** as failure, sends CLI version headers, and exits non-zero on non-running outcomes.
> 
> **arcade-core:** `NamedContext` / multi-context save-load (backward-compatible flat keys), `coordinator_api()` URL normalization, and unified `ARCADE_WORK_DIR` config path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 272e949e73613b002f45ebc2ca3456c6551b936f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pmdroid:
Pushed four commits that were sitting uncommitted on my machine, each one a fix found by actually driving this CLI against a self-hosted installation.

## What landed

**Reaching one coordinator endpoint whichever address was configured.** Cloud is configured with the coordinator's bare host, while an installation's discovery document publishes its API base, which already ends in the API path. Every call site appended the path unconditionally, so signing in to an installation asked for `https://cloud.bosslevel.dev/api/v1/api/v1/auth/cli_config` and got a 404 that read like a missing coordinator. Six call sites now normalise first, and the segment is only stripped when it ends the address.

**Reading the discovery addresses in either shape.** Engines from the discovery release onwards nest the addresses under a `urls` object and older ones publish them flat. A CLI that understood one shape refused to sign in to half the installations it was pointed at. Both are accepted, top level wins over nested, and a `urls` field that is not an object is ignored rather than fatal.

**Naming the provider a server deploys to.** An installation running more than one provider needs to be told which one a server belongs on. The field is omitted from the request entirely when unset rather than sent as null, because an installation that knows one provider rejects it and a null would break every deploy that does not need it.

**Reading usage credentials from the active context.** Found by Bugbot and confirmed against the writer. Named contexts moved the signed-in coordinator and tokens under the active context, and the file no longer carries them at the top level at all, but usage identity still read the top level. It found no access token, skipped the whoami call, and every event carried an anonymous id instead of the principal. Nothing surfaced because the whole lookup sits inside a silent `except`.

## Proof

```text
$ make check
ruff / ruff-format / mypy across six libs: all passed

$ pytest libs/tests/core/usage libs/tests/core/test_coordinator_urls.py libs/tests/cli
936 passed, 1 skipped
```

Each fix ships the test that fails without it: `test_coordinator_urls.py` for the doubled path, three cases in `TestDiscovery` for the nested and flat shapes and their precedence, `test_deploy_request.py` for the omitted provider, and `test_identity_credentials.py` for the context lookup including files written before contexts existed.

## Dismissed

**Login does not survive a corrupt credentials file.** Real, and not this branch. `save_credentials_from_whoami` catches only `FileNotFoundError`, while `Config.load_from_file` also raises `ValueError` for unparseable YAML or a missing `cloud` key. Both are unchanged on `main`, and neither line is touched here, so this predates named contexts rather than being caused by them. The error does carry recovery advice, but it arrives after the browser round trip has already completed, which is the part worth fixing. Filed separately rather than widened into this PR.

**The echo example is renamed to `EchoServer2332`.** Correct observation, deliberately not committed. It is local scratch from testing deploys against staging and stays out of this branch.

🐕 Written by Kyoto, an AI agent, on Pascal's behalf —

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -101,6 +114,14 @@ def login(
...
     Logs the user into Arcade using OAuth.
     """
+    from arcade_cli.context import ARCADE_URL_ENV
+
+    resolved_url = url or os.environ.get(ARCADE_URL_ENV)
+
+    if resolved_url:
+        _login_with_url(resolved_url, context_name, timeout, debug)
+        return
+
```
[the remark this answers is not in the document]
pmdroid: Confirmed and fixed in 2c97e698.

The asymmetry was exactly as described. Logging in by url skips the guard, plain login consulted the active context, so a user signed in to a self-hosted installation could not add Cloud at all. Worse than the refusal itself, the advice it printed was to run `arcade logout`, which deletes the credentials file and with it every saved context.

The guard now asks about the context the login would actually write rather than whichever is active, so it still refuses a genuinely redundant sign-in and no longer refuses a new one.

While there I passed the context flag through to the save on that path. It was accepted and ignored, so naming a context without a url silently wrote to the default one, and a guard reading the named context while the save wrote elsewhere would have disagreed with itself.

Three cases cover it in `TestLoginGuardIsPerContext`: a context that was never saved is not signed in, the named context is the one checked, and omitting the name still checks the active one. The first fails on the bug.

🐕 Written by Kyoto, an AI agent, on Pascal's behalf —

libs/arcade-cli/arcade_cli/authn.py
```diff
@@ -816,38 +830,36 @@ def _credentials_file_contains_legacy() -> bool:
...
-        auth = cloud_config.get("auth", {})
-        if auth.get("access_token"):
-            email = cloud_config.get("user", {}).get("email", "unknown")
-            context = cloud_config.get("context", {})
-            org_name = context.get("org_name", "unknown")
-            project_name = context.get("project_name", "unknown")
+        if auth and auth.access_token:
+            email = user.email if user else "unknown"
+            org_name = context.org_name if context else "unknown"
+            project_name = context.project_name if context else "unknown"
```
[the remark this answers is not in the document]
pmdroid: Right, and it was my own fix that introduced it. Corrected in dfff2d14.

The guard reads a named context, so the message had no business calling that context's organisation and project `Active` when a different context was. Someone signed in to a self-hosted installation running plain login was shown their Cloud organisation as active while every other command kept using the self-hosted one.

The message now names the context it found and says which one is actually active, and only uses the word active when those are the same context. The word appears for an active context and is absent for one that is not, which is what the two new cases assert.

The follow-up advice also led with `arcade logout`, the one action that deletes every saved context. It now leads with `arcade context use <name>` and mentions logout for what it actually does.

🐕 Written by Kyoto, an AI agent, on Pascal's behalf —

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -101,8 +114,20 @@ def login(
...
+    target_context = context_name or "default"
+
+    if check_existing_login(context_name=target_context):
+        console.print("\nTo make it the active context, use ", end="")
+        console.print(f"arcade context use {target_context}", style="bold green", end="")
+        console.print(". To delete every saved context, use ", end="")
```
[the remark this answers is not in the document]
pmdroid: Right, and it was mine again. Fixed in the commit above.

Moving the guard onto a named context left the follow-up advice in the caller, which has no way to tell whether that context is the active one. So it told everyone to switch to a context they were usually already using, printed directly under a line saying that context was active.

The advice now sits beside the message it belongs to, where activeness is already known. An active session gets the logout wording it has always had, and a session in a different context gets the switch. The tests assert both directions: `logout` appears and `context use` does not for an active session, and the reverse for one that is not.

🐕 Written by Kyoto, an AI agent, on Pascal's behalf —

pmdroid:
Fixed the deploy authentication switch, which was the one finding I had held back as a release blocker.

## The defect

`arcade deploy` resolves auth headers per request and read the CI credentials straight from the environment each time. Partway through the command it loads the project's env file into the process, and a tool server project very often keeps an API key there. Calls before that point authenticated as the signed-in user over OAuth, and calls after it authenticated with the project's key. `get_org_scoped_url` resolves the same way, so the organisation in the path changed with the credential and a deploy could land somewhere the user never chose.

## The fix

The environment is captured once when the command starts, and every later resolution reads that capture. Credentials present at start still drive unattended runs exactly as before, which is the case the environment variables exist to serve. A file read later configures the server being deployed, which was never a statement about who is deploying it.

The capture is per command rather than per process. That matters for the test runner, which invokes many commands in one process, and the CLI tests reset it between cases because a real invocation gets a fresh process and they do not.

Four cases cover it: a project file cannot introduce credentials, cannot replace credentials that were already there, credentials present at start still drive unattended runs, and without pinning the environment is still read live.

```text
pytest libs/tests/cli libs/tests/core
1501 passed, 1 skipped
```

The one remaining failure, `test_two_resources_sharing_a_path_fail_the_toolkit`, is in a file whose last change is an ancestor of `main` and does not fail in CI.

## Dismissed: the coordinator falls back to Cloud in CI mode

The observation is right and the consequence is not. With CI credentials set, `resolve_ci_context` supplies no coordinator URL, so `resolve_coordinator_base_url` does fall through to the production Cloud host. It is not silent. `guard_no_cloud` runs on that URL and refuses outright when the active context is self-hosted:

```text
Refusing to contact the Arcade Cloud host '...' while the self-hosted context '...' is active.
```

For a CI context that genuinely points at Cloud, the fallback is the correct answer. The two callers are the `org` and `project` commands rather than deploy.

There is a real nit underneath it: that refusal advises `arcade context use <name>`, which is useless when the context came from the environment and there is nothing to switch to. Worth better wording, not worth widening this PR, and not introduced by this change.

🐕 Written by Kyoto, an AI agent, on Pascal's behalf —

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -1108,6 +1230,10 @@ def main_callback(
         help="Print version and exit.",
     ),
 ) -> None:
+    from arcade_cli.context import pin_ci_environment
+
+    pin_ci_environment()
```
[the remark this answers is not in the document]
pmdroid: Correct, and it made the previous fix do nothing at all. Thank you for chasing the import chain.

`main.py` imports `arcade_cli.utils`, which imports `arcade_mcp_server.settings`, which builds a settings object at import time and loads the project env file as it does. All of that completes before any command body runs, so capturing inside the command callback recorded values the env file had already supplied.

The capture now happens when the `arcade_cli` package itself is imported, from a module whose only dependency is the standard library, so nothing it pulls in can have loaded a file first.

The proof runs in its own process, since the ordering only exists across a real import. It writes an env file naming an installation and an API key, imports the CLI from that directory, and asks whether unattended mode is armed:

```text
with the capture:     NOT_CI
without the capture:  CI
```

I verified that second line by emptying the package init and re-running, so the test fails on the defect rather than passing vacuously.

🐕 Written by Kyoto, an AI agent, on Pascal's behalf —

libs/arcade-core/arcade_core/config_model.py
```diff
@@ -278,8 +355,20 @@ def save_to_file(self) -> None:
...
 
-        # Convert to dict, excluding None values for cleaner output
-        data = {"cloud": self.model_dump(exclude_none=True, mode="json")}
+        if not self.contexts:
+            self.contexts = {"default": self._to_named_context()}
+            self.active_context = "default"
+        elif self.active_context and self.active_context in self.contexts:
+            self.contexts[self.active_context] = self._to_named_context()
+
+        cloud = {
```
sdreyer, line 364: **This makes the credentials file one-way.** `save_to_file` now writes only `active_context` + `contexts`, dropping the flat `auth` / `user` / `context` keys. `BaseConfig` is `extra="ignore"`, so arcade-core 4.18 parses the new file into an all-`None` Config. I ran the pre-PR model against a new-format file:

```
old-core reads new file -> authenticated: False | auth: None | user: None | ctx: None
```

The old->new direction is fine — I ran a legacy file through this loader and it migrates to `default`, stays authenticated, keeps `kind: cloud` and the org/project. It's the reverse that bites: a CLI downgrade, or any environment still on 4.18, sees a logged-out user.

And it isn't opt-in. `auth_tokens.py:106` saves on token refresh, so the first authenticated command flips the format permanently.

**Suggested fix — dual-write.** Keep emitting the flat active-context keys alongside `contexts`/`active_context`. `load_from_file` already branches on `"contexts" in cloud` and overwrites the flat values via `_apply_named_context`, so old and new readers both work off one file and the downgrade cliff disappears. Roughly five lines here, and it's worth more than the floor bumps because a PyPI release can't be unpublished and a migrated file can't be un-migrated.

**Separate nit, a few lines up:** the `if not self.contexts / elif active_context in self.contexts` pair has no `else`. If `contexts` is populated but `active_context` is `None` or names a missing key, neither branch runs and the in-memory flat state — including a token just refreshed — is silently dropped from the write. `load_from_file` always leaves those consistent today, so I couldn't reach it, but it's a data-loss branch with no diagnostic.

libs/arcade-cli/arcade_cli/context.py
```diff
@@ -0,0 +1,190 @@
...
+        is_ci=False,
+    )
+
+
+def try_resolve_active_context() -> ResolvedContext | None:
+    try:
+        return resolve_active_context()
+    except FileNotFoundError:
+        return None
+    except Exception:
```
sdreyer, line 176: **The no-Cloud guard fails open.** `guard_no_cloud` gets its context from `try_resolve_active_context`, which swallows every exception into `None`, and then returns early on `ctx is None`. So any failure to read the credentials file — malformed YAML, a `ValidationError` on one context, a permissions error — silently disables the guard rather than tripping it.

For a feature whose own error message promises "Artifacts and requests stay inside your environment," open is the wrong direction to fail. The blanket `except Exception` also hides the reason, so a self-hosted user gets no signal that the protection stopped applying.

Worth distinguishing the two cases: `FileNotFoundError` means genuinely no config and unguarded is correct; anything else means a config exists but couldn't be read, which should refuse rather than proceed. Cheap version — have `guard_no_cloud` catch the load failure itself and raise `NoCloudGuardError` with the underlying reason, leaving `try_resolve_active_context` alone for the callers that legitimately want a soft default.

libs/arcade-cli/arcade_cli/utils.py
```diff
@@ -657,6 +702,11 @@ def get_auth_headers(coordinator_url: str | None = None) -> dict[str, str]:
     from arcade_core.constants import PROD_COORDINATOR_HOST
 
     from arcade_cli.authn import get_valid_access_token
+    from arcade_cli.context import try_resolve_active_context
+
+    ci_context = try_resolve_active_context()
+    if ci_context is not None and ci_context.is_ci and ci_context.api_key:
+        return {"Authorization": f"Bearer {ci_context.api_key}"}
```
sdreyer: Not a defect — a request, while this code is open.

`get_auth_headers` returns `Authorization` only, with no `User-Agent` or `X-Arcade-CLI-Version`. That means the engine has no way to tell which CLI is talking to it, so if an installation begins reporting `degraded`, it cannot be gated per-client — it's all-or-nothing, and the only lever left is waiting on adoption.

Adoption will be slow: `arcade update` notifies rather than auto-installs, so the tail is pinned CI, Docker images, and `uv tool install` nobody re-runs. `cli_version` already goes to PostHog (`usage/constants.py:7`), so you can watch the distribution — but watching isn't gating.

A few lines adding the version header here turns step-5 of the rollout from "wait and hope" into something controllable.

libs/arcade-cli/arcade_cli/deploy.py
```diff
@@ -971,12 +988,19 @@ def deploy_server_logic(
...
-    # Offer to view full deployment logs
-    if all_logs and Confirm.ask("\nView full deployment logs?", default=False):  # type: ignore[arg-type]
+    interactive = sys.stdin.isatty() and sys.stdout.isatty()
+    if all_logs and interactive and Confirm.ask("\nView full deployment logs?", default=False):
         with console.pager(styles=True):
             console.print("[bold]Full Deployment Logs[/bold]\n", style="cyan")
             for i, log_line in enumerate(all_logs, 1):
                 console.print(f"{i:4d} | {log_line}", style="dim")
+
+    if final_status != "running":
```
sdreyer, line 1005: Worth calling out in the release notes: this changes the exit code. Previously a `failed` deployment printed "Deployment failed" and exited **0** — the `if running / elif failed` pair had no `raise`. Pipelines that were silently green on broken deploys will now correctly go red. Right fix, but it'll look like a regression to whoever owns those pipelines.

The mirror-image hazard is on older clients. The pre-1.16 poll loop is `while status in ["pending","unknown","updating"]` with `break` on `["running","failed"]`. Feed it `degraded` and the loop falls out, neither branch matches, and it exits 0 with no message — a silent pass on a degraded rollout. That's an argument for releasing this CLI ahead of any server change that introduces the status, not after.

Makefile
```diff
@@ -32,7 +32,7 @@ check-libs: ## Run code quality tools for each lib package
 .PHONY: test
 test: install ## Test the code with pytest
 	@echo "🚀 Testing libs: Running pytest"
-	@uv run pytest -W ignore -v libs/tests --cov=libs --cov-config=pyproject.toml --cov-report=xml
+	@ARCADE_WORK_DIR=$$(mktemp -d) uv run pytest -W ignore -v libs/tests --cov=libs --cov-config=pyproject.toml --cov-report=xml
```
sdreyer, line 35: Two things about this line — the intent is right, and I've done the same thing to my own credentials file.

**1. The guard only covers `make test`.** CLAUDE.md documents single tests as `uv run pytest libs/tests/core/test_toolkit.py::TestClass::test_method`, which is unprotected and still writes to the real `~/.arcade`. That's the exact accident the commit message describes. The commit is right that a fixture can't help — the constant is bound at import — but `os.environ.setdefault("ARCADE_WORK_DIR", tempfile.mkdtemp())` at the top of `libs/tests/conftest.py` runs before any test module imports `arcade_core`, and covers every invocation path rather than one target.

**2. Setting this var puts the suite in a split-brain config.** `ARCADE_WORK_DIR` resolves to two different files:

```
constants.py  -> $ARCADE_WORK_DIR/.arcade/credentials.yaml
config_model  -> $ARCADE_WORK_DIR/credentials.yaml
SAME: False
```

`constants.py:4` appends `.arcade`; `Config.get_config_dir_path()` uses the var as the directory itself. Pre-existing — both are on `main`, and `check_existing_login` behaves the same either way — but this line makes the divergent configuration the default for every test run. The sharpest symptom for anyone who sets the var: `login` writes via `Config.save_to_file()` while `logout` deletes `CREDENTIALS_FILE_PATH`, so **`arcade logout` prints "You're not logged in" and leaves the credentials on disk.**

History, for whoever fixes it: `get_config_dir_path()` is the original (#31, then #50 swapped the settings object for the env var, same semantics — the var *is* the config dir). `constants.py` introduced the second reading later in #635. Seniority argues one way but call sites argue the other: the `.arcade`-suffix reading has five consumers (`usage/identity.py`, `authn.py`, `main.py`, `command_tracker.py`, `update.py`) against essentially just `config_model` itself. Unifying `get_config_dir_path()` onto `ARCADE_CONFIG_PATH` changes one function, leaves the five callers alone, and keeps this Makefile line working as written.

Happy to send that as a separate patch if you'd rather keep this PR focused.

sdreyer (commented):
Reviewed for backwards compatibility, on-disk format, and CLI/engine wire compat. Ran the CLI suite at `0c628a4b` in a clean worktree: **903 passed, 1 skipped**. The DEP-305 import-ordering fix is correct and well-placed — capturing in `arcade_cli/__init__.py` from a stdlib-only module is the only spot early enough to beat the settings import that loads the project `.env`.

The wire story is clean in both directions. I diffed the full URL surface against `main`: every path an existing flow hits is byte-identical, `coordinator_api()` is pure normalization, and `provider` is dropped from the create-deployment body when unset (`CreateDeploymentRequest` has no other nullable fields, nested included), so an un-updated engine receives the exact old payload. `account_id` was already required on `main`, so there's no new coordinator obligation. New CLI against today's prod engine is safe.

One genuine break and one fail-open guard, plus rollout sequencing. Details inline; items that can't be anchored to the diff are below.

---

**Version floors weren't raised on the sibling packages.** `libs/arcade-mcp-server/pyproject.toml:24` and `libs/arcade-tdk/pyproject.toml:19` still declare `arcade-core>=4.18.0`, while core goes to 4.19.0 here. Combined with the credentials format change (see the inline comment on `config_model.py`), a pinned or Docker-built server env resolving 4.18 alongside a freshly-updated CLI silently loses its login — `server.py:538` just logs "Arcade access token not configured" and auth tools start returning login instructions. CLAUDE.md's rule about bumping minimum dependency versions on breaking inter-library changes covers this case.

**CI mode can still be armed by ambient shell env.** DEP-305 closed the `.env` vector, but `ARCADE_URL` + `ARCADE_API_KEY` exported in a shell still silently switches an authenticated cloud user from OAuth to API-key auth and reshapes `/secrets` from org-scoped to `/v1/admin` — with an existing login sitting right there, unused, and nothing printed. `ARCADE_API_KEY` is already set in plenty of shells; `ARCADE_URL` being new is the only thing making the pair rare today. One line on stderr when CI mode wins over a saved context would make this diagnosable instead of mysterious.

**Rollout ordering for `degraded`.** Whether an installation can report that status depends on the deployment provider behind it, so it is not a given everywhere. Where it can occur, a CLI older than 1.16 mishandles it (see the inline note on `deploy.py`). Releasing this CLI ahead of any server change that introduces the status costs nothing and means operators already have a client that handles it.

**Two nits.** `org.py` / `project.py` annotate `host: str` while defaulting to `None` — harmless at runtime since `typer.Option` returns `Any`, but the annotation now says the opposite of what the code does. And the autouse `forget()` in `libs/tests/cli/conftest.py` means every in-process test exercises `_startup_environment`'s live-env fallback rather than the captured path; the pinning behavior rests entirely on the one subprocess test. Deliberate, I assume — just worth knowing that's the single thread holding it.

None of this blocks the feature work, which reads well — the named-context migration is careful and the discovery/predates path is a nice touch.

libs/arcade-cli/arcade_cli/main.py
```diff
@@ -79,11 +81,25 @@
...
+        "--context",
+        "--context-name",
+        help="Name to save the context under (defaults to the installation host).",
+    ),
     host: str = typer.Option(
         PROD_COORDINATOR_HOST,
         "-h",
         "--host",
-        help="The Arcade Coordinator host to log in to.",
+        help="Legacy, use --url instead. Still the way to reach a local coordinator, "
```
EricGustin, line 101: Should we do this for every single CLI command that has `--host`? Is there a plan for a major bump to officially remove `--host` ?

libs/arcade-cli/arcade_cli/contexts_cmd.py
```diff
@@ -0,0 +1,99 @@
...
+        ctx = contexts[name]
+        is_active = "✓" if name == config.active_context else ""
+        table.add_row(name, ctx.kind, ctx.engine_url or "-", is_active)
+
+    console.print(table)
+    console.print("\nUse 'arcade context use <name>' to switch contexts.\n", style="dim")
+
+
+@app.command("use", help="Switch the active context")
+def context_use(
```
EricGustin: Why `use`? The existing pattern is `set` (see orgs and projects)

EricGustin (commented):
1. `arcade login` is scoped to one context, but `arcade logout` is scoped to all contexts (global), so they unexpectedly aren't inverses. Perhaps `logout` should be scoped to the active context, with an optional `--all` flag. And also possibly add a `context delete`.

2. Also consider `arcade context add <url>` as an alias for `arcade login --url <url>` so that `context` can have a complete lifecycle

3. Consider adding a per-command `--context <name>` and an `ARCADE_CONTEXT` env var so we can play nice with CI and scripts.

libs/arcade-cli/arcade_cli/contexts_cmd.py
EricGustin: `use` can still be a hidden alias if we feel strongly about `use`

libs/arcade-cli/arcade_cli/main.py
sdreyer: Left it out of the others since this is the only one that goes through a different flow than --host. All others are unnecessary now, but are still functional

sdreyer: And yeah we could do a major bump, but can probably wait until we need other things that require a major bump

EricGustin (approved):

PR #936, opened 2026-09-17
feat(arcade-core): add Calendly auth provider class (TOO-2178)

EricGustin:
Adds a `Calendly` helper class to `arcade_core.auth`, re-exported from `arcade_mcp_server.auth` and `arcade_tdk.auth`.

Tracked as [TOO-2178](https://linear.app/arcadedev/issue/TOO-2178/add-calendly-auth-provider-class-to-arcade-mcp).


samkujovich (approved):
LGTM. Reviewed against the monorepo side, since I'm shipping the Calendly toolkit that consumes this ([monorepo#4145](https://github.com/ArcadeAI/monorepo/pull/4145)) - this closes a real bug there.

**`provider_id: "calendly"` is the right string, confirmed in both engine configs:**

```yaml
# apps/engine/engine.dev.yaml
- id: arcade-calendly
  provider_id: calendly

# apps/engine/distribution/config/engine.yaml   <- self-hosted
- id: default-calendly
  provider_id: calendly
```

That is exactly the point your test comment makes, and it is load-bearing in a way worth spelling out: my toolkit currently hard-pins `OAuth2(id="arcade-calendly")`, which resolves on Cloud and **cannot resolve self-hosted**, where the instance is `default-calendly`. So today it is a Cloud-only toolkit, against [ADR-0002](https://github.com/ArcadeAI/monorepo/blob/main/docs/decisions/adr-0002-on-premises-deployment.md). `Calendly()` fixes that, and my `tests/test_auth.py` is currently asserting the pinned id, so it is holding the bug in place. Both are mine to fix once this lands.

Worth noting the pattern more generally - 497 `requires_auth` declarations in the monorepo use a helper class, 38 hard-pin an id, and the other five toolkits that pin (glean, salesforce, servicenow, snowflake, zendesk) are all per-customer-instance services, which is what "recommended for private tools only" is describing. Calendly is public SaaS, so it belongs in the helper-class group.

**Checked and clean:**

- Alphabetical placement correct in all three lists.
- Every file in the repo that enumerates providers is updated - searched for a neighbour (`Attio`) and the five files it appears in are exactly the five you touched. Nothing missed.
- Version graph is consistent and the floors are raised in the right direction: `arcade-tdk` 3.14.0 requires `arcade-core>=4.19.0`, `arcade-mcp-server` 1.31.0 requires both. So `arcade_tdk.auth.Calendly` cannot be installed against a core that lacks it.
- `scopes` defaults to `None`, which matters here more than usual - Calendly's authorize endpoint rejects the whole request if any granular scope is present, so the scope-free default is the only workable config. Your `assert auth.scopes is None` pins it.
- No error adapter needed. Only google, graphql, http, microsoft and slack ship dedicated ones; everything else uses the generic HTTP adapter, which I verified classifies Calendly's statuses correctly against the live API (404 as `UPSTREAM_RUNTIME_NOT_FOUND`, 429 retryable with `Retry-After` parsed).
- The README count is now actually right. It said 22 while listing 21 names; adding `Calendly` and `MicrosoftPowerBI` and bumping to 23 makes the number match the list for the first time.

One sequencing note for whoever merges: I can't adopt this until 1.31.0 is on PyPI (latest is 1.30.1 today), and I'd rather take it *before* merging #4145 than ship a toolkit that cannot authenticate self-hosted and fix it after.

